Issue #23, opened 2018-12-03
discussion about linux migration


yaute74:
To continue the discussion about linux migration [post=17225693.44]
When a user will upgrade to the 1.435 he will start with en empty database.
To prevent this problem I propose to modify the script ho.sh (to copy the database) like this:

```
#!/bin/sh
LAUNCHDIR="$(dirname "$0")"
LAUNCHDIR="`eval echo $LAUNCHDIR`"
HOHOME="~/.ho"
HOHOME="`eval echo $HOHOME`"
HOOLDHOME="~/.hattrickorganizer"
HOOLDHOME="`eval echo $HOOLDHOME`"

start(){

HODIR=/usr/lib/ho/

if [ ! -d "${HOHOME}" ]
then
echo "creating ${HOHOME}"
mkdir -m 777 "${HOHOME}"
echo "copying required ressource"
cp -R "${HODIR}/." "${HOHOME}"
fi

# For old versions
if [ ! -d ${HOHOME}/db ] && [ -d ${HOOLDHOME}/db ]
then
cp -r ${HOOLDHOME}/db ${HOHOME}/db
fi
..
```
Why did you say on the hattrick forum: "What you propose I guess will not work as you expect" ?

You propose to do it in the post install, but in the ho.sh script we create HOHOME directory, copy some files from HODIR. In that case it should be do in the post install, no ?

You plan to add it in the 'rpm' and 'deb' package ?
I don't know how Gradle work, but I can try to help you, if you need

akasolace:
@yaute74 ,  actually , I read too quickly your post in the forum this morning.
Now I see what you want to do, and you proposal looks good to me -yes, it will be done both for the .deb and the .rpm.
You can either open a pull request or I can make the changes you proposed myself, up to you, just let me know how you want to proceed

On another topic, as you are a Linux user.  What do you think of AppImage format? Do you think it would make sense to use it to distribute HO?

akasolace:
> -change ~/.ho by ~/.local/share/ho (because it's more standard for new product)

I actually did that in the version of the 1.435 beta, and some users complained about it, saying it was too difficult to find  :-)

> copy the db directory

copy it where ?

> do a backup of .hattrickorganizer

with current script, this folder is being left untouched, hence why do you want to back it ?

> If we can easly generate appimage with Gradle I think it's will be nice to propose deb, rpm and appimage.

Unfortunalty I do not think it is easy, I will leave as is for the time being,

akasolace:
closed by PR

PR #24, opened 2018-12-03
copy db from old user directory to new directory on Linux

yaute74:
my proposal for #23

### DIFF
--- a/buildResources/Linux/ho.sh
+++ b/buildResources/Linux/ho.sh
@@ -1,34 +1,42 @@
 #!/bin/sh
 
 LAUNCHDIR="$(dirname "$0")"
-HOHOME="~/.ho"
-HOHOME="`eval echo $HOHOME`"
 LAUNCHDIR="`eval echo $LAUNCHDIR`"
+HOHOME="~/.local/share/ho"
+HOHOME="`eval echo $HOHOME`"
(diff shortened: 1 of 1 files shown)


akasolace:
Thank you for the fast PR,

what ennoys me is that, there are now few beta testers that are running the Linux version.
This commit would break there install:
1) either we consider this in the script (maybe overdone for only 10 beta testers
2) I just communicate about it to beta testers

It is late, I'll think about it tomorrow. Meanwhile, you are input are welcome

yaute74:
For the backup you are right, it's better to do nothing. The user will be able to delete it after if he want.

For the ~/.local/ho it's not critical, it's just to be more standard. As we rename the home user, for the 435, I think it's the good time to do it. If users are annoyed by that leave it like that but it's a shame.

akasolace:
@yaute74  ok then I propose to keep the installation folder in ~/.ho , it is really a details and it would avoid to mess up things with the beta testers (I see it has been downloaded 26 times already).

Concerning the first point, you want to leave it untouched or do you want to ask the question to user within the script?

Could you please amend you PR accordingly, I will then merge it. 
Thank you

yaute74:
Personally I think the simples solution is the best: untouched

akasolace:
thank you

Issue #22, opened 2018-12-02
There no FW normal value in PlayerCompare (TAB Player Analysis)


mistermax80:
![schermata 2018-12-02 alle 17 25 49](https://user-images.githubusercontent.com/15519907/49342140-6f9dcb80-f657-11e8-97a7-149cf7ba4fa1.png)

mistermax80:
No this.
Following my preference - columns - squad:
![schermata 2018-12-02 alle 21 07 56](https://user-images.githubusercontent.com/15519907/49344372-66702700-f676-11e8-950a-8f36af538184.png)
And I have the issue in PlayerCompare:
![schermata 2018-12-02 alle 21 09 20](https://user-images.githubusercontent.com/15519907/49344391-b6e78480-f676-11e8-851c-631062830be7.png)
As you can see missing the FW value.

MMX80

akasolace:
@mistermax80 
are you making progress on this one ?

mistermax80:
No yet... Firstly I must complete my dev environment.
I have and issue on LookAndFeel Aqua when run in Intellij
I'm using macos.

akasolace:
@mistermax80 
It is a dependency issue. 
You should synchronize your gradle project. did you follow that [guide](https://www.jetbrains.com/help/idea/gradle.html#gradle_import)?

mistermax80:
I closed for error.

mistermax80:
@akasolace 
Sorry, but I don't understand how oper an PR.
Can you help me?

akasolace:
so you manage to run the code and to close the bug on your local installation, right?
Now you want to push  your changes?

I see that 

may be you can look at the contribution guidelines and in case you have further questions please let me know.
https://github.com/akasolace/HO/blob/master/docs/contributing.md

If you are new to git it may look intimidating, but you will soon appreciate how powerful it is

mistermax80:
@akasolace  ok thanks, I will update you soon...

mistermax80:
@akasolace 
I think to do done :)
Now I don't know what I will do.
Can you help me for the ultimate step?

akasolace:
Thank you !!

PR #25, opened 2018-12-07
#22

mistermax80:
There no FW normal value in PlayerCompare (TAB Player Analysis) #22


mistermax80:
@akasolace please... Now I stop activities.
But I'd like to know if my activities are correct.
1) fork your repo
2) in my master fork I create a branc named #issue22
3) commit and push
4) compare with your master and open request
I tried same test on git, because I never used.
Now I have in my rep 2 branc, what do I must?

PR #26, opened 2018-12-08
remove duplicated country

brokenelevator:
While adding new countries Palestine was duplicated which caused SQL exceptions.


akasolace:
Thank you

brokenelevator:
Happy to help :)

Issue #35, opened 2018-12-13
Question about jar in HOHOME


yaute74:
Hello,

On Linux, in the ho.sh , if the directory ~/.ho don't exist, we create it and copy jar from /usr/lib/ho/
After that we start the application from ~/.ho

When I install the new release, as the directory ~/.ho already exist it don't copy the new jar from /usr/lib/ho and I use the old revision.
Why we don't use jar from /usr/lib/ho ? If it's mandatory I think we must modify the script

PS: in the Splash screen I have 1.435 **DEV** (I don't know if it's normal)

akasolace:
but in case of update, jars in  ~/.ho would be more recent than jars in /usr/lib/ho
so we would need to check if there are more reccent and copy only if it is the case

yaute74:
Sorry but I don't understand.
How jar in ~/.ho could be more recent than jar in /usr/lib/ho ?
When we update the application it install the jar in /usr/lib/ho

akasolace:
# Intro

1) until 1.434 included HO was hosted in `~/.hattrickorganizer`
2) new .deb creates data in `/usr/lib/ho/`

# .sh execution

#### step 1
```
if ~/.ho don't exist:
    copy data from /usr/lib/ho/ into ~/.ho
    if ~/.hattrickorganizer exists:
          copy db from ~/.hattrickorganizer into ~/.ho
```

#### step 2

```
if script launched from "/usr/bin":
     cd ~/.ho
else:
   stay in zip directory
```

# fresh install .db
        new files are installed in /usr/lib/ho/ 
       ~/.ho won't be updated because it already exists
       => this is not working
       the fix could be that the data copy from /usr/lib/ho/ into ~/.ho should be part of the post install script instead of ho.sh ?

# fresh install zip
         => should be ok

# update from .db
     new data are installed in ~/.ho   => should be ok

# update from zip
    new data are installed in folder => should be ok

akasolace:
> 
> 
> Sorry but I don't understand.
> How jar in ~/.ho could be more recent than jar in /usr/lib/ho ?
> When we update the application it install the jar in /usr/lib/ho

No, when we update the app the jars and other files are installed in /usr/lib/ho if launched from usr/bin or installed in the unzipped folder, if launched from .sh within unzip folder

akasolace:
I just realized my proposed solution would be fine for .deb and .rpm but maybe not for your arch packaging so maybe instead we can use one of the following alternatives.
[...]

PR #46, opened 2018-12-15
Linux startup

yaute74:
issue: #35 
Finaly do it in postintall is not the good solution, because we have rights problems (package are install with root) and if we have several users it's problem because we must copy for all users.

The solution I propose is close to what you proposed with the date of the jar.
We extract the version from the manifests and compare them.


akasolace:
should it not be  copy if   "$HOMEVERSION" < "$LIBVERSION"  ?

tonight I made quite some change on the way version and version updates are managed
could you just confirm your code is working with latest master and then I will merge it.
Thank you !

yaute74:
I made it as simple as possible because I assume that the jar in ~/ho could not be more recent than jar in /usr/lib/ho. 
For me /usr/lib/ho is the reference (corresponding to the install package), so if the versions are different we must replace it.

Yes I make some tests and it was ok. I will retest it when you create a new dev release.
Thanks

akasolace:
> So the packager tell you that you are in 1.435 but when you start the application you are in 1.436, it's not clean.

You mean if you look at package version via the command line? Does it really matters?

>Do you think that disable the menu when we are on linux and using a packager is a good solution ?

For sure it is feasible to disable it for all linux user but that wont be a good idea, 
Maybe it is possible to disable it for arch user if we can identify them, I am not sure this is feasible.
We can also imagine, an hidden file like `.disableupdate` , which disable the update functionality. That hidden file would only be added by the arch installer.  The issue #35 would still have to be close for other Linux users.

Personaly, I would really prefer to keep is as it is. Maybe we can just imagine, a popup window for Linux user,  informing them that the OS will not be aware of the update and the version displayed by the system will not be updated

yaute74:
>You mean if you look at package version via the command line? Does it really matters?

For the operation of the application, it's not a problem, it's just not clean

>to disable it for arch user if we can identify them

It's not necessary to do a specific functionality for Arch because it's the same problem for all Linux distro and it's better to have the same behaviour for the maintenance.

>I would really prefer to keep is as it is

ok you're the leader :)

akasolace:
> For the operation of the application, it's not a problem, it's just not clean

Agree for Linux user, I will put a warning before executing the update

> >  I would really prefer to keep is as it is

> ok you're the leader :)

However, we still need to change the sh script, are you still wiling to do this?

Thank again for your help

yaute74:
>Agree for Linux user, I will put a warning before executing the update

Ideally only for Linux user, using packager. If you want I can do the modification

>However, we still need to change the sh script, are you still wiling to do this?

Yes of course

@akasolace  What is the dead line for the release ?

akasolace:
> Ideally only for Linux user, using packager. If you want I can do the modification
I  don't think it is feasible to identify the one who installed HO via a packager...
Additionaly, why should we treat them differently? The one who installed HO via .deb or .rpm and then update it via HO menu will face the same "issue", or am I missing something?

> Yes of course
What is the dead line for the release ?
Cool ! there are no deadlines really, we will release when ready. At the moment we have only 4 or 5 open bugs for 1.436, so still feasible to have a beta before end of week but it looks more probable that we will release first beta after the holiday period (personally I won't be able to do anything on the project during this 2 weeks holiday break)

yaute74:
>I don't think it is feasible to identify the one who installed HO via a packager...
Additionaly, why should we treat them differently? The one who installed HO via .deb or .rpm and then update it via HO menu will face the same "issue", or am I missing something?

I think I was not clear. For me "packager" is when you install via a packager (deb, rpm, arch, etc ...).
So we should treat differently install with "packager" and install with "zip".
I have an idea to do that.


I ask for the dead line because I search an alternative to InstelliJ On my old home laptop it is very very slow. So I have the time to test lightweight IDE :)

akasolace:
> I think I was not clear. For me "packager" is when you install via a packager (deb, rpm, arch, etc ...).
So we should treat differently install with "packager" and install with "zip".
I have an idea to do that.

hmmmm, for me it is the opposite: 
1) for users updating by downloading .deb, .rpm or vai package manage => no warnng needed
2) for users updating via HO GUI => warning needed only for Linux user

don't you agree?

> So I have the time to test lightweight IDE :)

lightweight Java IDE ... will be tough to find :-)

btw, you are welcome to join the dev team in discord:   https://discord.gg/CAE4Un6

yaute74:
>for users updating via HO GUI => warning needed only for Linux user

Yes but only if we have install via deb or rpm.

>lightweight Java IDE ... will be tough to find :-)

I test the same that I use at work for the c "vs code".

> https://discord.gg/CAE4Un6

I cannot use vocal because my child are sleeping when I work on HO :(

yaute74:
Modification done for issue #35

yaute74:
@akasolace  it's not the case with this line ?
`if [[ "$LIBVERSION" > "$HOMEVERSION" ]]`

akasolace:
@yaute74 
weired yesterday when I looked at it, I saw the old version with '=' instead of '>'
I will merge it master

Issue #42, opened 2018-12-15
wrong data player export by Tool - csv export player


mistermax80:
some data are wrong

akasolace:
closed by #51

PR #51, opened 2018-12-17
Fix for issue #42

mistermax80:
wrong data player export by Tool - csv export player #42
Now:
- Column name are translate
- Text value has double quote, and numerical value no


akasolace:
thank you

PR #62, opened 2018-12-21
Update README.md

brokenelevator:
made dependency licence details into links


brokenelevator:
It would be nice though if the files were actually there :)

PR #64, opened 2018-12-26
Fix for issue#15

mistermax80:
Fix for issue#15

In the HT copy function is not present the info
- Deadline
- Price
- Mother Club

When HT get out this information, it will need update the HO!.


mistermax80:
Hi @akasolace , @brokenelevator , @DavidatorusF 
I finished.
Now there is one only button, for both way.
I saw HT added the mother club info, and I added the support to catch this information.
Thank you to all!!!

mistermax80:
I correct the wording as suggest @brokenelevator 

Thanks!!!!

Issue #44, opened 2018-12-15
In player overview and in line up, mark player played in week


mistermax80:
new feature

akasolace:
are you imagining something like a new column
   - player on the linup for next amical (resp cup, league) has a faded amical (resp cup, league) icon
   -  player who played last amical (resp cup)  has a plain amical (resp cup) icon

is it correct ?

mistermax80:
Yes, You  read me in my mind

akasolace:
@yaute74 I hope you are doing well. 
In case you want to stay in touch with HO, I would suggest that bug to you.
Next release not expected before July so really no pressure here, I just want to make sure you stay active and we are not loosing you :-)

akasolace:
I guess the difficulty might be to handle game since last training as it differs per region.
I am not sure if this information is already available in HO and what is available though CHPP....

brokenelevator:
What if we just add 'week start' information to leagues(countries) table? I think my updates are Friday after 2 AM. If we gather all the times/dates we can just count if player played a (relevant) game (and position) since last 'week start'. We can even show what effectiveness (full/half/whatnot) and what position/skill.

I don't think the information is available in HO and don't know if it is in CHPP.
On second thought, HO knows who to give training update. If it didn't we wouldn't have subskill updates. So maybe just use the knowledge to inform us in advance. Unless it 'subskill recalculates' it.

I guess this needs some research.

akasolace:
@brokenelevator I was thinking it might be saved in the database because in my memory HO was increasing decimal of trained skills, however when I just checked I realized I have lost all my decimal information just as @mistermax80  reported a while ago.

akasolace:
@brokenelevator otherwise I just checked and that information is available via CHPP

```
TrainingDate : DateTime
The date and time when the next training update will start in this league.
```

akasolace:
@brokenelevator in HO the info is in table XTRADATA
column TRAINING DATE

akasolace:
but I also just notice my table TRAINING is empty
I guess it a bug ..

brokenelevator:
It appears that we have the data so this should not be much of a problem to show the information in advance.

About the sub skills they must be retained somewhere.
I have sub skills. Both in HO and DB.
[...]

PR #117, opened 2019-04-29
issue #44 training preview

yaute74:
issue #44 training preview


src/main/java/core/db/DBUpdater.java
```diff
@@ -641,6 +647,37 @@ private void updateDBv25(int DBVersion, int version) throws SQLException {
 		}
 	}
 
+	private void updateDBv26(int DBVersion, int version) throws SQLException {
```
akasolace: @yaute74 latest stable was using `DBVersion = 24`
Hence, `DBVersion = 25` and corresponding `updateDBv25()` are supposed to ensure transition from HO 1.436 to HO 2.0
Can't you merge your code within `updateDBv25()` instead of introducing a new DB version ?   
@brokenelevator what do you think ?

src/main/resources/sprache/French.properties
```diff
@@ -1455,4 +1455,10 @@ ls.club.arena.income.terraces = Income Terraces
 ls.club.arena.income.basicseating = Income Basic seating
 ls.club.arena.income.seatsunderroof = Income Seats under roof
 ls.club.arena.income.seatsinvipboxes = Income Seats in VIP boxes
-ls.club.arena.income.total = Income Total
\ No newline at end of file
+ls.club.arena.income.total = Income Total
+
+#Training preview
+trainpre.fulltrain=Entrainement complet
```
akasolace: @yaute74  Translations should be done via PoEditor. Other files are generated on the fly at compile time so manually edited translation files would get lost.

src/main/java/core/db/DBUpdater.java
yaute74: I check with my database in 1.436.
The dbversion is 24 and the column MATCHREPORT, HEIMHATSTATS and GASTHATSTATS in MatchDetails exist.
At the end of update we do a `dbZugriff.saveUserParameter("DBVersion", DBVersion - 1);`  so  updateDBv25 set dbversion to 24..
So it's why I create a new update.

src/main/resources/sprache/French.properties
yaute74: We must modify only the english version ?

akasolace: yes correct

src/main/java/core/db/DBUpdater.java
akasolace: the `dbZugriff.saveUserParameter("DBVersion", DBVersion - 1);` is only applied if `HO.isDevelopment() == True`.  It allows for devs to test DBUbdaters during the course of development

akasolace: so basically if you are on test, everytime you launch HO.
it goes through the update (in order to simulate) but does not bump version

akasolace: > the column MATCHREPORT, HEIMHATSTATS and GASTHATSTATS in MatchDetails exist.

correct, this is because @brokenelevator fixed a mistake that was made when I created updateDBv24

> MATCHREPORT was made longer here but not in table definition to fix this mistake and not repeat code updateDBv24 is falling through to updateDBv25

yaute74: ok I misread. I will correct it

Issue #84, opened 2019-01-15
[FEAT] use icon / splashscreen / headerbar  according to the version used


akasolace:
![image](https://user-images.githubusercontent.com/45043004/51177599-55931200-18c0-11e9-8a52-da448addd17b.png)

_Originally posted by @DavidatorusF in https://github.com/akasolace/HO/issues/73#issuecomment-454359423_

yaute74:
I can work on it.
Do we change the current logo? Because it's old-fashioned.
You can create logos and do a survey on the hattrick forum.

akasolace:
@yaute74   ok great!
Yes new logo would be very nice. 
Do you fell like asking it on HT forum? Otherwise we go for  it ourselves, plenty of free logo generators those days. We can also do both: ask for contribution and meanwhile give a try ourself.

akasolace:
@yaute74 I had a look on the dedicated forum and indeed it might worth to seek for contributions.
On a side note the HT user guvidex who made some very nice logo already accepted to help. You might also reach out to him if you want

yaute74:
Where is the dedicated forum ? [post=17237456.567] ?
You already contact him ?
I think we can ask and try ourself and after do a survey if necessary

When we talked about it, I made two of them:
![logo1b](https://user-images.githubusercontent.com/34601782/56987805-bce28280-6b8e-11e9-8109-f1ebd695df8f.png)  ![logo2b](https://user-images.githubusercontent.com/34601782/56987811-bfdd7300-6b8e-11e9-9936-beae1d252f3d.png)

akasolace:
@yaute74 
> Where is the dedicated forum ? [post=17237456.567] ?
yes

> You already contact him ?
yes

> I think we can ask and try ourself and after do a survey if necessary
Agreed

> When we talked about it, I made two of them
I will also git it a try but I am very bad at it...

yaute74:
I do the request on hattrick forum: [post=17237456.582]

akasolace:
> I do the request on hattrick forum: [post=17237456.582]

I just saw it ... perfect now lets hope we will have some nice contribution

yaute74:
>should you not ask for splashscreen contribution at the same time ?

@akasolace I wanted to do it after, but maybe it's a bad idea. I wanted it to match the logo we would have chosen.
I create 4 more logos and I receive 3 from contributor :)
I wait your logo :p

akasolace:
> I create 4 more logos and I receive 3 from contributor :)
[...]

PR #129, opened 2019-05-16
issue #84 logo depending version

yaute74:
issue #84 : color depending of version
- I cannot test windows and mac
- Don't hesitate to modify what I did in gradle


akasolace:
I review the PR, it looks good to me but I will test it locally before merging it into master.
I will do that tonight.
Cheers

Issue #105, opened 2019-03-19
[eFEAT] Implement support for extra subs


akasolace:
- [x]  1. introduce new subs in lineup tab + update GUI for better display
- [x]  2. fix lineup assistant to handle extra subs
- [x]  3. fix download lineup from match tabs
- [x]  4. update chpp file version of matchorders
- [ ]  5. fix upload lineup with new chpp version [requires CHPP fix from HT devs]
- [x]  6. fix download lineup with new chpp version
- [x]  7. remove swap button for subs2 as suggested by @DavidatorusF 
- [ ]  8. final testing

Implementation errors:

 - [x] 1. backup subs can only be filled by player already set as subs for a different position (tbc for extra)

akasolace:
I have done step 1 and 2. Still a lot of things are broken but I will do a first commit as there is alreday a lot of code change. Otherwise future mere could be very complex.

akasolace:
currently on hold until fixes are made on CHPP side

PR #130, opened 2019-05-18
Fix the exception throwed by opening the Experience tab of the PlayerAnalysis module.

konved:
Fix the exception throwed by opening the Experience tab of the PlayerAnalysis module.

It reverts a couple of excessive changes which get lost in the immense commit akasolace@8540ad6

Related to bug https://github.com/akasolace/HO/issues/105#issuecomment-475136413


akasolace:
Indeed, I was a bit too heavy with code refactoring that day :-)
Thank you for the commit (and for joining the project.)

Issue #132, opened 2019-05-25
[eBUG] Linenup tab: remove broken/useless buttons


akasolace:
Remove those buttons and all the related code:

- "show only the lineup" 
- "show only the lineup minimzed"
- "print lineup"

Is there any objection against ? 
@yaute74 @brokenelevator  @mistermax80 @DavidatorusF

mistermax80:
For ok to remove... But I don't know what the other users think... My suggestion is: remove button and wait the users reactions...

brokenelevator:
Like I already mentioned I would fix or leave them as they are from the same reason as @mistermax80. Removing the button to gauge reaction also works I guess.

akasolace:
@mistermax80 with git it is cleaner to remove it all in a single commit
if we regret we uncommit at any time

mistermax80:
Ok good!

akasolace:
closed by 7b999a6

PR #144, opened 2019-06-10
issue #132

yaute74:
issue #132: remove buttons 'print this line-up', 'show only the line-up' and 'only display line-up minimized', corresponding code and image


akasolace:
closes #132

Issue #149, opened 2019-06-15
[eBUG] ho don't start with deb package


yaute74:
ho don't start with deb package
correction #102 was not in stable version

PR #150, opened 2019-06-15
issue #149

yaute74:
Fixes issue #149
Maybe it would be better to rename ho.sh in ho.bash but it requires more modifications


konved:
I do not like this.
Many *nix systems **do not** contain `bash`. At all.
Even debian and its clones with their `dash` as default shell.
Meanwhile `/bin/sh` is always present everywhere (mainly as a link to some advanced shell).

PR #152, opened 2019-06-16
issue #148

yaute74:
Fixes issue #148
Exception when we started with an empty db


akasolace:
@yaute74 funny, when I looked at the trace it did not occur to me that it was linked to this.

Issue #133, opened 2019-05-25
[eFEAT] Lineup tab: better use of space to improve visibility


akasolace:
Option 1 ) The idea would be to have different width for the player box for player on the field and player set as substitutes. Player box on the field would be wider to expend as much as the substitutes lines, it would allow to have better name visibility. =>  It might be tricky to be done.

Option 2) Another possibility much simpler to implement would be to remove the shirt from the player box. I don't think it brings anything and it would give more space to player name display.
 
@brokenelevator @yaute74 @mistermax80  @DavidatorusF  Let me know what you think

brokenelevator:
@akasolace 
1) The substitutes don't even fit on my screen, so that is a problem but I can live with that. Aside from that I see no problem with option 1.

2) I kind of used to having shirt there but to be honest it seems quite useless. So option 2 seems like a reasonable idea. It won't give much space though. We can remove it along with numbers but those are not completely useless.

akasolace:
#146 can be dealt had the same time

akasolace:
closed by #175 (thank you @yaute74 !!)

PR #176, opened 2019-07-18
issue #133 size of combo in lineup

yaute74:
Small modifications for issue #133 
- reduce space between select player and other players
- add shirt on the same line than player name   

For "no player previously selected, there should be no blank", it work like this before I don't touch to it.
For the player name (full name, short name, etc ...) I have no opinion, I leave you the modification you want ;)


akasolace:
> For "no player previously selected, there should be no blank", it work like this before I don't touch to it.
ah ok I guess it was less visible before

Issue #182, opened 2019-07-27
[eBUG] Not correct building if it is launched from a branch


pierpaolocira:
This happens if you have a branch, e.g. HO-xxx and try to build the MacOSX executable.
Then into ./build/macApp you'll find both:
- HO-xxx.app - containing all the effective content for the application (I guess created this way by the macAppBundle plugin)
- HO.app - containing only the file changelog.html (this is manually done in the task `AddResourcesAndZipMacApp`)

This should not be problematic because I imagine release are preparred from the main (only `HO` branch), but build.gradle should be fixed in order not to propagate the issue in case of new features based on the same gradle code.

akasolace:
@pierpaolocira,  I guess the idea is to fix AddResourcesAndZipMacApp and to copy file in HO-xxx.app, right ?  Do you have an idea on how to do this?

pierpaolocira:
> @pierpaolocira, I guess the idea is to fix AddResourcesAndZipMacApp and to copy file in HO-xxx.app, right ? Do you have an idea on how to do this?

@akasolace no, as tolt I'm not an expert with Gradle... but if noone else has time for it, I can try to solve. Please, let me know.

akasolace:
closed by #193

Issue #166, opened 2019-07-06
v 2.0 1256 doesn't start on mac


mistermax80:
![Schermata 2019-07-06 alle 08 40 13](https://user-images.githubusercontent.com/15519907/60752775-f2aa4a00-9fc9-11e9-8a30-6b1cd0ffd003.png)
[HO-2019-07-06.log](https://github.com/akasolace/HO/files/3364480/HO-2019-07-06.log)

akasolace:
08:38:56 [Error]   core.model.FormulaFactors: Error while loading prediction/defaults.xml
Under mac prediction/defaults.xml is not found. I had to change that part recently, I tested it with zip, and windows exe but not with mac.
@mistermax80 can you do a step by step debug and tell me more about it ?

akasolace:
@mistermax80  I do not have access to a mac anymore so I really rely on you for this one.
I can't do it myself but I am happy to give you advice if you are struggling.

akasolace:
@pierpaolocira  maybe you can help @mistermax80  on this one.
One way to start is for example to look at how the prediction files are loaded and do the same within a ifmac clause.

pierpaolocira:
@akasolace OK, I should be able to take a look in some days... but I have some trubles in reproducing.
(Maybe stupid questions, but I have no experience both with this project and gradle - I'm a mvn fan :) )

From the screenshot it seems that it happens at startup from a compiled version, but the `:createDmg` task fails to me if I launch `gradle build`.
Indeed, if I launch it from the IDE I have no troubles.

Should I do something specific in order to build? I have a:
```
Execution failed for task ':createDmg'.
> Unable to determine image size with sip
```
I think there is something wrong with the path, because it searches the image file in
`/path/to/project/HO-pierpaolocira/build/macApp/.background/HO-pierpaolocira\buildResources\Mac\Back.png`
instead of
`/path/to/project/HO-pierpaolocira/build/macApp/.background/Back.png`

I launch the command from: `/path/to/project/HO-pierpaolocira/`.


Or, maybe, there is a quick way to reproduce by launching from the IDE?

akasolace:
@pierpaolocira Indeed from the IDE it will work but fails from the app because it will look the ressource in the home folder instead of looking at it within the prediction folder within the jar.

I don't think building the app will help in closing this issue, however if you want to build it, I suggest you create a minimal task with only the necessary steps. maye something like (I cant' test) :

task makeMacAPP(type: GradleBuild) {
[...]

PR #193, opened 2019-08-08
Issue-182: Not correct building if it is launched from a branch

pierpaolocira:
 
   - fixes issue #182 : Not correct building if it is launched from a branch
 
2. changelog and release_notes: do not require update

3. @akasolace I had to move the `def` into `doLast` in order to be able to use parameters from the `macAppBundle` plugin. I hope this is OK (as told, I'm new with Gradle)
I would ask you to check if a standard (full) build works, before accepting the pull request.
I only checked it on Mac (task `makeosX`) and it works fine (i.e. a single `.app` is created that crashes exatly as before the change, that is what is reported in #166).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -513,19 +513,20 @@ task ZipDistribution(type: Zip) {
 
 task AddResourcesAndZipMacApp {
     dependsOn 'init', 'createApp'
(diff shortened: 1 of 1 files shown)


akasolace:
I will test and merge it this weekend. Thank you

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #244, opened 2019-10-04
[eFEAT] Activate feedback plugin only if data recently refreshed


akasolace:
In order to gather only quality data, we will activate feedback plugin only if data recently refreshed

1) add label "Please make sure your players data are up to date before participating" in the feedback plugin label
2) store somewhere the timestamp at which update has been made (maybe already exist) and add as an additional check before activating "send" button, something like now()-RefreshTimeStamp < 1 h
3) Add corresponding message as trasnlated string i.e if button remains deactivated because of no recent Refresh,. Users need to be informed of what is the issue (similar to other checks)

mistermax80:
Today the server is unavailable... The free hosting service is unavailable...
This situation can be matter with any free hosting, but I reflected on this.
We should decide if the feedback plugin is active or ready to receive data. By json from gitub, we can do this, add other 'request' that HO uses for don't activate the feedback plugin... We can use this for manatainace server or similar situations matter today...
Other this we can insert in gitub json the URL server, this for We take a way to change URL without give an new update.
Let me know what you think.

akasolace:
> By json from gitub, we can do this, add other 'request' that HO uses for don't activate the feedback plugin... We can use this for manatainace server or similar situations matter today...

We can either, do that check beforehand or capture that error in a try/catch and indicate correct message like "Server currently not available, please try later." However, I am not sure what you mean by `by json from github`? 

> Other this we can insert in gitub json the URL server, this for We take a way to change URL without give an new update.

I am also ok with that idea but hopefully this is not something we will need to use anytime soon.

mistermax80:
For json git I mean https://akasolace.github.io/HO/feedback.json
, In this object we can insert if the feedback plugin is active or no, and the specific text for the communication. I think the effect the yesterday issue is an time out of the request... Adobe the users can see as in error don't managed from us.
In this moment th release 2.0 is waiting broken development... So we can try to insert these as options.

akasolace:
> In this object we can insert if the feedback plugin is active or no  

I am not sure I understand,  You want to hardcode in the json whether or not the server is online? Based on what? When/How the json will be updated ?

mistermax80:
I would like insert an information in these object in order to say to HO if the feedbackplugin is active or no.
The update json object will be manual, as when you change for change the request from GK to CD-normal.
[...]

PR #252, opened 2019-10-12
fix for #244:

mistermax80:
- if data old 5 minutes, an message appear in order to ask download data
- logic for read "server_status" in json
- logic for url server (param: "server_url") in json (in order to overwrite the current default)

@akasolace, If you think 5 minutes are many or not, we can change the value, or we can think to include this value in json (in order to have control on this param without make new release).
Let me do you think.


akasolace:
I would put older than 60 minutes ...

mistermax80:
Yes, I reflected only the case the user send feedback some minutes after training effect... So 5 minutes it should limit this.

Issue #266, opened 2019-10-27
[eBUG] Missing Download match orders


mistermax80:
I set an order for an player who will be substitute to 85 min if the match is in advantage for more 1 gol.
This order in HP doesn't download.
HT:
![image](https://user-images.githubusercontent.com/15519907/67643107-c9644f80-f913-11e9-9c0c-054276ee57e3.png)

HO:
![image](https://user-images.githubusercontent.com/15519907/67643098-ac2f8100-f913-11e9-8621-da9eedeb52e2.png)

DavidatorusF:
@mistermax80 
what is the position of the player ? on the field and as substitute ?

mistermax80:
GK

DavidatorusF:
work for me... if it's the next match after updated my team

@mistermax80 
lineup -> upload/download tab; select the match and [download lineup], do you see your order ?

mistermax80:
@DavidatorusF no, it doesn't work...
I made test from lineup -> upload. I tried in italian, english and french, same result.

mistermax80:
I found the issue... if in HT you don't set the captain and kicker, the match orders will not download...

akasolace:
closed by c1ab6b80376f1ab7190e14943e86863529962959

akasolace:
once BETA 2 is released, we will test issue is fixed before closing it.

PR #270, opened 2019-10-28
fix for #266

mistermax80:
fix for #266

### DIFF
--- a/src/main/java/core/file/xml/XMLMatchOrderParser.java
+++ b/src/main/java/core/file/xml/XMLMatchOrderParser.java
@@ -488,9 +488,15 @@ public class XMLMatchOrderParser {
 			}
 
 			// Treatment of SP taker and Captain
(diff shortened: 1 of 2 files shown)


mistermax80:
@akasolace 
The code goes in exception, but there is the try /catch block that doens't show the error.
Would you check if we want leave asis?
core/file/xml/XMLMatchOrderParser.java in line 509

akasolace:
@mistermax80  if I understand correctly, before the code was going into exception and was logged (hence user could not see match download was not complete). With what you did, it fixes #266, but if something else is not catch we will have same behaviour (i.e. logged but no error generated), right?

mistermax80:
> @mistermax80 if I understand correctly, before the code was going into exception and was logged (hence user could not see match download was not complete). With what you did, it fixes #266, but if something else is not catch we will have same behaviour (i.e. logged but no error generated), right?

Yes, You understood.

Issue #269, opened 2019-10-28
[eBUG] Unable to Set Future Trainings


Synthespian:
**Describe the bug**
On opening the the application for the first time, the future training was set to "Defending".
This meant the wrong players were shown to have received training.  On trying to change the "Future Training" I got an error report.

**To Reproduce**
1. Open "Training Tab"
2. Click on "Set All" (Future Training)
3. See error

**Screenshots**
![image](https://user-images.githubusercontent.com/48737160/67721882-8a93cf80-f9cf-11e9-9a6b-918816f4f730.png)

**Platform information:**
 - Windows 10
 - v2.0

**Error Content**
setSelectedIndex -5 out of bounds
java.lang.IllegalArgumentException: setSelectedIndex: -5 out of bounds
	at javax.swing.JComboBox.setSelectedIndex(Unknown Source)
	at module.training.ui.comp.FutureSettingPanel.jbInit(FutureSettingPanel.java:80)
	at module.training.ui.comp.FutureSettingPanel.<init>(FutureSettingPanel.java:46)
	at module.training.ui.TrainingPanel$1.actionPerformed(TrainingPanel.java:84)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.processMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.dispatchEvent(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Window.dispatchEventImpl(Unknown Source)
[...]

PR #271, opened 2019-10-28
fix for #269

mistermax80:
fix for #269


src/main/java/module/training/ui/comp/FutureSettingPanel.java
```diff
@@ -77,7 +77,11 @@ private void jbInit() {
         intensity = new IntensityComboBox(0);
         intensity.setSelectedIndex(firstFutureTraining.getTrainingIntensity());
         staminaTrainingPart = new IntensityComboBox(10);
-        staminaTrainingPart.setSelectedIndex(firstFutureTraining.getStaminaPart() - 10);
+        int staminaPart = firstFutureTraining.getStaminaPart();
+        if (staminaPart - 10 < 0){
```
akasolace: why not simply `if (staminaPart < 10) {staminaPart = 10;}`

mistermax80: it is only logical approach :D

Issue #234, opened 2019-09-30
[eFEAT] Training week preview in Training Tabs


mistermax80:
Continuing the discussion started on #222 
Insert in training tabs the information on training on week (icon and tooltip).
The final effect it will be as following:
![image](https://user-images.githubusercontent.com/15519907/65895415-b8bfc880-e3ab-11e9-9098-13794ba8f195.png)

In training tab (asis function) the colors are 2:
- green for player in position that receive all training
- yellow for player in position that receive partial training
for example in case of cross traiinig:
- green for wings
- yellow for back wings
this logics consider only position and future training and these don't see the lineup.

The yaute logic do:
green (ligth/or no light): for player received total training in terms of time
grey: for player will receive training in terms of time.

I would like see in the tab training both information. My idea is use yaute logic for to show the icon and tooltip, and leave current logic as is.

The yaute logics is an static function, and it is usable in each part of code.

@yaute74 , sorry, I would like ask you, explain to @akasolace that I say. Could you help me?

yaute74:
In principle I agree. I think you can use the same object than me for the user column and leave the rest as it is.

Just a remark, aren't you afraid it'll give too much information to the user ?

yaute74:
@mistermax80  you must use SpielerLabelEntry object

mistermax80:
@yaute, yes... Respect to too much information to user, if you think this I can think to insert a preference for this.
I can say I will use the all information. I will have these information in only screen it is comfortable!

yaute74:
@mistermax80  ok great :)

mistermax80:
Can it useful insert a new column for to show the only skill value that receive train?

akasolace:
>Can it useful insert a new column for to show the only skill value that receive train?
the user knows which skill he is training so I am not sure this will be very useful. Addiitonaly the tab is already quite packed

mistermax80:
Ok I agree!

mistermax80:
@yaute74 , I tried to do it, but I have some difficult.
When you can, can you help me on this? We will for version 3, so nothing hurry.

mistermax80:
I don't able to set the render cell in order to build a cell name with training icon and player name.
So when I update the table rendering, in the firts cell appear the java object name and not icon+player name.

yaute74:
@mistermax80 ok I will check when I have the time...

yaute74:
[...]

PR #300, opened 2019-12-02
Feature for #234 training tab

mistermax80:
PR ready!


src/main/java/module/training/Skills.java
```diff
@@ -16,30 +16,33 @@
     public static int getSkillAtPosition(int position) {
         switch (position) {
             case 0:
-                return PlayerSkill.KEEPER;
```
akasolace: Have you used same definition as per the CHPP API ? I mean is 0 really for form and not for keeper ?

mistermax80: @akasolace that method is for order in list detail panel "0" is refer to first position in following screen (after was keeper, after this PR is form):
image

akasolace: ok perfect thank you

Issue #306, opened 2019-12-06
[eBUG] No training for secondary cups 


yaute74:
Training is not taken into account for secondary cups.

akasolace:
same as #278

akasolace:
If it is a small one I will cherry pick for 2.1

akasolace:
closed by 75a5a5c

PR #307, opened 2019-12-06
#306 training secondary cups

yaute74:
issue #306


akasolace:
@yaute74 I am wondering if we should not instead fix the function `isOfficial()` in MatchType.java and use it in Training manager otherwise we have duplicated pieces of code ....

Issue #124, opened 2019-05-10
[eFEAT] Sort by training preview


yaute74:
Could be great to sort table by training preview  (issue #44)

akasolace:
you would do that by adding a column to the table ?

yaute74:
I don't know, new column, popup menu, button in header.

yaute74:
@akasolace  do you think something like this:
![image](https://user-images.githubusercontent.com/34601782/57615463-0910c800-757c-11e9-8769-08fa07491185.png)

With: training (estimate training)
And color depending 50%/100% position: color green/greelight (black/grey)
But I am not satisfied with the rendering, the color is not very visible

Question if we do that:
- column in which position ? last after column ?
- we remove icon and tooltip (I add for training preview) in player name column ?

akasolace:
Hello @yaute74
  I would rather stick with what you did initially.
Hence, I suggest,  we add a very thin column on the left where we have the training indicator (i.e. the green bar, would be in that new column instead of the player name column). We could sort by training in clicking on that new column header. 
I would keep the infobull and we might add a new info bull for the column header.

What do you think ?

yaute74:
@akasolace I will check this solution

mistermax80:
link also: #234

yaute74:
@akasolace I will propose another solution, manage a third click on the name column:
- first click sort by name ascending
- second click sort by name descending
- third click sort by training preview
- fourth click by name ascending
- etc ...

Add a column in first position is an important modification compare to the functionality and it will take more space.

akasolace:
Sounds a good idea, does the job and implementation is simple. Only issue is that users won’t be easily aware of that functionality but this is no big deal ... 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mistermax80:
yes good idea: please reminder also the training tab: #234 😄 
![image](https://user-images.githubusercontent.com/15519907/70064203-68522a80-15e9-11ea-9ff5-d45ef660d4e0.png)

yaute74:
@mistermax80 I do PR but I don't treat the training page because it's sorted from an invisible column and I don't want to broke something, if you want to do it....

akasolace:
closed by afd64be

mistermax80:
> @mistermax80 I do PR but I don't treat the training page because it's sorted from an invisible column and I don't want to broke something, if you want to do it....

Only default sort is on hidden column... Ok... I will do... 😭

[...]

PR #318, opened 2019-12-15
#124 sort by training preview

yaute74:
#124 sort by training preview

Maybe it's not the best way to do it, I'll let you have a look.
I don't do it for training page because it sorted frlom an invisible column


akasolace:
closes #124

Issue #330, opened 2019-12-28
[eBUG] MatchDayPanel download Button is not enabled when match is over


wsbrenk:
After downloading (F11) i cannot download the matches of the other teams in my series.

![matchdaypanel](https://user-images.githubusercontent.com/11821502/71548629-81b08180-29b1-11ea-9965-9b5b1eb2f087.jpg)

**Platform information:**
 - OS: [Windows]
 - Version 2.1r1463

**Additional context**
setMatchButton was changed on 03-10-2019, maybe in 3 hours i will be able to download the matches

wsbrenk:
after a waiting time of 3 hours i was able to download the  other matches.

akasolace:
@wsbrenk there is no bug then, right? should we close that issue?

wsbrenk:
i do not want to wait 3 hours until i can download. the version before october could download directly after the match is finished.

akasolace:
@wsbrenk ok I guess you can reduce it however keep a cushion in case of extra time.
You can make the same change in src/main/java/module/matches/SpielePanel.java

wsbrenk:
@akasolace, @mistermax80: do you know, why the waiting time was introduced? the earlier version only check if the goals were available to decide if the match was finished. did that no longer work?

akasolace:
@wsbrenk this was introduced when we fixed  #181 and #237
Maybe this change (3 hours dealy) was not mandatory and there are clever ways to do it ...

wsbrenk:
@akasolace: that could be a welcome distraction from the feature 299. I suspect that the goals are already displayed during the game. According to chpp, however, the finished_time should be a useful means of recognizing the end of the game.
if i attack this ticket, should i branch from master or better from release/2.1 ?

akasolace:
> if i attack this ticket, should i branch from master or better from release/2.1 ?  

you should branch from master. Tonight,  I will update the branches and prepare a dev release.

wsbrenk:
I read from the referenced tickets that there were problems if the matches in the DB were not updated for a long time. I found a very simple solution for this, but we won't be able to test it until next Saturday.

akasolace:
> but we won't be able to test it until next Saturday.

In case you are planning to do some testing while you have some ongoing game, maybe you can have a look to #291 at the same time... should be a very quick one

akasolace:
[...]

PR #334, opened 2019-12-30
MatchDayPanel enhanced game finished decision

wsbrenk:
 
   - fixes issue #330  
    
 
 - [ ] do not require update


akasolace:
I did not know `hatStattgefunden`. 
Also as a side remark, usually we (at least I) try to translate method/comment from German to English when I update part of the code which is concerned. In case you can participate to the effort it could be nice ... cheers !

Issue #337, opened 2019-12-30
[eBUG] MatchDayPanel NullPointerException


wsbrenk:
![MatchDayPanel](https://user-images.githubusercontent.com/11821502/71600786-d0e0e880-2b50-11ea-8917-b31f510c1555.jpg)

akasolace:
@wsbrenk can you tell me more about that issue: 
    - how does it occurs?
    - is it related to #291?

wsbrenk:
this is my embarrassing error that I built into the MatchDayPanel. It happens when you call up the Matchday panel.

akasolace:
closed by 4b79646

PR #338, opened 2019-12-30
Err#330

wsbrenk:
 
   - fixes issue #337

### DIFF
--- a/src/main/java/module/series/MatchDayPanel.java
+++ b/src/main/java/module/series/MatchDayPanel.java
@@ -122,8 +122,9 @@ final class MatchDayPanel extends JPanel implements ActionListener {
             gameFinishTime = gameFinishTime + 3 * 60 * 60 * 1000L; //assuming 3 hours to make sure the game is finished
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/module/series/MatchDayPanel.java
```diff
@@ -117,13 +117,15 @@ private void setMatchButton(JButton button, Paarung paarung) {
...
+        boolean gameFinished = false;
         if (paarung != null) {
             gameFinishTime = paarung.getDatum().getTime();
             gameFinishTime = gameFinishTime + 3 * 60 * 60 * 1000L; //assuming 3 hours to make sure the game is finished
+            gameFinished = paarung.hatStattgefunden() == true || gameFinishTime < nowTime;
         }
 
         // Hat das Spiel schon stattgefunden
-        if ((paarung != null) && gameFinishTime < nowTime) {
+        if ((paarung != null) && gameFinished) {
```
akasolace: I don't see how this fix #337?  At first look the restriction seems even less restrictive than before ...

PR #353, opened 2020-01-18
Fix osx build issues

tychobrailleur:
1. changes proposed in this pull request:
 
  This PR fixes a few issues preventing building the project on macOS with Gradle, and also updates the wrapper to 5.6.4.  The various commits comments explain the rationale for each change.
 
2. changelog and release_notes ...

 - [ ] have been updated
 - [x] do not require update — but happy to be convinced otherwise!

3. [Optional] suggested person to review this PR @___

@akasolace

PR #360, opened 2020-01-22
Fix macos zip generation following #353.

### LINKED ISSUE #359: [eBUG] bumping version of gradle has broken some task(s)
**Describe the bug**
Gradle script needs to be fixed in order to reproduced similar output has 2.0.
`osXjar` task is broken, others looks ok

@tychobrailleur, can you please work on this one? 

Thank you

tychobrailleur:
The `execute` method was not called any more on the nested task in
`AddResourcesAndZipMacApp` so the nested task was not being executed
any more.  Digging deeper, it seems like `execute` does not exist
more.

The workaround is to split the task into tasks, one depending on each
other.  This fixes #359.

   - fixes issue #359 
 
 - [ ] have been updated
 - [x] do not require update

@akasolace — would you be able to verify whether this fixes on your side?  Let me know if you're still having problems...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -163,6 +163,7 @@ distributions {
 
 // Configure macAppBundle task =======================
 macAppBundle {
+    appName = 'HO'
     mainClassName = "HOLauncher"
     icon = "${project.ext.osx_app_dir_sourcedir}/HO.icns"
     bundleJRE = false // true #issue-198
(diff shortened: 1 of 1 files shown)


akasolace:
@tychobrailleur  I actually started to work on refactoring build gradle yesterday.
I fixed couple of other issues including #359
Thank you for the help !

Issue #389, opened 2020-02-13
make build.gradle platform independant


pierpaolocira:
Hi @akasolace , 
I am working on 372 and I wanted to test the behaviour when the application is finally packaged as an .APP

But any attempt I do in order to build, I receive the same issue with gradle:
```
Evaluating root project 'HO-pierpaolocira' using build file '...build.gradle'.

FAILURE: Build failed with an exception.

* Where:
Build file '...build.gradle' line: 6

* What went wrong:
An exception occurred applying plugin request [id: 'nebula.ospackage', version: '5.0.4']
> Failed to apply plugin [class 'com.netflix.gradle.plugins.packaging.SystemPackagingBasePlugin']
   > Could not create an instance of type com.netflix.gradle.plugins.packaging.ProjectPackagingExtension.
      > Could not find matching constructor for: org.gradle.api.internal.file.copy.DefaultCopySpec(org.gradle.api.internal.file.BaseDirFileResolver, org.gradle.internal.instantiation.generator.DependencyInjectingInstantiator)
```

Do you have any idea on the issue?

Netflix??? Are you able to replicate? 

It happens also with a simple:
`gradle clean installDist AddResourcesInDistrib ZipDistribution`

Thanks

pierpaolocira:
Additional comment:
I'm sorry to say, but the `build.gradle` file seems to be a disaster now, and it cannot be easily used (maybe with the exception of some computers).

Some comments:
- I hat to change version of the nebula.ospackage plugin
- As usual I had to comment to apiKey line
- I had to empty the githubRelease task
- I had to remove any reference to "AddResources*" (why don't they exist anymore?)
- I had to empty the "resetDB" task because it cannot find "D:\\Program Files\\HattrickOrganizerDev\\db" on my computer
- Still the file is full of references to "D:" , Windows directory separator (e.g. \docs\changelog\md)... that I cannot manually change everytime I get the new version from the repository

To do a PL for the #372 I need to run proper tests with the .APP (IDE is not enough) and I cannot find an easy way... I cannot spend more time to build than to develop and test. :)
Is there any suggestion on how to do?

Some months ago it was working like a charm... how is it possible to have such big regression?

akasolace:
>  hat to change version of the nebula.ospackage plugin

the version specificied in build.gradle work flawleslly for me. I don't know why it is not working for you. You are also running the script using gradlew, not with your local install, right?

> As usual I had to comment to apiKey line

[...]

Issue #372, opened 2020-01-29
[eFEAT] Change database folder on Mac OS X


pierpaolocira:
Under Mac OS X the dabase folder is placed under the home directory of the user.
[edit: both the folder "db" and the file "user.xml"]

I would not recommend to find a way to place it "under" the application folder (because it might be overwritten in case of updates).

MacOSX application data should be placed under "userhome/Library/Application Support/".

If this is not possible, then I would recommend to add anothere level in the direcoty organisation: "hattrick" and to place it in the user folder. In this way final destinations will be "\~/hattrick/db/*" and "~/hattrick/user.xml".

Otherwise there might be conflict with other applications using the folder "db" [or the user.xml file] and this could also be counterintuitive for users that might e.g. delete it by mistake.

In both case we should also include an automatic "migration procedure" that would simply:
- detect if the new location exists
- if not (and the old location exists) to perform the migration.
This procedure could be kept for 2-3 versions and then being removed.

tychobrailleur:
Totally agree with that one, and may even get a crack at it — this has been nagging me for some time. ;-)

akasolace:
@pierpaolocira @tychobrailleur  I fully agree with this

I have been trying (in a vm) to set a diferent location for HO folder, but I did not find a way. This is also why it is not possible to upgrade HO under macOS while it is possible under windows and Linux. 

So if you manage it, I guess it means you will also make possible upgrading HO from macOS, that would be a big one

tychobrailleur:
> So if you manage it, I guess it means you will also make possible upgrading HO from macOS, that would be a big one

Not sure I get what you mean there?

akasolace:
In windows and Linux version of HT, it is possible to do help -> check new version and upgrade directly via the app. If new version exist, it dowload files and upgrade directly, no need to do this manually as it is the case on macOS

![image](https://user-images.githubusercontent.com/1136496/73396767-5edff880-42e2-11ea-9f89-4fd13029568a.png)



![image](https://user-images.githubusercontent.com/1136496/73396662-25a78880-42e2-11ea-882a-e80217dbd90d.png)


![image](https://user-images.githubusercontent.com/1136496/73396705-3c4ddf80-42e2-11ea-9c12-696cf7d1b3e3.png)

tychobrailleur:
[...]

PR #399, opened 2020-03-01
Isue 389

pierpaolocira:
 
   - fixes issue #389  
 
It allows on MacOS systems to build a working .APP application.
All the changes should not affect Win compatibility for main building process, so it should be tested before accepting the PR


build.gradle
```diff
@@ -628,6 +633,10 @@ createApp.dependsOn installDist
 //tasks.getByName('githubRelease').dependsOn compute_sha_sums
 
 
+poeditorPull.onlyIf {
+	project.hasProperty('apiKey') && project.hasProperty('projectId')
+}
```
akasolace: it should be replaced with
`poeditorPull.onlyIf {project.hasProperty('POEDITOR_APIKEY')}`

pierpaolocira: @akasolace just tried again to be sure, but as expected it didn't work (it's not enough as a condition).
I can add it, but I cannot remove the other two (that exists only if previous tasks on POEditor succeded).

akasolace: @pierpaolocira  for me the task PoEditor does not run (i.e. it is skipped) if I let  this condition `project.hasProperty('apiKey') && project.hasProperty('projectId')`. 
Maybe what we need is a way to check that `POEDITOR_APIKEY != ""`

pierpaolocira: > @pierpaolocira for me the task PoEditor does not run (i.e. it is skipped) if I let this condition `project.hasProperty('apiKey') && project.hasProperty('projectId')`

Oh, this was not clear to me, sorry... I hope to go for a PR by today, or tomorrow at most.

pierpaolocira: @akasolace I did't understand it was skipping the task for you. Now I tried both with empty value (skipped) and non-empty value (not skipping, but failing on my case because the Key is no valid, as expected).
Please check and let me know

akasolace: @pierpaolocira sounds promising, I will test it tonight ...

pierpaolocira: Great... I hope it works 100% to you, so I can finally merge it to my branch on #372 and start testing that...

Issue #425, opened 2020-03-24
[eBUG]another team analyzer ArrayIndexOutOfBoundsException


wsbrenk:
java.lang.ArrayIndexOutOfBoundsException: 8
	at core.model.player.Player.hasSpeciality(Player.java:1311)
	at core.specialevents.UnpredictableEventPredictionAnalyzer.analyzePosition(UnpredictableEventPredictionAnalyzer.java:60)
	at core.specialevents.SpecialEventsPredictionManager$Analyse.analyzeLineup(SpecialEventsPredictionManager.java:57)
	at core.specialevents.SpecialEventsPredictionManager.analyzeLineup(SpecialEventsPredictionManager.java:208)
	at module.teamAnalyzer.report.TeamReport.addSpecialEvents(TeamReport.java:99)
	at module.teamAnalyzer.report.TeamReport.addMatch(TeamReport.java:88)
	at module.teamAnalyzer.manager.ReportManager.buildReport(ReportManager.java:52)
	at module.teamAnalyzer.manager.ReportManager.updateReport(ReportManager.java:72)
	at module.teamAnalyzer.SystemManager.updateReport(SystemManager.java:181)
	at module.teamAnalyzer.ui.FilterPanel$2.actionPerformed(FilterPanel.java:170)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.processMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.dispatchEvent(Unknown Source)
[...]

PR #426, opened 2020-03-24
Bug#425

wsbrenk:
 
   - fixes issue #425  
 
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/constants/player/Speciality.java
+++ b/src/main/java/core/constants/player/Speciality.java
@@ -1,7 +1,7 @@
 package core.constants.player;
 	// enum for speciality
 	public enum Speciality {
-		NO_SPECIALITY(0), TECHNICAL(1), QUICK(2), POWERFUL(3), UNPREDICTABLE(4), HEAD(5), REGAINER(6), SUPPORT(8);
+		NO_SPECIALITY(0), TECHNICAL(1), QUICK(2), POWERFUL(3), UNPREDICTABLE(4), HEAD(5), REGAINER(6), _UNUSED(7), SUPPORT(8);
 		private final int value;
 
 		Speciality(int spec) {
--- a/src/main/java/core/model/player/Player.java
+++ b/src/main/java/core/model/player/Player.java
@@ -597,13 +597,12 @@ public class Player {
     }
 
     public String getAgeWithDaysAsString(Date date) {
-        // format = yy (ddd)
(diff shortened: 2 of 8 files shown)


wsbrenk:
this may be a candidate for a changelog entry, since the missing specialty value could cause exceptions in other cases.

akasolace:
@wsbrenk can you please rebase to cleanup the PR ?

wsbrenk:
@akasolace rebase was done. i included the other npe candidates of special event classes.

akasolace:
@wsbrenk  maybe you can do 3 different PR:

- 1 with 5cf5fe7 and 43083ba
- 1 with 5fe054c
- and 1 with c87b818

as they address 3 different bugs, what do you think ?

src/main/java/core/constants/player/Speciality.java
```diff
@@ -1,7 +1,7 @@
 package core.constants.player;
 	// enum for speciality
 	public enum Speciality {
-		NO_SPECIALITY(0), TECHNICAL(1), QUICK(2), POWERFUL(3), UNPREDICTABLE(4), HEAD(5), REGAINER(6), SUPPORT(8);
+		NO_SPECIALITY(0), TECHNICAL(1), QUICK(2), POWERFUL(3), UNPREDICTABLE(4), HEAD(5), REGAINER(6), _UNUSED(7), SUPPORT(8);
```
akasolace: it might have impact on other places ... 
this one will need to be tested or at least we need to check where that enum is being used using ordinal value ...

wsbrenk: this is what i meant- there are several instructions like this one
        Speciality s = Speciality.values()[iPlayerSpecialty];

they all will fail if iPlayerSpecialty is SUPPORTER

i don't find any usage, where _UNUSED would make trouble

wsbrenk:
>@wsbrenk maybe you can do 3 different PR:
>
>   1 with 5cf5fe7 and 43083ba
>   1 with 5fe054c
>   and 1 with c87b818

the first 2 items are following topics from previous tickets. i think they are not worth for extra tickets/prs. But if you want, i will do it.

akasolace:
>  But if you want, i will do it.

If you don't mind I would rather stick to separate commit for separate bugs ... makes things easier down the road ...

PR #442, opened 2020-04-10
Fix an error in personality translations

maurooaranda:
A friend of mine found this error: Personality translations in spanish are inverted.
Agreeability and honesty are mixed, so the display in the player info is inaccurate.

 I don't develop in Java, so I'm really not sure this fixes the problem, but it will be easy for some dev to check it.

Best regards,
Mauro.

Personality attributes agreeability and honesty are inverted, so fix them.

 - [ ] have been updated
 - [X] do not require update

I think this doesn't need an update.


akasolace:
@maurooaranda  thank you for the PR but unfortunately, this is not how it can be fixed.

translations should be fixed using https://poeditor.com/join/project/jCaWGL1JCl  
maybe you or your friend can fix it there,  very appreciated !

maurooaranda:
Ok, no problem.  I've already joined to the project in POeditor, but it told me I should wait for something like an approval.  So I'll wait for that and try to fix it there.

Thanks for taking over this project.

akasolace:
@maurooaranda  you should be approved now ...

maurooaranda:
Ok, thanks.  I've now made the change in POEditor.

akasolace:
@maurooaranda  perfect thank you

Issue #477, opened 2020-05-01
[eFEAT] removal of Match Analyzer module


akasolace:
Selected functionalities will be recovered as part of #476

PR #479, opened 2020-05-01
#477: removed MatchAnalyzer module

wsbrenk:
 
   - fixes issue #477
   
 - [x] have been updated
 - [ ] do not require update


akasolace:
@wsbrenk have you try what this would give for someone having this module currently activated and upgrading to 3.0? Is there not something in the user preference database that should be fixed also ?

wsbrenk:
@akasolace before applying the changes i activated match analyzer module. when invoking ho after the changes the module simply was gone. 
i checked if database tables exists for this module and didn't find any. i forgot the user configuration table. good point. i will check that and also the language strings could be cleaned if some of them were exclusivly used by the  match analyzer.

akasolace:
>  i forgot the user configuration table. good point. i will check that and also the language strings could be cleaned if some of them were exclusivly used by the match analyzer.

@wsbrenk  ok great I wait your feedback on those 2 points before merging, cheers

wsbrenk:
i did not find any exclusiv entries in database tables

akasolace:
@wsbrenk last think is there not some icons that can also be removed, like home.png / away.png, ....
worst case we could clean this later on if you don't want to look at it now. As I guess @tychobrailleur will have a clear view of the used/unused icons while working on the new themes

wsbrenk:
i'll check that

wsbrenk:
@akasolace @tychobrailleur  found 2 icons exclusively used by the match analyzer

PR #487, opened 2020-05-01
fix matchkurzinfo column types for BETA 1 users

akasolace:
follow up of f919b2ecc4cbf3c15aada4ebd6b280a0e6a0de89
This treats the case of uses who have already downloaded 3.0 BETA 1 and that will upgrade to BETA 2

### DIFF
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -115,6 +115,10 @@ final class DBUpdater {
 	private void updateDBv301(int dbVersion, int version) {
 
 		try {
+
+			m_clJDBCAdapter.executeUpdate("ALTER TABLE MATCHESKURZINFO ALTER COLUMN isDerby SET DATA TYPE BOOLEAN");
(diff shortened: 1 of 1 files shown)


src/main/java/core/db/DBUpdater.java
```diff
@@ -115,6 +115,10 @@ void updateDB(int DBVersion) {
 	private void updateDBv301(int dbVersion, int version) {
 
 		try {
+
+			m_clJDBCAdapter.executeUpdate("ALTER TABLE MATCHESKURZINFO ALTER COLUMN isDerby SET DATA TYPE BOOLEAN");
```
akasolace, line 119: @tychobrailleur I was not sure if we need to check the type of the column first of if we could do this directly (I checked and it does not throws an error if called on integer columns)

tychobrailleur: I don't think you need to check the type first, no.  To be sure I did a quick test too, and that's fine.

It works with integer to boolean, that's the main thing, because it's narrowing the column size.

akasolace: ok cool thank you

Issue #506, opened 2020-05-06
[eBUG] One player can not occupy more than one substitute slot


wsbrenk:
Dillotu wrote in Hattrick forum:

Hi

In Lineup >>> Substitutes we cannot set the same player in two or more positions

akasolace:
@wsbrenk this was also a limitation in 2.X right?  Is it a change in HT or was it broken in HO for super long?  Also I guess this should go in a quick 3.1 release, what do you think?

wsbrenk:
@akasolace I am afraid that I have made this error with issue #299 in february

PR #507, opened 2020-05-06
Bug#506

wsbrenk:
 
   - fixes issue #506
  
 - [x] have been updated
 - [ ] do not require update


akasolace:
@wsbrenk by experience that part of the code is error prone. If you are very confident about the commit I create a 3.X branch and we do a quick 3.1 release otherwise even if it is not ideal we take time to test it and we do it for 4.0 (but this one can take a while). The thing is I had a quick look to your commit and it seems bigger change that I was expecting .... but up to you I let you decide.

Issue #520, opened 2020-05-15
[BUG]experience sub calculation does not work if players got not at least osmose training


wsbrenk:
i'm training keepers since 3 weeks and only they are getting experience subs

![grafik](https://user-images.githubusercontent.com/11821502/82044011-daf12700-96ac-11ea-9519-ddce67c2ba4f.png)

PR #522, opened 2020-05-16
#520: calculation experience sub of each minute played

wsbrenk:
 
   - fixes issue #520 

sorry, but not only was the home office a bit stressful for me lately, my bee colonies also wanted to be worked on. I quickly got a suggestion for bug # 520. Please check that. I have no idea whether this should make sense in a version 3 or only is for version 4.

And since my bees are again waiting for me, i only did a very short test on this.


akasolace:
ok I merge it and we will test it on 4.0 ....

Issue #489, opened 2020-05-03
[eBUG] In "Best Position" I cannot select "Coach" anymore


amoll1910:
**Describe the bug**
In combobox "Best Position", I cannot select "Coach" anymore.

**To Reproduce**
1. Go to tab "Squad"
2. Select a player
3. Click on combobox "Best Position"
4. The entry "Coach" is not listed anymore

**Screenshots**
![NoCoachBestPosition](https://user-images.githubusercontent.com/64730829/80912388-efdcca80-8d3c-11ea-8b9f-909f059da561.jpg)

**Platform information:**
 - HO 3.0 r2716
 - OS: Windows 10
 - Version 1909 18363.778

akasolace:
This was removed by mistake but we have to decide if we want to restore this possibility or not (I do not really see the purpose of this)
The only result of this is I guess to make sure the player is not selectionable by auto lineup (right @amoll1910 ? )   if that is the case I suggest to make it cleaner and to remove "coach" by "can't play" and that would be linked to the checkboxes in lineup tab. There would still be plenty of ways to do the same thing but at least we would remove one.   @wsbrenk  @tychobrailleur  what do you think ?

wsbrenk:
@akasolace i use it, only to classify my players. As far as i know, it is only a display feature.

amoll1910:
"can't play" instead of "coach" is nice and describes it even better for my purposes.
Then the player would not appear at training estimation as well (this was one reason why I marked some players as "coach").

PR #524, opened 2020-05-16
issue 489

akasolace:
fixes issue #489

- removes coach from best position in Squad Tab (the intent of that option was never clear)
- add unselectable option in best position in Squad Tab
- link unselectable option from Squad Tab and Unselectable Checkbox in Lineup Tab (i.e acting on one update the other ...)
- add translation terms
- update and rename the 2 asssociated icons

I can revert the icons if you prefer the current ones but I think this one are sharper:

before:
image

now:
image

 - [ ] have been updated
 - [x] do not require update

3. [Optional] suggested person to review this PR @tychobrailleur or @wsbrenk


src/main/java/module/lineup/AustellungSpielerTable.java
```diff
@@ -13,9 +14,11 @@
 import core.gui.model.UserColumnFactory;
 import core.model.HOVerwaltung;
 import core.model.UserParameter;
+import core.model.player.IMatchRoleID;
 import core.model.player.Player;
 import core.util.Helper;
 import module.playerOverview.PlayerTable;
+import module.playerOverview.SpielerUebersichtsPanel;
```
tychobrailleur, line 21: I think this import is unused.

akasolace: correct I removed it in new commit along some more code cleaning in this file

PR #528, opened 2020-05-18
closes #505

### LINKED ISSUE #505: [eBUG] Module Evil Card is broken
@wsbrenk @tychobrailleur   Should we repair it or is it also a candidate for removal?  I personally never used it ...

akasolace:
Remove evil card module

I was not sure about the icons so I did not touched them


src/main/java/core/module/ModuleManager.java
```diff
@@ -56,9 +55,7 @@ private void initMap(Map<Integer, IModule> map) {
 		map.put(Integer.valueOf(IModule.TSFORECAST), new TSForecastModule());
 		map.put(Integer.valueOf(IModule.SPECIALEVENTS), new SpecialEventsModule());
 		map.put(Integer.valueOf(IModule.TEAM_OF_THE_WEEK), new TeamOfTheWeekModule());
-		map.put(Integer.valueOf(IModule.EVIL_CARD), new EvilCardModule());
```
tychobrailleur, line 59: Do we want to remove `EVIL_CARD` from `IModule` too?

akasolace: yes you are right

Issue #531, opened 2020-05-21
[eBUG] Bad error message when db is lock


yaute74:
**Describe the bug**
Bad error message when the database is lock by another process

**To Reproduce**
1. Start HO (don't close it)
2. Start another time HO 
3. Error message is: 'Fatal database error. Exiting HO! You should restore the db-folder from backup or delete that folder.'

PR #532, opened 2020-05-21
#531: Bad error message when db is lock

yaute74:
Issue #531: better error message


akasolace:
welcome back @yaute74

yaute74:
Thank @akasolace 
I'll try to give some time to HO, it won't be much but better than nothing ;)

akasolace:
cool ...

PR #535, opened 2020-05-23
closes #533

### LINKED ISSUE #533: [BUG] add support for new leagues
**Describe the bug**

**To Reproduce**

**Screenshots**

**Platform information:**
 - OS: [Windows / macOS / Linux]
 - Version (including build number)

**Additional context**

akasolace:
 
   - fixes issue #__  
   
 - [ ] have been updated
 - [ ] do not require update


src/main/java/core/model/WorldDetailsManager.java
```diff
@@ -7,10 +7,10 @@
...
 
 	private static WorldDetailsManager WMANAGER = null;
-	private WorldDetailLeague[] leagues;
-	private HashMap<Integer, WorldDetailLeague> countryMap = new HashMap<Integer, WorldDetailLeague>();
-	private HashMap<Integer, WorldDetailLeague> leagueMap = new HashMap<Integer, WorldDetailLeague>();
-	private int totalUsers;
+	private static WorldDetailLeague[] leagues;
+	private static HashMap<Integer, WorldDetailLeague> countryMap = new HashMap<Integer, WorldDetailLeague>();
+	private static HashMap<Integer, WorldDetailLeague> leagueMap = new HashMap<Integer, WorldDetailLeague>();
+	private static int totalUsers;
```
tychobrailleur, line 13: Any reason for making all these `static`?  `WorldDetailsManager` seems to be a singleton, so `WorldDetailsManager.instance()` should do, and you can then call `getWorldDetailLeagueByCountryId` on it.

akasolace: no I guess we can go for your approach ... for my info what are the advantages of not making it static ?  I thought it would be cleaner this way ...

tychobrailleur: I usually dislike static because they are “viral” (makes other members static in the same class), and it makes it hard to unit tests, but here this is more for consistency: “state” classes like `HOVerwaltung`, `HOLogger`, `HOParameter`, `DBManager`, `User`, etc. etc. are all singletons, so this one would be the odd one out.

akasolace: Ok I am convinced ... I will do that then :-)

akasolace:
I will merge it and do the changes you requested in another commit because I have too many conflicts to manage ...

Issue #66, opened 2019-01-01
Improve usability team analyzer


ArneGloe:
Here is the issue as requested in the hattrick forum. Thanks for your work! Really appreciate it!

A great add on for me would be to save modified ratings of my next opponent. 
I often modify these according to my expectations while testing different line-ups in the team analyzer. Right now I manually reset them to my expectations every time I change the line-up.

wsbrenk:
![grafik](https://user-images.githubusercontent.com/11821502/80868037-3af0d200-8c98-11ea-91b7-63a97cd5cc42.png)

my idea is to add an extra row "adjusted" beneath the average row ("Durchschnitt"), when the user changes the opponents ratings in the MatchPredictionPanel. This row could be selected when user changes his lineup to simulate again with his adjusted match ratings.

akasolace:
@wsbrenk ok yes and how would the reset be done?
   - in the simulation we decide average or adjusted ?
   - in team analyser a button to reset adjusted to average ?
   - in team analyser a button to toggle adjusted row 
   - ...

wsbrenk:
@akasolace as long as the opponent team is not changed no reset is needed, i think.
if opponent team is changed the "adjusted" row will disappear, until the ratings of the new opponent team is adjusted. then the new adjusted ratings will be added to the list.

akasolace:
@wsbrenk ok sounds good !

wsbrenk:
progress report: it was a hard fight until i understood what is happening where. so i did a little refactoring of the code and i hope other developers confirms that it will get easier to understand. currently the ui looks like this - there is a new "adjusted" row after simulation was run with changed rating or tactic values.
![grafik](https://user-images.githubusercontent.com/11821502/83333148-6ff44280-a29f-11ea-9539-7c79faec8455.png)

currently i'm looking for the correct match type icon which should be copied from the adjusted match or empty - but not a wrong cup match icon.
And there is a lot of testing, which has to be done.

akasolace:
@wsbrenk I followed this one from far... but thank you for the status update !

akasolace:
might be implemented by ebc278f

@wsbrenk I will let you close the issue once tested and confirmed to work

wsbrenk:
@akasolace, @ArneGloe  we should try to get the author of the ticket for a test when a new dev version is in place

wsbrenk:
@akasolace i recognize that i worked longer than a month on this ticket - i'm really tired of this;-)

[...]

Issue #514, opened 2020-05-11
[eBUG] manage local vs HT time zone


fjrn:
Even after updating the time difference in preferences and restarting HO, the country's local time is not updated.

![TimeHT](https://user-images.githubusercontent.com/64761641/81593147-63ae5100-9395-11ea-980d-cb0a62b4f602.JPG)

akasolace:
Can you put a screenshot of the match tab from HT ? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

fjrn:
It is in Portuguese, if you need it I can send it in English.

![Match_HT](https://user-images.githubusercontent.com/64761641/81602474-0f11d280-93a3-11ea-8bbf-07b851265e8f.JPG)

akasolace:
@fjrn ok thank you, last thing, could you also join a copy of your db in the ticket.

fjrn:
> 

> @fjrn ok thank you, last thing, could you also join a copy of your db in the ticket.

Sorry, but I don't know what this information is.
Can you guide me, please?

akasolace:
If you can copy your db folder from HO, then zip it and paste it here as an attachment.

fjrn:
Please see if this is the information you want.

[db.zip](https://github.com/akasolace/HO/files/4617821/db.zip)

PR #557, opened 2020-06-06
Feature#66

wsbrenk:
 
   - fixes issue #66
   

 
 - [x] have been updated
 - [ ] do not require update

3. [Optional] suggested person to review this PR @akasolace , @tychobrailleur


flagsExtractor.py
```diff
@@ -0,0 +1,32 @@
...
+# sheet = Image.open("assets/icons.png")
+# count = 0
+#
+# for x in range(12):
+#     for y in range(97):
+#         a = (x + 1) * 40
+#         b = (y + 1) * 30
+#         icon = sheet.crop((a - 40, b - 30, a, b))  # Problem here
+#         icon.save("assets/icons/{}.png".format(count))
+#         count += 1
```
akasolace, line 32: I don't think you want that file in HO/  it has been moved into utils/pythonScripts/ ... I don't know why it happened but I think you have it twice in your branch ... can you check ?

src/main/java/core/db/DBUpdater.java
```diff
@@ -102,6 +102,8 @@ void updateDB(int DBVersion) {
 						updateDBv300(DBVersion, version);
 						updateDBv301(DBVersion, version);
 					case 302:
+					case 399:
+					case 400:
```
akasolace, line 106: I don't understand the logic: `case 400: updateDBv400`   is it not entering into a loop then ? 
To be frank, the all block since case 27 seems a bit strange, we used to have one db version per release and that was simpler to understand.
 Could we not simplify this by merging some of the update function ?   @tychobrailleur what do you think of this ?

src/main/java/core/db/DBUpdater.java
```diff
@@ -1189,21 +1189,23 @@ private boolean columnExistsInTable(String columnName, String tableName) throws
...
-		return rs.next();
-	}
+		if ( rs != null )return rs.next();
+		return false;	}
 
 	private boolean indexExists(String indexName, String tableName) throws SQLException {
 		String sql = "SELECT * FROM INFORMATION_SCHEMA.SYSTEM_INDEXINFO WHERE INDEX_NAME = '"
 				+ indexName.toUpperCase() + "' AND TABLE_NAME = '" + tableName.toUpperCase() + "'";
 		ResultSet rs = this.m_clJDBCAdapter.executeQuery(sql);
-		return rs.next();
```
akasolace, line 1206: is this due to a change in java 14 ?  before `rs.next()` was returning false and now he throws an error if rs is null ??
Is this why you had to change it everywhere ?

src/main/java/core/model/match/MatchKurzInfo.java
```diff
@@ -269,8 +269,16 @@ public final void setMatchDate(java.lang.String m_sMatchDate) {
...
+					try {
+						// Hattrick
+						SimpleDateFormat simpleFormat = new SimpleDateFormat(
+								"yyyy-MM-dd");
+						this.matchDateTimestamp = new Timestamp(simpleFormat
+								.parse(m_sMatchDate).getTime());
+					} catch (Exception ex) {
+						HOLogger.instance().log(getClass(), ex);
+					}
+				}
```
akasolace, line 281: ideally we should have separate commit / PR for separate issue ...

flagsExtractor.py
wsbrenk: @akasolace i was also wondering how this file got into my branch. must have happened in one of the merges that I've done over and over again. i will remove it.

src/main/java/core/db/DBUpdater.java
wsbrenk: @akasolace case 400 should be shifted behind the call of updateDBv400. case 399 is necessary to handle the DEV-Versions which are used by us and maybe some brave users. the space in between is left for patch versions of release 3 when version 4 development is in progress.
if you increment only by 1 you give away the chance to deliver patches which needs db updates.

src/main/java/core/db/DBUpdater.java
wsbrenk: git commit asked me to review the code and gives warnings that rs could be null

src/main/java/core/model/match/MatchKurzInfo.java
wsbrenk: isn't #514 the separate issue? we discussed it on slack. see this https://github.com/akasolace/HO/commit/eaa1b9b246a9d791049df3adfd0b41afd1178459#commitcomment-39510796

the commited change is one possibility to repare it. Another one yould be to update the database and introduce long time formats.

akasolace: yes this is related to #514, when I closed that issue I introduced the problem because I did not know/thought that very old entries could be missing time information.
What you just did here fix the issue I introduced but it does not directly relate to this PR.  But no problem leave it here, it is a small one and it does not pollute the PR a lot

Issue #512, opened 2020-05-09
Sub-level recalculation error


fjrn:
When requesting HO to Recalculate sublevels or Recalculate sublevels (7 weeks), the following error occurs:
java.lang.NullPointerException
![java lang NullPointerException](https://user-images.githubusercontent.com/64761641/81481744-92e38780-9208-11ea-8880-9360d4184cb9.JPG)

**Additional information**
![Error](https://user-images.githubusercontent.com/64761641/81481571-89a5eb00-9207-11ea-950e-d8ee8509541b.JPG)

fjrn:
Obrigado pela atenção. Se precisar de mais informações, eu estou a disposição.

akasolace:
@fjrn  a copy of your db would help looking at this issue ... thank you

fjrn:
[db.zip](https://github.com/akasolace/HO/files/4617841/db.zip)

wsbrenk:
db contains only one single training. subskill recalculation seems to skip the very first training.

with current master i could not reproduce the npe

akasolace:
@wsbrenk just to clarify is it that we can't reproduce the error or that the error only occur if user tries to recalculate subskill when there is only one training in the database ?

wsbrenk:
@akasolace i could not reproduce the npe given by @fjrn. i think we still did repare it the last weeks. wasn't there another ticket concerning "getAufstellung"?!

what i saw, was that there is still another bug which happens if only one single training is in the database and all stored hrfs are after the trainings date. Then NO recalculation is done.

wsbrenk:
![grafik](https://user-images.githubusercontent.com/11821502/83980902-8e9aaf00-a919-11ea-93be-a7fa900c8a6f.png)

recalculation is done now also on trainings that took place before the first hrf was loaded.

i did introduce a lazy loading option to HOModel, since a lot of stuff was loaded by HOVerwaltung.loadModel which is not needed by the recalculation method.

fjrn:
Thank you for your service.
I can't interpret the error report, so I can't comment on the error. I believe that the proposed improvement will increase the yield.
The sub-level recalculation is working perfectly

PR #559, opened 2020-06-08
Bug512

wsbrenk:
 
   - fixes issue #512 
 
 - [x] have been updated
 - [ ] do not require update

a lot of the files were attached since i translated some german names

a recalc of my 15 years HO history took place in less than 5 minutes - that's OK i think.

### DIFF
--- a/src/main/java/core/model/HOModel.java
+++ b/src/main/java/core/model/HOModel.java
@@ -41,14 +41,13 @@ public class HOModel {
     private Lineup m_clAufstellung;
     private Lineup m_clLastAufstellung;
     private Basics m_clBasics;
-//    private EPV epv = new EPV();
     private Finanzen m_clFinanzen;
     private Liga m_clLiga;
     private Spielplan m_clSpielplan;
     private Stadium m_clStadium;
     private Team m_clTeam;
-    private Vector<Player> m_vOldPlayer = new Vector<Player>();
-    private Vector<Player> m_vPlayer = new Vector<Player>();
+    private static List<Player> m_vOldPlayer;
+    private List<Player> m_vPlayer;
     private Verein m_clVerein;
     private XtraData m_clXtraDaten;
     private int m_iID = -1;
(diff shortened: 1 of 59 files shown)


akasolace:
This is a lot of cleanup !

src/main/java/core/model/HOModel.java
```diff
@@ -100,29 +103,25 @@ public final void setAllOldSpieler(Vector<Player> playerVector) {
...
     }
 
     //---------Player--------------------------------------
 
     /**
-     * Gibt alle Player zurück
+     * Returns all current players
      */
-    public final Vector<Player> getAllSpieler() {
+    public final List<Player> getCurrentPlayer() {
```
tychobrailleur, line 118: This should be made plural, `getCurrentPlayers`, it's confusing otherwise.

src/main/java/module/lineup/AufstellungsAssistentPanel.java
```diff
@@ -346,12 +346,10 @@ private void startAssistant(HOModel hoModel, HOMainFrame mainFrame) {
 			}
 		}
 
-		final Vector<Player> vPlayer = new Vector<Player>();
-		final Vector<Player> allePlayer = hoModel.getAllSpieler();
+		final List<Player> vPlayer = new Vector<Player>();
```
tychobrailleur, line 349: I don't have full context on my phone, but shouldn't `Vector` be an `ArrayList` here too?

tychobrailleur:
> This is a lot of cleanup !

It sure is, that's great, thanks @wsbrenk !!

akasolace:
@tychobrailleur sorry I did not see your review was ongoing ...

src/main/java/core/model/HOModel.java
wsbrenk: getCurrentPlayers would be my favorite
there is another getter getAllOldSpieler (sorry i missed this one) which fetches the complete list of historical players of your club. getAllPlayers could lead to confusion i think.

src/main/java/module/lineup/AufstellungsAssistentPanel.java
wsbrenk: confirmativ - i only changed the declaration as List, whereas i never changed the allocation. But i think it would be a good idea to change it to ArrayList

Issue #564, opened 2020-06-11
[FEATURE] change instalation folder on Linux


akasolace:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Additional context**

Issue #548, opened 2020-06-03
[BUG] build and test artefacts made using install4j


akasolace:
**Describe the bug**

**To Reproduce**

**Screenshots**

**Platform information:**
 - OS: [Windows / macOS / Linux]
 - Version (including build number)

**Additional context**

Issue #202, opened 2019-08-16
[eFEAT] Bundle JRE in Executable


pierpaolocira:
This request follows #198 

When #106 is done, verify if the MacOS `.app`  can be distributed by embedding the JRE.

From the related issue

> It is not clear if it is related either to the code (not yet compatible with Java 11 (see issue #106 ) or to an issue in the plugin used to build the app (edu.sc.seis.macAppBundle, see crotwell/gradle-macappbundle#56 and crotwell/gradle-macappbundle#49 )... or perhaps both.

akasolace:
@tychobrailleur @wsbrenk @yaute74  do you think this is something we want to do ?

pierpaolocira:
Hello... I did something alike for my company few years ago, because it was in line with Oracle guidelines.
This was just a .EXE (with a budle of libraries in subfolders) to be launched without the need to install/configure any JRE (and, even better, without generating conflicts with any already existing JRE on the host machine).

Accordingly to Oracle and what I have read (and experienced), in my personal opinion, this should be the way to go: the user download and use it... without caring of the JRE versione used to build, which should change more frequently, since Oracle has introduced the new release plan.

If I remember well, this feature was removed in the MacOS APP because the gradle plugin used to build for MacOS wasn't supporting JRE different than version 8 at that time (and was generating an invalid APP without triggering any exception).

But ok... just my opinione...

akasolace:
@pierpaolocira we will propose both ....

PR #568, opened 2020-06-11
handover db location from script to HO.

### LINKED ISSUE #566: [BUG] repair artefacts build process following transition to install4j
**Describe the bug**

**To Reproduce**

**Screenshots**

**Platform information:**
 - OS: [Windows / macOS / Linux]
 - Version (including build number)

**Additional context**

akasolace:
 
   - closes #566 , #564, #548, #202 and  #372

   - information about portable vs installed App is passed by launcher

   - db location (and logs output) are now managed in HO based on platform and install type:
      - portable: same directory
      - windows: appdata 
      - linux  ~.ho 
      - maxOS Application Folder

   - Offers artefacts with bundled JREs

  Awaiting feedack before merging


src/main/java/core/HO.java
```diff
@@ -20,34 +20,24 @@
...
+	}
 
-	/**
-	 * A RELEASE is when a build artifact gets delivered to users. Note that
-	 * even a DEVELOPMENT version can be a RELEASE ("Beta"). So when a version
-	 * is build (no matter if DEVELOPMENT or not), this flag should be set to
-	 * true. The main purpose for the flag is to disable code (unfinished new
-	 * features, debug code) which should not be seen in a release.
-	 */
+	private static boolean portable_version; // Used to determine along the os the location of te
```
tychobrailleur: location of `te`?

src/main/java/core/db/User.java
```diff
@@ -1,138 +1,146 @@
...
-				load();
+	public static String getDbParentFolder() {return dbParentFolder;}
+
+	private static void setDBParentFolder(){
+		if (! HO.isPortableVersion())
+		{
+			if (HO.getPlatform() == OSUtils.OS.LINUX) {
+				dbParentFolder = System.getProperty("user.home") + "/.ho";}
+			else if (HO.getPlatform() == OSUtils.OS.MAC) {
+				dbParentFolder =  System.getProperty("user.home") + "/Library/ApplicationSupport/HO";}
```
tychobrailleur, line 63: There should be a space between `Application` and `Support`.

akasolace: I think it was correct before so I don't know how I end up changing this one ...

src/main/java/core/HO.java
akasolace: I fixed this in last commit

src/main/java/core/db/User.java
akasolace: > Apologies, I had just started the review when you merged.

No worry, I merged the PR because it started to be big and rebasing to master made the PR really messy. We might still have some issue but I think we are not far to have something solid

akasolace: done in 9c8ddeeab3cce923541ef88f0397bae914caa0cc

pierpaolocira:
@akasolace I would like just to comment this description, that I could read only now.
I could say nothing (because this is already a PL and I know this is late), but I would prefer mentioning it now instead, at least you know.
As we have already mentioned several times, saving settings in "maxOS Application Folder" is not a good idea on MacOS.
They should be saved in the "Application Support" folder... as all other applicaitons do. In this way you're probably linking the DB with the specific installation and version.
Furthermore not only DB folder, but also other folders are now saved in the main folder of the user... please check that also those are saved in the same folder of the DB...

PR #580, opened 2020-06-16
Double click and click+shift

steffanots:
Double click on row in Player Analysis open Match report
Click+SHIFT on row in Player Analysis open ht website

For this purpose I need that column 54 - MatchID would be locked.

Somebody need to test URL in linux and mac!!!
I use firefox for linux, and firefox must be in root path. Maybe I can test tommorow at office (we use centos).


akasolace:
Thank you for the PR @steffanots      @yaute74  could you review the PR and maybe test it under Linux before merging ?

PR #588, opened 2020-06-18
Demonstrate how to use darklaf IconLoader to generate jersey icon.

weisJ:
   
This PR shows how to use the `IconLoader` to load svg icons with variable colors with the jersey icon as an example.
This implementation doesn't depend on the internal implementation of the svg icon (as that is already handles by the IconLoader). 

**Note** that using `IconLoader#loadSvgIcon` the caching done by the `IconLoader` is bypassed (as it should be for icons that have individually assigned colors.)

Example result: 
`ImageUtilities.getJerseyIcon(IMatchRoleID.keeper, (byte) 0, 49, 100)`
jersey

It wasn't quite clear to be what the color of the outline stripe should be, so I made it to be the text color. This can be easily adjusted though.

How it's done:

The important part here is the setup in the `jersey.svg` file:
````xml
<defs id="colors">
    <linearGradient id="jerseyColor">
        <stop offset="0" stop-color="#FF0000"/>
    </linearGradient>
    <linearGradient id="collarColor">
        <stop offset="0" stop-color="#00FF00"/>
    </linearGradient>
    <linearGradient id="outlineColor">
        <stop offset="0" stop-color="#0000FF"/>
    </linearGradient>
</defs>
````
These colors can be references using `url(#key)`. They are replaced with the colors in the map passed to `IconLoader#loadSvgIcon`. If no map is passed then the colors are assumed to be retrievable using the `UIManager` e.g.
````xml
<linearGradient id="Label.foreground">
    <stop offset="0" stop-color="#FF0000"/>
</linearGradient>
````
is resolved to the color `UIManager.getColor("Label.foreground")`. One can also specify fallback keys for when one key might not be available in a specific Laf.
````xml
<linearGradient id="key" fallback="fallback1,fallback2,fallback3">
    <stop offset="0" stop-color="#FF0000"/>
</linearGradient>
````
The keys are tried in the order `key -> fallback1 -> fallback2 -> fallback3` as long as the resulting color is `null`.

**Note**: The specific colors in the definition are just stubs as they get replaced when loading the icon. An equivalent definition would be
````xml
<defs id="colors">
    <linearGradient id="jerseyColor"/>
    <linearGradient id="collarColor"/>
    <linearGradient id="outlineColor"/>
</defs>
````
with the drawback, that viewing the svg outside of the program will result in an empty image.
 - [x] do not require update

### CODE REVIEW

File: src/main/java/core/gui/theme/ImageUtilities.java [hunk 1]
```diff
@@ -352,369 +352,318 @@ public static ImageIcon getWideImageIcon4Veraenderung(int value, boolean current
...
+            Icon numberIcon = EmptyIcon.create(0);
+            if (trickotnummer > 0 && trickotnummer < 50) {
+                int baseline = Math.round(height * 13f / 16f);
+                int fontSize = Math.round(height * 8f / 16f);
+                numberIcon = new TextIcon(String.valueOf(trickotnummer),
+                                          textColor,
+                                          new Font(Font.SANS_SERIF, Font.BOLD, fontSize),
+                                          width, height, baseline);
+            }
+            komplettIcon = new OverlayIcon(jerseyIcon, numberIcon, size, size);
```
tychobrailleur, line 564: The formatting changes make it hard to see what the actual change, but this is the bulk of it right here, right?

src/main/java/core/gui/theme/ImageUtilities.java
```diff
@@ -352,369 +352,318 @@ public static ImageIcon getWideImageIcon4Veraenderung(int value, boolean current
...
+        String key = "trickot_" + posid + "_" + taktik + "_" + trickotnummer + "_" + size;
+        Icon komplettIcon = ThemeManager.getIcon(key);
+
+        if (komplettIcon == null) {
+            Color jerseyColor = getJerseyColorByPosition(posid);
+
+            double brightness = ImageUtilities.getBrightness(jerseyColor);
+            Color textColor = brightness < 130 ? Color.WHITE : Color.BLACK;
+
+            Map<Object, Object> colorMap = Map.of("jerseyColor", jerseyColor,
```
tychobrailleur, line 547: So `IconLoader` can handle mapping SVG IDs to their fill colour for us?    That's nice!

build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp:4.3.1'
     implementation 'com.github.weisj:darklaf-core:2.2.3'
     implementation 'com.github.weisj:darklaf-theme:2.2.3'
-    implementation 'com.formdev:svgSalamander:1.1.2.1'
```
tychobrailleur, line 23: @yaute74 You may need to check whether this will affect the changes you are working on.

src/main/java/core/gui/theme/ImageUtilities.java
weisJ: Yes. I'll try to fix the commit so it doesn't change the line endings.

weisJ: I have recommitted the changes without the changes to the formatting.

src/main/java/core/gui/theme/ImageUtilities.java
weisJ: Also this is essentially a no-op until the icon actually gets used.

build.gradle
yaute74: Ok. I will first work on issues before come back to the svg, we're starting to get a lot of issue :(

tychobrailleur:
Hmmm... Following this merge I am not seeing any jersey any more.  I am investigating.

weisJ:
Do you see a jersey when running `JerseyIconTest`?

tychobrailleur:
> Do you see a jersey when running `JerseyIconTest`?

I do, and also see in debugger when the icon is loaded so the actual icon loading is fine, so somehow it's in the display.

tychobrailleur:
Problem seems to be with the cached icon, disabling caching restores the jerseys on both Squad and Lineup tabs.

weisJ:
The problem is that `ThemeManager#getIcon` assumes that the icon is of type `ImageIcon`. The simplest solution is to replace
````java
Icon komplettIcon  = ThemeManager.getIcon(key);
````
with
````java
Object cachedIcon = ThemeManager.instance().get(key);
Icon komplettIcon = cachedIcon instanceof Icon ? (Icon) cachedIcon : null;
```` 
For a long term solution `ThemeManager#getIcon` should probably be agnostic about the type of the image i.e. not immediately delegate to the `ImageIcon` implementation.

tychobrailleur:
> The problem is that `ThemeManager#getIcon` assumes that the icon is of type `ImageIcon`. 

Yes, also figured that out, I am in the process of handling `Icon`s everywhere.

Issue #584, opened 2020-06-17
[BUG] Fatal database error


Dillotu:
Hi guys

HO does not start anymomre.

Error message Fatal database error
You should restore the db-folder from backup or delete that folder.

If I click OK I have the option to set the last 3 db in the list

14/06
15/06
17/06

No one works.

I also tried to replace the db manually in the folder from the zip file with no luck unfortunately.

Can you please let me know how can I recover the DB?

Thanks

**Platform information:**
 - OS: [Windows]
 - Version HO_3.0.2.2728

 [3 Db link Google Drive](https://drive.google.com/drive/folders/10kgDYlEhZ011hMv8TPGAgPU4OkNGLdr-?usp=sharing)

akasolace:
@tychobrailleur might be a good test for your new db tool?

tychobrailleur:
In the first of the three DBs:

```
sql> select * from userconfiguration where config_key = 'DBVersion';
CONFIG_KEY  CONFIG_VALUE
----------  ------------

Fetched 0 rows.
sql> select * from userconfiguration ;
CONFIG_KEY                  CONFIG_VALUE
--------------------------  ------------
transferTypePane_splitPane  302
updateCheck                 true
waehrungsID                 3
xmlDownload                 true
zahlenFuerSkill             true
zellenbreitenFaktor         1.0

```

which is quite weird. Same with the last.
Without `DBVersion` in `userconfiguration`, the app will not start.

akasolace:
how on earth could that be missing ?

tychobrailleur:
> how on earth could that be missing ?

No idea, also odd is that `transferTypePane_splitPane` is 302, like the version – maybe a coincidence...

I have changed the repair tool to add the `DBVersion`, the app now starts, but then goes into loop:

```
URL: https://chpp.hattrick.org/chppxml.ashx?file=matchorders&matchID=662815588&sourceSystem=hattrick&version=3.0&teamId=281586
 [Info]    HOMainFrame: This is HO! version 0.0, have fun!
 [Info]    HOMainFrame: Operating system found: Mac OS X on x86_64 (10.14.6)
 [Info]    HOMainFrame: Using java: 1.8.0_77 (Oracle Corporation)
URL: https://chpp.hattrick.org/chppxml.ashx?file=matchorders&matchID=662815588&sourceSystem=hattrick&version=3.0&teamId=281586
 [Info]    HOMainFrame: This is HO! version 0.0, have fun!
[...]

PR #593, opened 2020-06-18
Ensure the HOMainFrame does not call itself whilst still loading.

tychobrailleur:
Under certain conditions HOMainFrame constructor ends up calling
`HOMainFrame.instance()` which in turn calls the HOMainFrame
constructor, as the null check of the singleton pattern is still
null (the constructor has not completed yet).

This PR ensures the main frame is not loading before display the OAuth
dialog is created.  If it still is loading, the dialog is created with
a null window, which should only cause the dialog to not be centered.

(This PR is for the 3.x release branch – I will create a separate PR for master tomorrow)

   - fixes issue #584 (partly)
   
 - [x] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/gui/HOMainFrame.java
+++ b/src/main/java/core/gui/HOMainFrame.java
@@ -60,6 +60,7 @@ import java.util.Calendar;
 import java.util.List;
 import java.util.Locale;
 import java.util.Vector;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.swing.InputMap;
 import javax.swing.JFrame;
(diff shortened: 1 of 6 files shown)


akasolace:
wow, how did you find about it?  I don't think I ever experienced it ....

tychobrailleur:
> wow, how did you find about it? I don't think I ever experienced it ....

The problem was due to the db being fully populated with upcoming matches without info, but with no valid OAuth token in the user config table.

When starting the app, the Squad tab was trying to load info about the matches (I think for training prediction), but because no OAuth token was present, was trying to open the OAuth dialog, referencing the HOMainFrame code.

I thought at first it was a problem of synchronization in the singleton code, but even with `synchronized` keyword, or eager loading, the problem was there.  Thankfully I have seen this problem before in the singleton pattern where the constructor is not fully complete, and somehow calls itself again, causing infinite loop.

Although it seems like an unlikely scenario (not a valid OAuth token), I think it does explain some of the hanging app reported.  It still doesn't explain the corrupted db in the first place though.

akasolace:
ok thanks for the explanation

Issue #569, opened 2020-06-12
[FEATURE] Clicking on a match in Player Analysis should bring to the Match report 


tychobrailleur:
**Is your feature request related to a problem? Please describe.**

No, it is a UX enhancement.

**Describe the solution you'd like**
When clicking on a match in the Matches tab of the Player Overview module, this should bring to the Match report in the Matches module:

![Screenshot 2020-06-12 at 08 28 52](https://user-images.githubusercontent.com/114285/84477007-deb1a280-ac86-11ea-9d88-f65abd1ca735.png)
Clicking (or double-clicking if preferred) the first row should bring to:

![Screenshot 2020-06-12 at 08 30 41](https://user-images.githubusercontent.com/114285/84477137-128cc800-ac87-11ea-8f57-637e9e2b6df9.png)

The same pattern used in other places could be used here, Shift-click could bring directly to the Hattrick match report.

**Additional context**

N/A

akasolace:
and for consistency  shift+click open HT website  (this is linked to #547)

tychobrailleur:
> and for consistency shift+click open HT website (this is linked to #547)

Thanks, I have corrected from ctrl-click to shift-click.

steffanots:
take

akasolace:
@steffanots  it should be without the quote otherwise it is interpreted as quote block ...

steffanots:
I try take, 'take', "take"

steffanots:
I finish link to match tab but I can find any code with link to ht website!
do we have any code?

akasolace:
@steffanots look here  https://github.com/akasolace/HO/blob/d4abaa7df667a1e3908c3f8c68b4b528257d6d1d/src/main/java/module/specialEvents/Match.java#L151

Issue #547, opened 2020-06-03
[FEATURE] clickable match link open ht website


akasolace:
@steffanots while you are at it, for consistency the behaviour should be the same as in the special event modules (shift click opens in browser):
     - player details 
     - overview / lineup tab #461

steffanots:
teke

akasolace:
@steffanots  it should be without the quote otherwise it is interpreted as quote block ...

steffanots:
I don't understand where it is necessary to have a link to a hattrick

akasolace:
@steffanots you mean where or why ?

steffanots:
Where and what :)
Player or match?

akasolace:
the idea is to implement shift+click => open HT website

![image](https://user-images.githubusercontent.com/1136496/85059284-b1438800-b1a3-11ea-9049-64b1b01f4f51.png)

steffanots:
I create link on shift+click

akasolace:
@steffanots perfect

PR #595, opened 2020-06-19
SHIFT + click to WEB page for matchId

steffanots:
#547 
I add link to hattrick match by Id
I also add method `isCupMatch` in class MatchType because for hattrick it is official match. I didn't want to change the method `isOfficial`.
When you merge all branches we must change for #569 this condition and create one class or method for web pages.


akasolace:
@steffanots well spoted but I think that we should fix isOfficial() instead .....

PR #600, opened 2020-06-20
Code by abstraction, and use `Icon` rather than `ImageIcon`.

tychobrailleur:
 
  This PR makes sure the code uses the more abstract `Icon` rather than `ImageIcon`.  This solves the issue of the cached icon not being returned for the new jersey icons, which are not imageicons.
 
 - [ ] have been updated
 - [x] do not require update


akasolace:
@tychobrailleur  I suggested @yaute74  to do the review. He is more into that part of the code than me and also I have a bit less time than usual at the moment with the moving ....

src/main/java/core/gui/theme/ImageUtilities.java
```diff
@@ -648,6 +654,31 @@ public static Icon getSvgIcon(String key, int width, int height) {
...
+			return ((ImageIcon)icon).getImage();
+		}
+		else {
+			int w = icon.getIconWidth();
+			int h = icon.getIconHeight();
+			GraphicsEnvironment ge =
+					GraphicsEnvironment.getLocalGraphicsEnvironment();
+			GraphicsDevice gd = ge.getDefaultScreenDevice();
+			GraphicsConfiguration gc = gd.getDefaultConfiguration();
+			BufferedImage image = gc.createCompatibleImage(w, h);
```
weisJ, line 674: This should probably be more explicit with the type of image to create i.e. `Transparency.BITMASK` (as icons could be transparent).

src/main/java/core/gui/theme/ThemeManager.java
```diff
@@ -108,24 +108,24 @@ public Object get(String key){
 		return tmp;
 	}
 
-	private ImageIcon getImageIcon(String key){
+	private Icon getImageIcon(String key){
```
weisJ, line 111: Maybe a new name should be chosen if it doesn’t necessarily return an `ImageIcon` anymore.

weisJ:
@tychobrailleur I have created a PR into this branch for using `DerivableIcon`s for creating scaling `Icon`s. This way lossless re-scaling can be achieved for icons that aren't of type `ImageIcon`.

Issue #587, opened 2020-06-18
[BUG] missing several players in "prediction" tab of trainings


johari75:
**Describe the bug**
I´m missing several players in the "prediction" tab of trainings. 
In the the overview of the "training" tab there are all players available.
I set the "Futur Trainings" for 32 weeks. (I also tried 48 weeks without success)
I´m using HO since several years in different versions. So I had a lot of .hrf files imported.
I found out that the amount of players in the "prediction" tab depends on the flags at the bottom of the "training" tab in the options. By enabling/disabling of each flag the amount varies. But no combination leads to desired result.

**To Reproduce**
1. Go to 'Training' module
2. go to 'Training" tab
3. count players
4. go to 'prediction" tab
5. count players
6. go to "Main" menu
7. click "options"
8. restart HO
9. go to 'Training" tab
10. change one of flags at the bottom of the page
11. repeat step 1 till 5

You can check all combinations concerning these flags. 

I got asked to add my database
[db.zip](https://github.com/akasolace/HO/files/4797721/db.zip)

**Screenshots**
![options](https://user-images.githubusercontent.com/67096604/85005106-a4e60d80-b158-11ea-8df4-bf7d169ee675.JPG)

![Training_overview](https://user-images.githubusercontent.com/67096604/85005108-a57ea400-b158-11ea-9da5-4c1a2d55691d.JPG)

![Training_prediction](https://user-images.githubusercontent.com/67096604/85005111-a6173a80-b158-11ea-81b4-8c3e6e9ab35d.JPG)

**Platform information:**
 - OS: Windows 10 home
 - Version 1909 (Build 18363.904)

akasolace:
@wsbrenk  maybe we want to rethink which players are displayed and/or highlited in the prediction tab.  What do you think ?

wsbrenk:
@akasolace we can do so. the current version should list all players that will get skill ups during the period that is analyzed, if configured as shown by @johari75 . 
if skill downs should be displayed too, the list could be a little longer.
Another chance to include further players to the forecast list, will be to change their position in the player overview list, so that they are no longer handled to get no trainings effect.
If they should get skill ups on the selected position during the analyse period and they are not displayed in the forecast list, it is a bug, that we should fix.

wsbrenk:
Currently it is intended that player with no skillups or -downs are NOT listed in the forecast
by  configuration the skilldown display can be suppressed - only players with skillups are listed
[...]

PR #602, opened 2020-06-20
#587: fix trainingspeed calculation

wsbrenk:
 
   - fixes issue #587  (not completly)


akasolace:
yes this is a bug fix but I dont think it closes #587

Issue #461, opened 2020-04-21
[eFEAT] put last match information in both squad and Lineup tables


tychobrailleur:
**Is your feature request related to a problem? Please describe.**

For friendlies, I want to make sure I am not overlooking a player that hasn't played recently, and therefore hasn't received training.  It'd be good to add the last played date as a column in the Lineup table (the training indicator is great, but I still would like to know how long some players have not played,  they could be prime sacking candidates)

**Describe the solution you'd like**

 A new column in the Lineup table with the date of the last played game.

**Additional context**
N/A

Next steps before closing the issue (edited by @akasolace ):

 - [ ] the new column need to be visible by default otherwise users won't find about it
 - [ ] add links to player rows in lineup and squad tab (#609)
 - [ ] I would put the new column on the left (maybe after best position)  (can be done as part of  #484)
 - [ ] remove rating column (to be tackled by #610)

akasolace:
@tychobrailleur for my info is that an info available by hrf direclty or should it be calculated ?

tychobrailleur:
I am not sure whether it is in the HRF, but according to CHPP docs it can be found in playerdetails as part of the `LastMatch` field.

akasolace:
@tychobrailleur sorry was a typo I meant CHPP...

wsbrenk:
@tychobrailleur According to the chpp documentation, this information should also be available in the players file, which is already downloaded by HO. We could add those information to the database and to the player overview panel.

steffanots:
Last match is last match in HT. I think it's an irrelevant type of match (league, cup, friendly, duel or preparation). I don't know that we want this.
I add last match into Player Detail with link on match.
I add columns into players tab and lineup tab.
I add property into english.properties.

Because I have problem with loading matches (create repo with bug) , I need help. 
I need master and my branch in new repo to finish testing.

![image](https://user-images.githubusercontent.com/64475678/81229070-8506ea00-8fef-11ea-9674-632035f6c7cc.png)

akasolace:
It is is linked to #485

akasolace:
#485 introduced the following: 

![image](https://user-images.githubusercontent.com/1136496/81501601-9561e200-92d9-11ea-9a92-17a716526d4f.png)

Current ticket is for including these information inside both Lineup and Squad tables replacing existing Rating information


This would summarize all in we want in a single place. In the column, I guess we could here also introduce 3 clicks sort (date defaults, rating and match type)

[...]

PR #607, opened 2020-06-22
LastMatch column (date + rating) order by date, rating

steffanots:
#461 
LastMatch column (date + rating) order by date, rating

Resolve bug on PlayerDetailPanel ... clear add before set last match info

Notice: Last match is not translated for Serbian(cyrillic)


akasolace:
ok thank you

> Resolve bug on PlayerDetailPanel ... clear add before set last match info

can you give me more info on this one ?

> Notice: Last match is not translated for Serbian(cyrillic)

yes as soon as I merge the PR you will see this available in POEditor

steffanots:
image
Last match in label is from above of selected player

tychobrailleur:
I am not sure whether this is the relevant PR, but on my machine the new column is the last one both on Squad and Lineup, and it is truncated, which is an issue as the column cannot be resized:

Screenshot 2020-06-23 at 19 10 14

Issue #572, opened 2020-06-13
[BUG] error for users using fatcow icons 


akasolace:
**Describe the bug**
As reported by davidatorus,  upgrade to 4.0 fails for users using fatcows : icons removal should be treated !

akasolace:
hello @tychobrailleur  I think this could be treated as part of darklaf light/dark icons ... what do you think?

tychobrailleur:
Sorry, @akasolace would you have more context on this?  Where are the fatcow icons used?  It sounds like their usage is platform-dependent? And what is the error encountered?

akasolace:
@tychobrailleur actually until now this was an additional set of icons available in /themes.
I removed those extra icons as they were unmaintained (lots of icons were missing) and I don't think we have the resources to offer this. To my surprise at least one user was still using those icons despite having a terrible experience. Hence he encountered an error while testing 4.0.  I think this is a very easy one to fix and I could do it (especially I created the issue) but I thought it could make more sense if you take care of it as if I understand correctly at some point you will introduce dark icons (with community help). Hence it is best if you look after those icons/themes related issues. Does it make sense to you? Is that ok?

tychobrailleur:
> Hence it is best if you look after those icons/themes related issues. Does it make sense to you? Is that ok?

@yaute74 is the icon virtuoso, so I am happy to look into this, with his help (& the community's)! ;-)

I still don't understand where those icons were used, though: 

> To my surprise at least one user was still using those icons despite having a terrible experience.

Where are these icons being configured?  I have never come across a place where you could select your own icons, is it in the old themes formerly loaded through a zip file?

akasolace:
yes. I think (never used it) the icons were unziped and used instead of default one. @DavidatorusF could maybe tell us more about it ... possibly he was the only user of those icons ;-)

yaute74:
@akasolace so maybe we will have a problem, no ? if we rename the files (.png -> .svg) ?
And think it will not work as now we use SVG icons.

akasolace:
you mean fatcow? but we are dropping it .... or is it something else?

yaute74:
ok great, I misunderstand

weisJ:
If you choose to fully move to dark/light icons I would be happy to help you migrate them. I have seen that you are already using some svg icons. It should be noted that painting the `SVGIcon` directly provided by svgSalamander is rather slow. I would advise to use
````java
IconLoader.get().getIcon(path, width, height);
ˋˋˋ
[...]

Issue #586, opened 2020-06-18
[BUG]  Select File->Preference throws a NPE


akasolace:
**Describe the bug**
Select File->Preference throws a NPE

**To Reproduce**
Select File->Preference

**Screenshots**

![image](https://user-images.githubusercontent.com/1136496/84986292-a0acf680-b13e-11ea-9f8a-b32ce5f825d1.png)

**Platform information:**
 - OS: Windows
 - Version: 4.0 r2946

tychobrailleur:
I have a fix for this, but the colours set in the preferences are not persisted.

We have two options:
- Either we remove entirely the colour selection panel in the preference panel, and let themes handle those colours,
- Or persist colours in userconfiguration table.

Any preference @akasolace ?

akasolace:
@tychobrailleur what do you think ?   Personnaly I would rather remove colours in userconfiguration table...

akasolace:
@tychobrailleur 
more generally, I don't really like those colours, maybe we want to remove that ....
First of all it is absolutly no clear what they mean and also they are redundant with Status column.  Only missing icons is tranfert listed but we can simply add this one and get rid of the colours, what do you think ?

tychobrailleur:
> @tychobrailleur what do you think ? Personnaly I would rather remove colours in userconfiguration table...

Sorry, missed that one: I was favouring not making those colours configurable, and add them to theme, but I wouldn't miss them.

tychobrailleur:
Can this be closed @akasolace , now that the status colours have been removed?

PR #619, opened 2020-06-26
Remove player state colours.

tychobrailleur:
Attempt at displaying icons on multiline player label:

Screenshot 2020-06-26 at 17 01 19

   - fixes issue #572 and #586
 
 
 - [x] have been updated
 - [ ] do not require update

3. [Optional] suggested person to review this PR @akasolace @yaute74


akasolace:
@tychobrailleur  wow I love this  best commit of 2020 !

src/main/java/core/db/UserConfigurationTable.java
```diff
@@ -171,7 +172,7 @@ int getDBVersion() {
 	 * @param obj
 	 */
 	void store(Configuration obj) {
-		final HashMap<String, String> values = obj.getValues();
+		final Map<String, String> values = obj.getValues();
```
akasolace, line 175: just for curiosity is there any reason for this ?   usually I do the opposite I use hashmap instead of map, but honest I don't really what I am doing when doing this. I just thought it was best to be more specific ....

src/main/java/core/gui/comp/entry/DoppelLabelEntry.java
```diff
@@ -33,7 +38,7 @@ public DoppelLabelEntry(Color color) {
         super();
         m_clLinks = new ColorLabelEntry("", ColorLabelEntry.FG_STANDARD, color,
                                         SwingConstants.RIGHT);
-        m_clRechts = new ColorLabelEntry("", core.model.UserParameter.instance().FG_INJURED, color,
+        m_clRechts = new ColorLabelEntry("", DIFF_COLOR, color,
```
akasolace, line 41: I am sorry if it is a dumb remark , but as a first look it seems this should not be required if we drop coloring player name ... I guess I am missing something

src/main/java/core/gui/comp/entry/DoppelLabelEntry.java
```diff
@@ -17,6 +20,8 @@
     private IHOTableEntry m_clLinks;
     private IHOTableEntry m_clRechts;
 
+    private final static Color DIFF_COLOR = ThemeManager.getColor(HOColorName.FG_INJURED);
```
akasolace, line 23: same remark as below

src/main/java/core/gui/theme/HOColorName.java
```diff
@@ -188,4 +188,12 @@
...
 	public static final String HRF_DARKBLUE_BG = "hrf.darkblue.bg";
 	public static final String HRF_RED_BG = "hrf.red.bg";
+
+	// Player State colours
+	String FG_STANDARD = "player.state.standard";
+	String FG_TRANSFERLISTED = "player.state.transferlisted";
+	String FG_INJURED = "player.state.injured";
+	String FG_BRUISED = "player.state.bruised";
+	String FG_TWO_YELLOW_CARDS = "player.state.two-yellow-cards";
+	String FG_RED_CARD = "player.state.red-card";
```
akasolace, line 198: so color name are still used ?

src/main/java/core/db/UserConfigurationTable.java
tychobrailleur: It's a good question!  It is usually preferrable to code to abstraction (google “code to an interface, not an implementation” for research): this means that the client of the method only knows about the behaviour of the object, not its internal implementation.

For example, here `HashMap` is a specific implementation of a `Map`, but for some reason we could decide we'd prefer a `Hashtable`, `WeakHashMap`, a `TreeMap` or a `LinkedHashMap`: if we code to implementation, we need to change the type everywhere it's used; if we code to abstraction, you only change in one place. (an example of that is the recent `ImageIcon` change which required to change to `Icon` in lots of places)

src/main/java/core/gui/comp/entry/DoppelLabelEntry.java
tychobrailleur: It is the colour of the delta for the different skills of the player when downloading data from HT.

src/main/java/core/gui/theme/HOColorName.java
tychobrailleur: They are not used, but they are still defined in the main theme.  I kept them there in the case we ever need to use them again, but maybe this is a bad case of YAGNI?  I can remove if needs be.

src/main/java/core/model/HOParameter.java
```diff
@@ -52,8 +51,8 @@ public static HOParameter instance() {
     public int HOUsers = 0;
 
 	@Override
-	public HashMap<String,String> getValues() {
-		HashMap<String,String> map = new HashMap<String,String>();
+	public Map<String,String> getValues() {
+		Map<String,String> map = new HashMap<>();
```
tychobrailleur, line 55: Here is a good illustration of what I meant earlier: if we ever decide to use a `TreeMap` to always get sorted keys, you would only change code here, the rest would remain the same: it would still handle a `Map`.

src/main/java/core/db/UserConfigurationTable.java
akasolace: thanks for the detailled explanation

tychobrailleur: With pleasure: you're usually the one taking the time to explain to me. 🙂

PR #620, opened 2020-06-26
Delete Classic and System themes.

tychobrailleur:

  Remove outdated themes Classic and System.
 
 
 - [ ] have been updated
 - [x] do not require update


akasolace:
I merged it quickly but now I am thinking:
    - what will happen for user having one of those currently as default themes ?   Looks like this is not handled
   - also -i think it worth updating the release notes

PR #621, opened 2020-06-26
fixes loop in loading of user parameters

akasolace:
fixes loop in loading of user parameters
also make a single db call for loading parameters for better efficiency


src/main/java/core/db/UserConfigurationTable.java
```diff
@@ -117,6 +118,24 @@ private String getStringValue(String key) {
...
+	private HashMap<String, String> getAllStringValues() {
+		final StringBuffer sql = new StringBuffer(100);
+		sql.append("SELECT * FROM ");
+		sql.append(getTableName());
+
+		HashMap<String, String> map = new HashMap<>();
+		final ResultSet rs = adapter.executeQuery(sql.toString());
+		try {
+			while (rs.next()) {
+				map.put(rs.getString("CONFIG_KEY"), rs.getString("CONFIG_VALUE"));
```
tychobrailleur, line 130: Some of the values are not strings, can that cause issues in other places?

akasolace: I think I kept the same logic as before ... it was converting everything to string ... but might worth to check

Issue #625, opened 2020-07-03
[BUG] Can not see the next opponent at Rubin-Cup


PecataM:
**Describe the bug**

In the Drop-Down Menu in "Тeam analysis" I can not see the next opponent for the Rubin Cup.

tychobrailleur:
Is that something you've seen before, @wsbrenk?

wsbrenk:
@yaute worked on that last

tychobrailleur:
Would you be able to share your db with us @PecataM so we can try to replicate?

miloszatko:
[db_singleUser-2020-7-4.zip](https://github.com/akasolace/HO/files/4873158/db_singleUser-2020-7-4.zip)

I have the same problem

miloszatko:
![image](https://user-images.githubusercontent.com/20202008/86516820-fa424000-be23-11ea-9709-f936ade644e9.png)

Only next league match is showed in drop down menu.

tychobrailleur:
This seems to be working on master:

![Screenshot from 2020-07-04 21-49-29](https://user-images.githubusercontent.com/114285/86520804-5ca93980-be40-11ea-97fc-d3f35bb763a3.png)

tychobrailleur:
I have cherry-picked the relevant commit and created a PR to the 3.x release branch.  This should be fixed in the next 3.x release.

PR #627, opened 2020-07-05
Show consolation cup in 3.x release

tychobrailleur:
Cherry-picked da4f3a4a9c7a1b4a2757c0793f6f0cea0c9bee75 to address issue #625 

   - fixes issue #625 
 
 
 - [x] have been updated
 - [ ] do not require update

3. [Optional] suggested person to review this PR @yaute74

### DIFF
--- a/src/main/java/core/db/DBManager.java
+++ b/src/main/java/core/db/DBManager.java
@@ -943,7 +943,11 @@ public class DBManager {
 		return ((MatchesKurzInfoTable) getTable(MatchesKurzInfoTable.TABLENAME))
 				.getMatchesKurzInfo(teamId, matchtyp, asc);
 	}
-	
+
+	public MatchKurzInfo[] getMatchesKurzInfoUpComing(int teamId) {
+		return ((MatchesKurzInfoTable) getTable(MatchesKurzInfoTable.TABLENAME))
+				.getMatchesKurzInfoUpComing(teamId);
+	}
 
 	public MatchKurzInfo getMatchesKurzInfo(int teamId, int matchtyp,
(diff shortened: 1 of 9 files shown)


akasolace:
@tychobrailleur  @yaute74  with the move to installed,  building artefacts for 3.x will be challenging ..... It is feasible but not sure I wan to invest a lot of time just to build a last 3.x .... I would rather stick to 4.0 release.  What do you think ?

yaute74:
@akasolace as long as we don't have anything blocking, I think it's not necessary

tychobrailleur:
> @tychobrailleur @yaute74 with the move to installed, building artefacts for 3.x will be challenging ..... It is feasible but not sure I wan to invest a lot of time just to build a last 3.x .... I would rather stick to 4.0 release. What do you think ?

I would also personally stick to 4.0 (that would also address the Linux distribution issues, I would think?), but I guess this depends on the release timeline.  I think we are still a fair bit away from being ready, so some of the of the 3.x issues may need an earlier release?

akasolace:
> > @tychobrailleur @yaute74 with the move to installed, building artefacts for 3.x will be challenging ..... It is feasible but not sure I wan to invest a lot of time just to build a last 3.x .... I would rather stick to 4.0 release. What do you think ?
> 
> I would also personally stick to 4.0 (that would also address the Linux distribution issues, I would think?), but I guess this depends on the release timeline. I think we are still a fair bit away from being ready, so some of the of the 3.x issues may need an earlier release?

yes that was the point of the 3.x release branch but the swith to install4j makes it a bit heavy .... let see how things develop ...

Issue #618, opened 2020-06-26
[BUG]  layout issue in case of multiple screens


Lactoo:
I don't know if this is a bug or a feature request. Anyway, to me it feels like a bug, as I did not have the same issue in HO 1.

When I start HO 3.0 the layout is a bit weird. It has full screen height, but is very narrow.

I can then full screen it and adjust the windows.

The the "player name" and "player details" windows are very narrow and to the left. I then adjust the windows to how I like it.

Is there any way to save the layout so it will look the same the next time I open HO 3.0?

As it is now it resets the layout every time I start HO.

Possible useful note:
I'm using 2 screens. HO opens on the right screen.

Below some screenshots:

1. When I open:
![image](https://user-images.githubusercontent.com/67468579/85858872-9c2ab280-b7bc-11ea-87b3-a509b71dde94.png)

2. After I maximize HO
![image](https://user-images.githubusercontent.com/67468579/85859026-deec8a80-b7bc-11ea-8de6-5f50a833902e.png)

3. After I have adjusted layout to how I want it:
![image](https://user-images.githubusercontent.com/67468579/85859096-faf02c00-b7bc-11ea-8016-57bdfe4ad231.png)

4. After I have closed HO 3.0 and re-opened:
![image](https://user-images.githubusercontent.com/67468579/85859212-25da8000-b7bd-11ea-910b-c182e972029a.png)

**Platform information:**
 - OS: Win10
 - HO 3.0 build 2728 (the latest according to check for update)

akasolace:
The issue still exist on master branch

akasolace:
I just had a quick look. I haven't found the issue et but something look really broken in method load of userconfigurationtable.java

tychobrailleur:
Is there a remote chance this could be the cause of these weird issues disappearing from `USERCONFIGURATION`, sometimes causing the app to refuse to start?

akasolace:
@tychobrailleur  so I know why I observe this on master  ....because of updateDBv400
I have to check if I have this on 3.0

tychobrailleur:
> @tychobrailleur so I know why I observe this on master ....because of updateDBv400

OK.  The deletes in `updateDBv400` are supposed to reset the divider location, is this causing a problem for you on master?

akasolace:
> OK. The deletes in updateDBv400 are supposed to reset the divider location, is this causing a problem for you on master?

[...]

PR #630, opened 2020-07-12
Calculate width with coords of left and top of current screen.

tychobrailleur:
When calculating the width to be saved in the user parameters upon
exiting HO, the width of the screen is calculated as the min of the
current width and the difference between the width of the current
screen minus the current X location of the screen.

In a setup where multiple screens are present, the leftmost X may not
be 0 (e.g. X = 1096), causing the difference `width_screen -
current_pos_x` to be ridiculously small, and being stored as the
current width. (Ditto for height)

This commit offsets the difference with the leftmost X of the screen
to take into account multiple screens.

This commit also:
- translates some the comments in German into English;
- removes the deprecated user parameter `bestPostWidth` which seems to
never be read.

   - fixes issue #618
 
 - [x] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/gui/HOMainFrame.java
+++ b/src/main/java/core/gui/HOMainFrame.java
@@ -117,7 +117,7 @@ public final class HOMainFrame extends JFrame implements Refreshable, ActionList
 	private Vector<JPanel> m_vOptionPanels = new Vector<JPanel>();
 
 	private boolean isAppTerminated = false; // set when HO should be terminated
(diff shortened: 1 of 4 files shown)


akasolace:
@tychobrailleur  I can't test it as I have a single monitor but it seems all good.
I merge it and we will ask for testing beore closing the issue.

aloMalbarez:
Bug Still present if the "current screen" is on the left side.
- it works on screen 3 : right side
- it does not work on screen 2: left side ( maybe coords reported are negative )

image

Issue #623, opened 2020-06-27
[BUG] no training reported?


noujazeg:
Hello,

HO didn't report any training last friday. Concrete: my trainees didn't get any better in their skill i'm training (defending). Nevertheless they all got a higher TSI, so I suspect they did get their training after all. All players did play, although one match was a walkover, because the opponent didn't lineup a full team.

These are the 2 matches in which my 10 trainees played:

https://www97.hattrick.org/Club/Matches/Match.aspx?matchID=661971790&SourceSystem=Hattrick&TeamId=209495&UpdateViewedReport=False&BrowseIds=658785615,661971782,662591943,661971785,662790133,661971790,662954689,661971794,663097408,661971798,661971802,661971805

https://www97.hattrick.org/Club/Matches/Match.aspx?matchID=662954689&SourceSystem=Hattrick&TeamId=209495&UpdateViewedReport=False&BrowseIds=658785615,661971782,662591943,661971785,662790133,661971790,662954689,661971794,663097408,661971798,661971802,661971805

I'm running HO 2.0 on a Windows 10 laptop.

I sorted the players on their defending skill. The first 10 players are my 10 trainees. Overste even declined in his skill, but he's being trained (he's gonna be my coach someday) and normally his skill is increasing a bit every week. 

![2020-06-27](https://user-images.githubusercontent.com/67507299/85922605-f8f1a000-b884-11ea-9e7f-28e3aa81771f.png)

a second screenshot, which includes the training of June 19th, so you can see how it normally looks:

![2020-06-27 (1)](https://user-images.githubusercontent.com/67507299/85922691-a19fff80-b885-11ea-9b62-605d4eca9043.png)

wsbrenk:
first match is a walkover match
second one is saphir cup

noujazeg:
> first match is a walkover match
> second one is saphir cup

yes, but players receive training in matches like that as well.

wsbrenk:
@noujazeg i know - I just wanted to write down what the reason for the error you found might be

but as far as i know, there is no experience increase in case of walkover, isn't?

wsbrenk:
#306 could solve the training of the saphir cup match

noujazeg:
> @noujazeg i know - I just wanted to write down what the reason for the error you found might be
> 
> but as far as i know, there is no experience increase in case of walkover, isn't?

ah, thanks then.

i don't know about the latter, but it would be sad as i probably have 4 walkover matches this season.

[...]

PR #640, opened 2020-07-20
Individual training plans

wsbrenk:
 
   - fixes issue #587 and #623 

changes of branch 623 are included. Feel free to merge them one by one or in a single step.
 
 
 - [x] have been updated
 - [ ] do not require update


akasolace:
@yaute74  I think you are the best person to do this review but if you have no time I will do it as I would like to have it merged before my holiday. Please let me know how you want to proceed.

yaute74:
@akasolace I am sorry, but I will have  not the time actually to do a correct review

akasolace:
> @akasolace I am sorry, but I will have not the time actually to do a correct review

ok no worry I will do it then ...

src/main/java/core/db/DbUtil.java
```diff
@@ -5,10 +5,16 @@
...
 public class DbUtil {
 
-    static public Integer getNullableInt(ResultSet rs, String cl) throws SQLException {
+    public static Integer getNullableInt(ResultSet rs, String cl) throws SQLException {
         Integer res = rs.getInt(cl);
         if (!rs.wasNull()) return  res;
         return null;
     }
 
+    public static Boolean getNullableBoolean(ResultSet rs, String cl) throws SQLException {
```
akasolace, line 14: usually it is good practice to put the @nullabale operator to faciilitate the work of the debugger

src/main/java/core/model/match/MatchStatistics.java
```diff
@@ -91,6 +105,7 @@ public int getTrainMinutesPlayedInPositions(int spielerId, int[] accepted) {
 		// Done with substitutions, add end if necessary
 
 		if (inPosition) {
+			// TODO: if opponent is walkover +=90
```
akasolace, line 108: is it a todo you intended to implement before submiting the PR ? do you want to open a dedicated ticket for this?

src/main/java/core/net/OnlineWorker.java
```diff
@@ -266,6 +266,8 @@ public static boolean downloadMatchData(int matchid, MatchType matchType, boolea
 
 	public static boolean downloadMatchData( MatchKurzInfo info, boolean refresh)
 	{
+		if ( HOMainFrame.launching.get() ) return false; // do not call this on HO launch
```
akasolace, line 269: is this linked to the fix you made regarding skill of player who recently join the team? HO! tries to dtermine how many training the players received since last skill before joigning?  
also I haven't check but could we not simply avoid the function to be called during startup ?

src/main/java/core/training/TrainingManager.java
```diff
@@ -179,79 +177,182 @@ public void recalcSubskills(boolean showBar) {
...
+					// Player appears the first time
+					// - was bought new
+					// - promoted from youth
+					// - it is the first hrf ever loaded
+					old = new Player();
+					old.setSpielerID(player.getSpielerID());
+					old.copySkills(player);
+					if (HOVerwaltung.instance().getModel().getCurrentPlayer(player.getSpielerID()) != null) {
+						// PLayer is in current team (not an historical player)
+						List<TrainingEvent> events = player.downloadTrainingEvents();
```
akasolace, line 212: is this the line which is trigered at staturp and cause match download ?  with the modifictation to match downloader would that stillwork correctly ?

akasolace:
> new database table futurePlayerTraining is ready. FutureTrainingsManager calculation is changed. Since there is no entry in the new table, the results of the training forecast are unchanged.  Now i need a good idea how to implement the editing of the individual future training plans. my very first idea was to add combo boxes to the cells of the training forecast overview table. can someone tell me if that's a stupid idea?

@wsbrenk we agree this is not supposed to be treated in this PR, you just prepare the table but there is still no logic to fill it or use it, right ?

akasolace:
#587 seems closed by this PR for what I can tell, also it seems player have the right colors based on best positions (which in my case is very different to the actual trained positions so looking forward for the individual training plans as you don't want to use last training as a basis for the future as you explained in a different ticket )

akasolace:
I can't check if #623 using my database but I think we could check this one down the road

wsbrenk:
further answers tomorrow

wsbrenk:
> 
> 
> > new database table futurePlayerTraining is ready. FutureTrainingsManager calculation is changed. Since there is no entry in the new table, the results of the training forecast are unchanged.  Now i need a good idea how to implement the editing of the individual future training plans. my very first idea was to add combo boxes to the cells of the training forecast overview table. can someone tell me if that's a stupid idea?
> 
> @wsbrenk we agree this is not supposed to be treated in this PR, you just prepare the table but there is still no logic to fill it or use it, right ?

in the meantime there the popup that fills this table and future training calculation uses this information.

wsbrenk:
> 
> 
> #587 seems closed by this PR for what I can tell, also it seems player have the right colors based on best positions (which in my case is very different to the actual trained positions so looking forward for the individual training plans as you don't want to use last training as a basis for the future as you explained in a different ticket )


@akasolace are you able to select some cells (future training weeks)? If you release the mouse a popup should apear, that let's you select the training priority you want for the selected weeks.

src/main/java/core/db/DbUtil.java
wsbrenk: public static @Nullable Integer getNullableInt(ResultSet rs, String cl) throws SQLException

@akasolace  do you mean it this way? i will change that

wsbrenk:
> Also it is likely I missed it but I did not come across the db intiliaer. Just to be sure that users starting a fresh db will have the new training tables (i.e. same table as the ones going through dbUpdater)

to be honest: i didn't test this either. but i'm optimistic that introducing the table FuturePlayerTrainingTable in DBManager.initAllTables will do the job.

src/main/java/core/model/match/MatchStatistics.java
wsbrenk: this line is left over from my first plan. finally, the treatment of walkover matches is done a few lines above. 
i will remove this line.

src/main/java/core/net/OnlineWorker.java
wsbrenk: I completely agree with you - but I have no idea how this can be achieved. This is triggered by a change in MatchStatistic.getHomeTeamInfo, which I needed to avoid an npe. In my database I found matches with missing lineups ("Verlegenheitstruppe") that I wanted to repair by reloading.
Unfortunately, this method is also called at HO start.

src/main/java/core/training/TrainingManager.java
wsbrenk: see above - but i'm not sure if this method will also call the OnlineWorker

src/main/java/core/db/DbUtil.java
akasolace: yes, and you will need to import org.jetbrains.annotations.Nullable;

src/main/java/core/model/match/MatchStatistics.java
akasolace: ok

src/main/java/core/net/OnlineWorker.java
akasolace: @wsbrenk ok maybe then you can add a todo explaining the issue for when a dev would have some time ... without explanation it will be quite confusing ....

src/main/java/core/training/TrainingManager.java
akasolace: ok let just put a todo as mentioned above and we can leave it as is for the time being ....

akasolace:
> to be honest: i didn't test this either. but i'm optimistic that introducing the table FuturePlayerTrainingTable in DBManager.initAllTables will do the job.

If you put it in initAllTables  it should do the job ...

akasolace:
> @akasolace are you able to select some cells (future training weeks)? If you release the mouse a popup should apear, that let's you select the training priority you want for the selected weeks.

I can select the training type but I am not sure this is whawt you are talking about ....

image

but I can 't find a way to select whic player recieve which kind of training (full/partial), is this doable ? As already mentioned in my case the detection based on best position is completly inaccurate and I think it will be the case for plenty of users ....

wsbrenk:
@akasolace sorry, no you have to do it in the prediction panel:

grafik

akasolace:
> @akasolace sorry, no you have to do it in the prediction panel:
> 
> 
> 
> grafik
> 
> 

Ah ok i will try again later

akasolace:
> @akasolace sorry, no you have to do it in the prediction panel:
> 
> grafik

ah ok I found it no but I have some  remarks:
      - I think we should add this contextual menu to the Training tab to set the full / partial for the full period (the prediction panel could be used for fine tuning for a specific week but I think that usually users will want to set it for the full future)
      - in the prediction panel switching to full / partial / osmoesis works well but swithing to none is not working

wsbrenk:
> - I think we should add this contextual menu to the Training tab to set the full / partial for the full period (the prediction panel could be used for fine tuning for a specific week but I think that usually users will want to set it for the full future)

the intention was to select training priorities for each player individually. it is not realistic/virtualistic to set full training for all of your players, because the number of available trainings positions will be limited.

> - in the prediction panel switching to full / partial / osmoesis works well but swithing to none is not working

that's a bit tricky. we have to distinguish between reset of the individual training setting for one player and setting NO training. in the moment the first is implemented and by resetting the individual training the automatic training depending on players best position is in place.

wsbrenk:
> usually users will want to set it for the full future)

now i got your point. you want to select the priority in training panel for the whole time range by overriding the best position information via popup menu. that could be done i think. good idea!!!

akasolace:
> > usually users will want to set it for the full future)
> 
> now i got your point. you want to select the priority in training panel for the whole time range by overriding the best position information via popup menu. that could be done i think. good idea!!!

yes it is exactly that ....

wsbrenk:
this commit moves popup to an extra class, that i want to use in training panel too.

No training and resetting to best position training are separated items now.

src/main/java/core/net/OnlineWorker.java
wsbrenk: @akasolace downloadMatchData is no longer called on HO launch. I removed this line.

src/main/java/core/training/TrainingManager.java
wsbrenk: @akasolace line is removed

wsbrenk:
@akasolace now you can select training prio in training panel too:

grafik

build.gradle
```diff
@@ -147,8 +147,7 @@ distributions {
 
 
 install4j {
-    if (install4jHomeDir == null) installDir = file("foo")
-    else installDir = file(install4jHomeDir)
+    if (hasProperty('install4jHomeDir'))  installDir = file(install4jHomeDir)
```
akasolace, line 150: I have to revert this change as it breaks my gradle task.
It seems install4j refuse to run if during the configuration phase variable installDir is not set this is why it has to be also specified in else clause

wsbrenk: > 
> 
> I have to revert this change as it breaks my gradle task.
> It seems install4j refuse to run if during the configuration phase variable installDir is not set this is why it has to be also specified in else clause

hi @akasolace : could you try it this way: 

    if (hasProperty('install4jHomeDir'))  installDir = file(install4jHomeDir)
    else installDir = file("foo")

this would work in my case.

akasolace: @wsbrenk arf not yet but the following should (hopefully) work for both of us :-)
```
    if ((hasProperty('install4jHomeDir')) || ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
    else installDir = file("foo")
```

wsbrenk: nope: A problem occurred evaluating root project 'HO'.
> path may not be null or empty string. path=''

what's about this on your side?

    if (hasProperty('install4jHomeDir')){
        if (install4jHomeDir != null) installDir = file(install4jHomeDir)
        else installDir = file("foo")
    }

Issue #648, opened 2020-07-28
[BUG] Release Channels Panel, darklaf related ? 


akasolace:
**Describe the bug**
Display error

**To reproduce**

File -> Preference -> Release channels

tychobrailleur:
Yes, more than likely: this doesn't happen with Nimbus.  I am creating a PR.

PR #649, opened 2020-07-28
Set a preferred size to avoid display issues.

tychobrailleur:
For some reason (which I have no time to look into), when
displaying "Release Channel" panel in options, the text area's width
is being drawn by darklaf with an excessive width, causing display
issues in the panel.

Setting the preferred size on the text area solves the problem.

This PR also adds an empty border on the panel to avoid the help text to be set against the left side.

Screenshot from 2020-07-28 18-40-54

   - fixes issue #648

 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/option/ReleaseChannelPanel.java
+++ b/src/main/java/core/option/ReleaseChannelPanel.java
@@ -3,16 +3,10 @@ package core.option;
 import core.model.HOVerwaltung;
 import core.gui.comp.panel.ImagePanel;
 
-import java.awt.Insets;
-import java.awt.GridBagLayout;
-import java.awt.GridBagConstraints;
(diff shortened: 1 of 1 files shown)


akasolace:
so quick .. awesome !

weisJ:
> For some reason (which I have no time to look into), when
> displaying "Release Channel" panel in options, the text area's width
> is being drawn by darklaf with an excessive width, causing display
> issues in the panel.
> 
> Setting the preferred size on the text area solves the problem.

This is connected to [this issue](https://github.com/weisJ/darklaf/issues/201) which should be fixed with the next version. I can confirm though that setting the preferred size will absolutely solve the problem.

Issue #660, opened 2020-08-11
[FEATURE] support of man-marking order


wsbrenk:
![manmarkingOrder](https://user-images.githubusercontent.com/11821502/89227524-221cf300-d5de-11ea-94a9-4fad0a7993ca.png)

Currently teamanalyzers team view is my favorite. a 4th button in SubstitutionOverview could start the team analyzer and some drag+drop action could create the man marking match order. But i don't have any idea how to implement that - up to now.

would be great if you could help me.

_Originally posted by @wsbrenk in https://github.com/akasolace/HO/issues/632#issuecomment-668243486_

akasolace:
@wsbrenk I haven't followed everything regarding man marking order and it is a functionality I only used once or twice in game. 
However, do we really need a new GUI for this? can't we not simply list the players in opponent team like what HT is doing? Relying on last position might be misleading, no?

wsbrenk:
@akasolace it is not as easy as it looks like. in lineup panel, there is no access to opponents information until now. the idea is to use the existing team analyzer view. Editing the lineup (man marking order in substitution view) has to set lineup information in teamanalyzer and the rest of the man marking editing has to happen in team analyzer.

akasolace:
@wsbrenk it sounds very complicated and I would have clearly taken a much simplified approach: I would have done it all in the order tabs, a new man marking order along substitution and replacement .... simply with a dropdown listing opponent players. 

> in lineup panel, there is no access to opponents information

I think there are 2 ways to deal with it:

 1. at data download we fetch list of players for upcoming opponenent and that information is stored in a hashmap like 5 teamIDs and 5 list of strings
 2. in lineup tab we add a field for opponent : 
   - if selected info are populated accordingly (home/away, TS, .....)
   - have possibility to put 'sandbox' to keep existing behaviour (which is lineup simulator in HT)


For sure, I would not have done any GUI .... but it is up to you :-)

wsbrenk:
@akasolace up to now i really don't know what's the heavier way. as far as i understand you right, we have to double all or nearly all the download stuff that is already done by the team analyzer. the gui part proposed by you is easier to implement, that's right.

akasolace:
> as far as i understand you right, we have to double all or nearly all the download stuff that is already done by the team analyzer

I might be wrong but I think it is quite straightforward: we can reuse the parser of players.xml but we will just need to store the list of opponent player last names (similar to HT).

akasolace:
[...]

PR #676, opened 2020-08-31
fix recursion in player overview panel (HOMainFrame constructor)

wsbrenk:
as mentioned in #660 

 - [ ] have been updated
 - [x ] do not require update

sorry @akasolace, again i changed the build.gradle

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -152,7 +152,7 @@ distributions {
 
 
 install4j {
-    if ((hasProperty('install4jHomeDir')) || ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
+    if ((hasProperty('install4jHomeDir')) && ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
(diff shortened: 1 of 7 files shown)


akasolace:
@wsbrenk can you highlight which part exactly was causing the recursion. Lokking at the PR I don't see that ...

wsbrenk:
HOMainFrame init called init player overview panel
called   getTraining
called  getMatchdetails.lastMinute
called  OnlineWorker.downloadMatch (if Match details have to be updated because of new columns)
called WaitDownloadDialog
called HOMainFrame.instance() ...

(the complete stack trace is given in #669)

solution was not to call WaitDownloadDialog when HOMainFrame is launching. 
then matches panel needs some fixes because of wrong initialized MatchDetail objects

tychobrailleur:
I remember fixing a similar problem in d692021bbb50552cfc340d7ac96597b586456791

Issue #539, opened 2020-05-30
[BUG] Training Plugin - Analyser and Effekt tab


Sturzbach:
**Describe the bug**
There is a bug, which invalidates trainings data since Update to Version 3.0 i Think.
In a old Database everything is alright and the Data are shown with correct entries for Season and week. But then I import data now, the Acutal data are shown as of the season before.
This is the fact in Tab Training / Analyse and Effect.

The result is i think it does not caslculate the training effect correctly in this tab.
In main player attribute it seems to b correctly.

**To Reproduce**
I tokk my database from before the update and there it shows data for my players for Season 62 (Scotland  IV.31).
Then i have imported now data from today, int the traings analyser tab and effect tab i got data shown named season 61.
Also effect is empty although i had 3 Skillups on thursday and week 16 is not shown.

**Screenshots**
Good data from 21.5.2020
![grafik](https://user-images.githubusercontent.com/59139029/83326368-d8c5c580-a273-11ea-9f6d-22dfbcca0088.png)

Bad Data after import from today
![grafik](https://user-images.githubusercontent.com/59139029/83326505-e3cd2580-a274-11ea-9d4e-abf9637868bd.png)

**Platform information:**
Windows 10 Home 1909
Version 3.000

akasolace:
@wsbrenk  I assigned it to you as you are the most experienced with that part of the code.
I let you check whether or not we have an issue here and what should be the appropriate labels and milestone...

wsbrenk:
currently i could not reproduse this effect, but i remember that i saw it some days ago. i guess it is a week 16 effect.

akasolace:
ok lets look again at this one on week 16 then ...

wsbrenk:
OldTrainingManager uses trainings date to initialize PastSkillChange object. this uses HTCalendarFactory.getHTSeason(...,useTrainingCalendar=false) to set HtSeason and HTWeek. Don't know if using the economyCalendar could explain this problem.

akasolace:
@wsbrenk sorry but I am not familiar at all with that part of the code... but I don't really understand why a function getHTSeason would return different values for HtSeason and HTWeek whether or not useTrainingCalendar is set to True

akasolace:
@wsbrenk for a different reason I was considering a refactoring of HTCalendar and HTCalendarFactory  ( #678).  I am planning to do that for v5.0.  It would be nice to have your inputs for what you would need for the training manager. 
Also:
[...]

PR #683, opened 2020-09-02
#539 fix ht season correction

wsbrenk:
 
   - fixes issue #539  
    
 - [x] have been updated
 - [ ] do not require update

How to verify this pull request:

in week 16 downloads

1. before economy date HO V3 and V4(with this PR) should show HTSeason 75 in training effect table
2. after economy date, but before training date HO 3 shows HTSeason 74; V4(with PR) correctly shows HTSeason 75
3. if i'm wrong with 2.; after training date but before match date HO 3 shows HTSeason 74; V4(with PR) correctly shows HTSeason 75.
4. after Match date both versions will show correctly HTSeason 76

### DIFF
--- a/src/main/java/core/util/HTCalendarFactory.java
+++ b/src/main/java/core/util/HTCalendarFactory.java
@@ -22,8 +22,6 @@ public final class HTCalendarFactory {
      * Creates a HTCalendar to calculate local values for a league,  using the economy date to flip
      * over to the next week.
      *
-     * @param model HO data model
-     *
      * @return a HTCalendar.
      */
     public static HTCalendar createEconomyCalendar() {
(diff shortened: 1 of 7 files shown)


akasolace:
@wsbrenk  I still not have a clear understanding of how this issue should be treated. Information is confusing and not consistent between the website (week and season always start in Monday), the xml and between the HTdevs answers

For sure league game are played Saturday and Sunday. But I think some leagues (brazil ?) even have game on Monday which if it is the case makes really confusing to start the HT week on Monday.  We could decide to base HO! on a different calendar that would start with the economic update. But are we sure the economic update starts at the same time for everyone? If it is not the case we need a calendar per league. The only issue, would be that we will display an HTWeek different in HO! that what we have on the website but it might be the only solution to have something robust. 
I will try again to clarify things with HTdevs, I will also try to collect few worldetails.xml around the end of week 16 to confirm our assumptions. What is certain is that HTCalendar needs a refactoring.

wsbrenk:
> For sure league game are played Saturday and Sunday. But I think some leagues (brazil ?) even have game on Monday which if it is the case makes really confusing to start the HT week on Monday. We could decide to base HO! on a different calendar that would start with the economic update. But are we sure the economic update starts at the same time for everyone? If it is not the case we need a calendar per league. The only issue, would be that we will display an HTWeek different in HO! that what we have on the website but it might be the only solution to have something robust.
> I will try again to clarify things with HTdevs, I will also try to collect few worldetails.xml around the end of week 16 to confirm our assumptions. What is certain is that HTCalendar needs a refactoring.

I think we have to know, when worlddetails flip the season - i think it will be the match date. and i think we need the season number from worlddetails too, since not all countries are in season 75 as we are, so the season correction has still to be calculated in future. instead of arranging the htseason calculation on economy date, as it is today, we have to use the match date, which will of cource be different in different countries.
And of course the bug is only happening for a few hours around the season flip.

wsbrenk:
on the other hand, there is a SeasonOffset in worlddetails. if that is the same as our calculated ht season correction, we could use this and then it is our choice, when to flip the seasons.

Issue #679, opened 2020-09-01
[FEATURE]  leagueOffset need to be store in database


akasolace:
- [ ] leagueOffset need to be store in database (XTRADATA) It needs to be retrieved only once from worlddetails.xml
might be part of misc.basics information, it will be accessed from HTCalendar

 - [ ] similarly at that time an offset for economy, match and training data need to be calculated

Training Date is used for Training Manager, @wsbrenk is matchdate also used or do you get that info somewhere else?
Similarly I am not sure that at the moment economydate is used but it would be useful when #530 is implemented (would trigger I imagine delta in finance .....)
 
		columns[4]= new ColumnDescriptor( "SeriesMatchDate",	Types.TIMESTAMP,false );
		columns[5]= new ColumnDescriptor( "TrainingDate",		Types.TIMESTAMP,false );
		columns[6]= new ColumnDescriptor( "EconomyDate",		Types.TIMESTAMP,false );

wsbrenk:
download of SeasonOffset from worldetails is done: same procedure as for Season (Saison). this info is in the BASICS part now:

[basics]
application=HO
appversion=0.0
date=2020-09-13 14:51:48
season=75
seasonOffset=0
matchround=16
teamID=520472
teamName=Juventus Brenk
activationDate=2004-09-25 02:57:00
owner=wsbrenk
ownerHomepage=
countryID=3
leagueID=3
regionID=238
hasSupporter=True

wsbrenk:
![image](https://user-images.githubusercontent.com/11821502/93018702-e8bea680-f5d1-11ea-8b4a-935e5042127c.png)

wsbrenk:
`    /*
    alternative to getHTSeasonCorrection based on worlddetails info SeasonOffset
    (not used up to now)
     */
    private static int getSeasonOffset(){
    	return HOVerwaltung.instance().getModel().getBasics().getSeasonOffset();
	}
`

akasolace:
@wsbrenk this is something you recently added ?

wsbrenk:
@akasolace yes, with the first item of this issue i'm just ready and i will push it soon. with the second one i'm still struggling. i could easily produce a static method that would examine the time difference of those dates to monday 0 AM. but i don't see any use of it!?!

- if those dates would change in the future, any calculation of those dates will be wrong
- estimations of future economy or trainings date could easily be done by adding weeks to the current date from xtra_data

PR #693, opened 2020-09-13
#679 HTCalendarFactory.getSeasonOffset

wsbrenk:
 
   - fixes issue #679  
   
 
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -152,7 +152,7 @@ distributions {
 
 
 install4j {
-    if ((hasProperty('install4jHomeDir')) || ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
+    if ((hasProperty('install4jHomeDir')) && ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
     else installDir = file("foo")
     disableSigning = true
 }
--- a/build.gradle
+++ b/build.gradle
@@ -152,7 +152,7 @@ distributions {
 
 
 install4j {
-    if ((hasProperty('install4jHomeDir')) && ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
+    if (install4jHomeDir != null && install4jHomeDir != "") installDir = file(install4jHomeDir)
     else installDir = file("foo")
     disableSigning = true
 }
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -152,7 +152,7 @@ distributions {
 
 
 install4j {
-    if ((hasProperty('install4jHomeDir')) || ((install4jHomeDir != null) )) installDir = file(install4jHomeDir)
+    if (install4jHomeDir != null && install4jHomeDir != "") installDir = file(install4jHomeDir)
```
akasolace, line 155: ok let's try this version :-)

src/main/java/core/db/DBUpdater.java
```diff
@@ -123,6 +123,10 @@ private void updateDBv400(int dbVersion, int version) throws SQLException {
 		m_clJDBCAdapter.executeUpdate("DELETE FROM USERCONFIGURATION WHERE CONFIG_KEY = 'aufstellungsPanel_horizontalRightSplitPane'");
 		m_clJDBCAdapter.executeUpdate("DELETE FROM USERCONFIGURATION WHERE CONFIG_KEY = 'aufstellungsPanel_horizontalLeftSplitPane'");
 
+		if ( !columnExistsInTable("SeasonOffset", BasicsTable.TABLENAME)){
+			m_clJDBCAdapter.executeUpdate("ALTER TABLE BASICS ADD COLUMN SeasonOffset INTEGER");
```
akasolace, line 127: here should we put a default value to 0 to be on the safe side ?

wsbrenk: @akasolace why do you think a default 0 would be a good idea?

akasolace: @wsbrenk we had quite a lot of problems with the database, when entering new column and having some null values. 
I see you treat the case with a catch/try block but maybe it is safer to avoid the null value all together.
Otherwise, you can put a dummy value like -1 and if you have -1 when getting the value, the code try to redownload worlddetails .... but I leave it up to you ... your call

wsbrenk: -1 would be quiet better than 0, since 0 is a valid number, which gives in countries other than sweden, france or germany and some others wrong information. but why isn't null not the perfect solution for that?
the try catch block you mentioned is a copy paste relict, i'm not very happy with.

akasolace: @wsbrenk ok fine, let's keep it as is ..... I go ahead with the merge then ....

Issue #705, opened 2020-09-28
[BUG] Release Channel not stored in userconfig table


akasolace:
**Describe the bug**

**To Reproduce**

**Screenshots**

**Platform information:**
 - OS: [Windows / macOS / Linux]
 - Version (including build number)

**Additional context**

akasolace:
it is actually not fixed

![bug705](https://user-images.githubusercontent.com/1136496/94850954-fd030000-0427-11eb-9660-dfd457d37122.gif)

wsbrenk:
OK, i thought it would be an update issue

Issue #667, opened 2020-08-26
[BUG] if not specified by user, the release channel should be based on version type 


akasolace:
**Describe the bug**
As noticed by @tychobrailleur a fresh build of HO! always default release channel to stable but it should be based on version type. A user downloading a DEV version would expect the releasechannel to point to DEV by default.

Quickly browsing the code, it seems the following need to be modified:

**DB initlization**
https://github.com/akasolace/HO/blob/23d8b3299ba3a773550bc01963cacd54c0c6ff01/src/main/java/core/model/UserParameter.java#L99

could also be modified in DB updater but not mandatory
https://github.com/akasolace/HO/blob/d844ed396cc17ca65460c6eac42987a7668cadb5/src/main/java/core/db/DBUpdater.java#L1168

PR #709, opened 2020-10-01
#705 edit release channel config

wsbrenk:
 
   - fixes issue #705 
   
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/model/UserParameter.java
+++ b/src/main/java/core/model/UserParameter.java
@@ -429,12 +429,6 @@ public final class UserParameter extends Configuration {
         map.put("aufstellungsAssistentPanel_verletzt", String.valueOf(aufstellungsAssistentPanel_verletzt));
         map.put("xmlDownload", String.valueOf(xmlDownload));
 
-        if (HO.getVersionType() != null) {
-            switch (HO.getVersionType()) {
-                case "DEV" -> ReleaseChannel = "Dev";
-                case "BETA" -> ReleaseChannel = "Beta";
-            }
-        }
         map.put("ReleaseChannel", ReleaseChannel);
 
         map.put("fixtures", String.valueOf(fixtures));


akasolace:
@wsbrenk this will work indeed but it reopens #667

wsbrenk:
> @wsbrenk this will work indeed but it reopens #667

@akasolace i would have thought that the commit https://github.com/akasolace/HO/commit/ef6fc65c6a62cc60a5b04f552683f18c2435289f would handle that, isn't it?

akasolace:
> > @wsbrenk this will work indeed but it reopens #667
> 
> @akasolace i would have thought that the commit [ef6fc65](https://github.com/akasolace/HO/commit/ef6fc65c6a62cc60a5b04f552683f18c2435289f) would handle that, isn't it?

I am not sure this will have the intended effect ... I think it will change release channel of users during upgrade => which we do not want to do (if a user is on dev channel and upgraded to latest stable/beta he will still want to have his release channel on dev)

The bug #667 is only about new artefacts (all new artefacts, dev, beta or stable have release channel pointing to stable, whereas it should point to de version type).  I think the approach would be to have:
   -   a value bool releaseChannelSET = false, set as default in userparameter. 
   -   code setting release channel:
       - if releaseChannelSET  read from preference panel
       -  else allocated from version 

when the release channel is modified put     releaseChannelSET   = true

it is a bit the logic I had when I tried to clause #667 but I forgot the bool releaseChannel so effectively I was resetting the channel at each close.

@wsbrenk  not sure if all this is clear for you ... if you want to have a look please go ahae

wsbrenk:
@akasolace as far as i understand, the update4000 method will work as desired if a user starts with a fresh installation of HO. Depending on the artefact the channel is initialized as required (i hope).

Only if the user is working with an existing database and release channel is set to any value, this value remains as it is. this could be an unintentional one, initialized by a wrong updateconf method of an old HO version. But i think we have no chance to destinguish if the given value was set intentional or not.

akasolace:
I was under the impression that if the user start with fresh install release channel will always be setup to Stable (even for a beta or a dev build). For existing user going through updateDB400 it will réinitialise their release channel based on HO version (erasing their choice). I might be wrong though as you sound convinced this will work :-)

wsbrenk:
@akasolace i think you're right. I should not state anything on saturday morning;-(

akasolace:
😃

PR #716, opened 2020-10-04
fixes #667. Test is successful.

### LINKED ISSUE #667: [BUG] if not specified by user, the release channel should be based on version type

akasolace:
 
   - closes #667 
   
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -1075,29 +1075,6 @@ final class DBUpdater {
 			HOLogger.instance().log(getClass(), "Updating configuration to version 1.436...");
 			updateConfigTo1436(HO.isDevelopment() && lastConfigUpdate == 1.436);
 		}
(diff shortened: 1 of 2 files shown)


wsbrenk:
Tested it successful.

PR #721, opened 2020-10-10
Display man marking rebased

wsbrenk:
next step of displaying man marking order in the teamanalyzer panel.

same as previous pull request, but rebased commits.

### DIFF
--- a/src/main/java/module/teamAnalyzer/ui/TeamPanel.java
+++ b/src/main/java/module/teamAnalyzer/ui/TeamPanel.java
@@ -125,17 +125,21 @@ public class TeamPanel extends JPanel {
                 var manMarker = manMarkingOrder.getSubjectPlayerID();
                 var manMarkerPos = ownLineup.getPositionBySpielerId(manMarker).getId();
                 var manMarkedPos = teamLineup.getPositionByPlayerId(manMarkingOrder.getObjectPlayerID());
+                var from = lineupPanel.getMyTeam().getPanel(manMarkerPos);
+                if ( manMarkingOrderDisplay == null){
+                    manMarkingOrderDisplay = new ManMarkingOrderDisplay(grassPanel);
+                }
                 if ( manMarkedPos > 0 ){
-                    var from = lineupPanel.getMyTeam().getPanel(manMarkerPos);
                     var to = lineupPanel.getOpponentTeam().getPanel(manMarkedPos);
-                    if ( manMarkingOrderDisplay == null){
-                        manMarkingOrderDisplay = new ManMarkingOrderDisplay(grassPanel);
-                    }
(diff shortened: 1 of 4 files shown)


wsbrenk:
image

akasolace:
I put it for 5.0 but you will have to explain it to me.
Still don't understand how this is supposed to work as the man we might want to mark was not aligned in previous lineup. ..

wsbrenk:
@akasolace hi, great to see you back. Everything OK on your site?

Concerning your question: if the marked opponent is not in the current selected lineup the red arrow goes to (the ass of) the marking defender. Kind of displaying the negative impact of such constellation.
I think the feature is not dangerous. So i suggested to include it. 
And it is include in the both other pull requests. So this one i wanted to delete - but i didn't find a way to do it.

akasolace:
hi @wsbrenk yes all good on my side. Just have been quite busy lately: more work and I more sport on my free time :-).

So you want me to close this PR, right ?

akasolace:
if you are 100% positive and if you insist I will include it before the release. But honestly I would rather wait for next version ... let me know

wsbrenk:
@akasolace i do not insist. it is of course a not completly ready feature. So it is not a problem for me to move it to the next version.

But there is already a more incomplete part of it on the master branch. So consequently we should  remove it completly.

src/main/java/module/teamAnalyzer/ui/TeamPanel.java
```diff
@@ -118,27 +118,31 @@ public void reload(TeamLineup teamLineup, int week, int season) {
...
+                        var to = lineupPanel.getOpponentTeam().getPanel(manMarkedPos);
+                        manMarkingOrderDisplay.set(from, to);
+                    } else {
+                        // TODO: Display warning about failed man marking order
+                        manMarkingOrderDisplay.set(from, from);
+                    }
+                } else if (manMarkingOrderDisplay != null) {
+                    manMarkingOrderDisplay = null;
                 }
             }
```
wsbrenk, line 144: this block should be commented out, it we want to remove the feature completly

src/main/java/module/teamAnalyzer/ui/TeamPanel.java
```diff
@@ -263,7 +267,7 @@ private void fillPanel(JPanel panel, PlayerPanel playerPanel) {
         	// But leave a box the size of a player panel...
         	Box box = new Box(BoxLayout.X_AXIS);
         	box.setPreferredSize(playerPanel.getDefaultSize());
-        	panel.add(new javax.swing.Box(BoxLayout.X_AXIS));
+        	panel.add(box);
```
wsbrenk, line 270: this block is useful to fix a minor "bug" in the display of the lineups

Issue #751, opened 2020-10-26
[BUG] NumberFormatException on download of complete match history 


wsbrenk:
murko wrote:

In the log file with todays date I had a lot of these errors:
15:34:15 [Error] core.net.OnlineWorker: downloadMatchData: Error in downloading match: java.lang.NumberFormatException: null

wsbrenk:
yesterday i fixed the exception - now i see an endless recursion if i want to try to download my matches since the beginning.

akasolace:
can I llok at it somewhere in your branch ?

wsbrenk:
https://github.com/wsbrenk/HO/tree/%23751 

no progress today;-(

akasolace:
I just ran your branch, I downloaded new data and I did not experience the bug you are referring too. I just notice some missing/incorrect match event I opened tickets #756 and #757 for those 

Otherwise I saw strange message like a lot of message like:
   -[Debug]   XMLMatchdetailsParser: XMLMatchdetailsParser[665979475]: Cannot parse matchreport from matchdetails, lineup MUST be available!

I did not have time to investiguate but it does not seems normal. However the match reports for those games looks normal at first look ....

So:
  1) do you have a clue of which kind of games can throw the errors. If not could I do some test with your db ? 
2) do you have a clue about the error:  Cannot parse matchreport from matchdetails, lineup MUST be available!  ?

wsbrenk:
"AWT-EventQueue-0@2042" prio=6 tid=0x15 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at core.net.OnlineWorker.downloadMatchData(OnlineWorker.java:279)
	  at core.net.OnlineWorker.downloadMatchData(OnlineWorker.java:261)
	  at core.model.match.Matchdetails.getHighlights(Matchdetails.java:888)
	  at core.file.xml.XMLMatchdetailsParser.parseMatchReport(XMLMatchdetailsParser.java:488)
	  at core.file.xml.XMLMatchdetailsParser.createMatchdetails(XMLMatchdetailsParser.java:50)
	  at core.file.xml.XMLMatchdetailsParser.parseMachtdetailsFromString(XMLMatchdetailsParser.java:27)
	  at core.net.OnlineWorker.fetchDetails(OnlineWorker.java:786)
	  at core.net.OnlineWorker.downloadMatchData(OnlineWorker.java:368)
	  at core.net.OnlineWorker.downloadMatchData(OnlineWorker.java:261)
[...]

PR #760, opened 2020-10-30
#751download match history

wsbrenk:
 
   - fixes issue #751 
   
 - [ ] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/file/xml/XMLMatchLineupParser.java
+++ b/src/main/java/core/file/xml/XMLMatchLineupParser.java
@@ -128,12 +128,11 @@ public class XMLMatchLineupParser {
 
 		// nur wenn Player existiert
 		if (spielerID > 0) {
-			tmp = (Element) ele.getElementsByTagName("FirstName").item(0);
-			name = tmp.getFirstChild().getNodeValue();
-			tmp = (Element) ele.getElementsByTagName("LastName").item(0);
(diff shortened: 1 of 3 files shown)


src/main/java/core/file/xml/XMLMatchLineupParser.java
```diff
@@ -128,12 +128,11 @@ private static MatchLineupPlayer createPlayer(Element ele) {
...
-			name = tmp.getFirstChild().getNodeValue();
-			tmp = (Element) ele.getElementsByTagName("LastName").item(0);
-			if (tmp.getFirstChild() != null) { // there are players without a lastname
-				name = name + " " + tmp.getFirstChild().getNodeValue();
-			}
+			// First- and LastName can be empty
+			name = getStringValue((Element)ele.getElementsByTagName("FirstName").item(0));
+			var lastName = getStringValue((Element)ele.getElementsByTagName("LastName").item(0));
+			if ( name.length()>0 && lastName.length()>0) name = name + " ";
+			name = name + lastName;
```
akasolace, line 135: is there not an else block missing ?

wsbrenk: no, it isn't. 

maybe we should create the name with a StringBuilder?!  This could be better readable. is it the better way to do string concatination?

Issue #766, opened 2020-11-02
[BUG] Thematize specialty icons


akasolace:
**Describe the bug**
Specialty ions are at the moment not properly visible with dark themes

PR #768, opened 2020-11-02
wip thematize specialty icons

akasolace:
 
   - fixes issue #766
   
 - [ ] have been updated
 - [x] do not require update


src/main/java/core/gui/theme/ImageUtilities.java
```diff
@@ -581,6 +581,19 @@ public static Icon getSmileyIcon(String smileyName) {
 		return null;
 	}
 
+	public static Icon getSmallPlayerSpecialtyIcon(String playerSpecialtyName) {
+		return getPlayerSpecialtyIcon(playerSpecialtyName, 15);
+	}
+
+	public static Icon getPlayerSpecialtyIcon(String playerSpecialtyName, int size) {
+		if (Arrays.stream(HOIconName.SPECIALTIES).skip(1).anyMatch(playerSpecialtyName::equals)) {
```
tychobrailleur, line 589: I believe `ThemeManager` was caching those images, we probably should make sure those ones are cached as well, as it has caused display perf issues in the past.

akasolace: I have another issue at the moment with this one that I can't explain in Lineup tab.

**Before:**

image


With this commie:
image


The bottom line is truncated. I can't figure out what I did that is causing it. Looks like the preferred size is now wrongly calculated.

akasolace: if you have an idea of what is going on ....

tychobrailleur: I have the same problem after switching to darklaf 2.5.x -- reverting back to 2.4.x fixes the issue.

akasolace: >I have the same problem after switching to darklaf 2.5.x -- reverting back to 2.4.x fixes the issue

indeed you are right ...  @weisJ any idea what is going on?

akasolace: > I believe ThemeManager was caching those images, we probably should make sure those ones are cached as well, as it has caused display perf issues in the past.

good catch ... I added the caching

akasolace:
@tychobrailleur I think it is ready for merge. In parallel I created #771 for the rendering of playerCB

PR #807, opened 2020-11-28
Upgrade to darklaf 2.5.3

tychobrailleur:
Also cleanup by moving mouse listener creation in `addListeners`.

Upgrade darklaf to most recent version, and a bit of cleanup.
 
 
 - [ ] have been updated
 - [x] do not require update


akasolace:
Hello, you changed your mind ? Master was already on latest version of darklaf but what about the second part of your PR ?

tychobrailleur:
I had forgotten to sync with upstream (somebody's a bit rusty...), it looks like it is no longer relevant...

akasolace:
If you are up to it, there are couple of message on the forum regarding a bug on mac version. Apparently lot of users experience it: need to put CHPP code at each start of the application.

tychobrailleur:
So the OAuth access token is not preserved or something?

akasolace:
Yes it seems it is not stored. It only affect mac build so I can't test

> I am using HO 4.1 on the mac. My problem is that HO doesn't remember any settings inluding the CHPP-Code. Obviously the settings file is read only. What do I have to change?
Thanks in advance

>> Do you mean like every time you download info you need to give authorization?
If yes, I have same problem on Mac.

Issue #367, opened 2020-01-29
[eFEAT] Youth Academy


wsbrenk:
Hattrick Youthclub is a great tool for youth academy, but i cannot integrate selected youth player to the future trainings plan of my senior team. This could be done in HO.
Additional we could show the complete history of our homegrown players, not only the time, when they were adults.
Some overviews in the youth club could be improved, I think. When I train my youth for the market, only the skills that can be achieved with 17.0 count. this is not shown by the youthclub, but could be done by HO.

akasolace:
@wsbrenk you basically suggest to have a module youth academy within HO?

wsbrenk:
Yes, this would be my idea, that i would like to discuss. since i'm managing 3 teams of homegrowns, i'm missing the integration of my youth team application with HO.On the other hand, I accept that it would be a lot of work to do. Finally, the HO application could eventually become too fat, including the database.

akasolace:
> Yes, this would be my idea, that i would like to discuss. since i'm managing 3 teams of homegrowns, i'm missing the integration of my youth team application with HO.

I agree that it would be a nice addition, quite often requested

>On the other hand, I accept that it would be a lot of work to do.

I do not have a YA myself, but last time I played (few year ago) I had one, and for what I recall HYC was doing a lot of different things, so yes I guess this would be a lot of work !

>  Finally, the HO application could eventually become too fat, including the database.

I am not too concern by this. Also does HYC keeps fully history o youth players incl games ?

It is a topic that defintly worth to be discussed.

wsbrenk:
> Also does HYC keeps fully history o youth players incl games ?

No, i don't think so. HYC only displays the current status of skills. There are progress calculations and only Money is managed historically.

akasolace:
ok

yaute74:
With foxtrick the infos are automatically send to HYC it's a big advantage. With HO we must do all the updates manually, so I am not sure that a lot of player will use it if we integrate it.

wsbrenk:
@yaute74 hi, i don't understand what you mean with "we must do all the updates manually". i think it is possible to add a new selection box "youth academy" to the download filter and every time the user will download (F11), the youth data is updated automatically with last match data and so on. 

on the other hand i'm with you. if we do not integrate some feature, that is not already handled by HYC, then no one will use it. 
and yes, i think it would be a good idea to integrate some manual input options, for example entering the starting skills of the players.

[...]

PR #817, opened 2020-11-30
#367youth

wsbrenk:
 
start of #367


akasolace:
@wsbrenk you moved to java 15, any specific reason ?

wsbrenk:
@akasolace someone mentioned that the flicker issue could be caused by java - so i tried the new version with no success and no visible negative impacts. 

we can easily change back to java 14 - there are no changes in source necessary

PR #818, opened 2020-11-30
#367youth

wsbrenk:
- fixes DBManager.getInteger and
   - Exception if Matchdetails were missing in db because of download errors


akasolace:
could you open a PR just for this one after updating our master branch ?

PR #847, opened 2021-01-09
Add alltid cert to truststore.

tychobrailleur:
Also Base64-encode truststore password


akasolace:
@tychobrailleur  thank you !  I would like to have one of those days few explanation on how all this work. Have a nice we !

PR #854, opened 2021-01-17
Solve issue with image loading in darklaf.

tychobrailleur:
 
Explicitly retrieve image when darklaf is used, as image only gets loaded after interacting with UI.
 
 
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/module/lineup/ratings/LineupRatingPanel.java
+++ b/src/main/java/module/lineup/ratings/LineupRatingPanel.java
@@ -797,7 +797,7 @@ public final class LineupRatingPanel extends RasenPanel implements core.gui.Refr
         m_dRightAttackRating = previousRatings.getRightAttack().get(t);
         m_dCentralAttackRating = previousRatings.getCentralAttack().get(t);
         m_dLeftAttackRating = previousRatings.getLeftAttack().get(t);
-        }
+    }
 
 
     private void addListeners() {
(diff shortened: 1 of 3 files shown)


akasolace:
@tychobrailleur is this only a bug in darklaf ? I am not sure but I thought I experienced the same behaviour in Nimbus ...

tychobrailleur:
I could not reproduce on Nimbus, it only seemed to happen on darklaf themes for me.

akasolace:
> I could not reproduce on Nimbus, it only seemed to happen on darklaf themes for me.

ok

@weisJ  I cc you to keep you informed of that issue

tychobrailleur:
Did it fix the issue for you, @akasolace ?

akasolace:
@tychobrailleur tbh, I did not try it yet. I will do that as soon as I rebase my branch (today or tomorrow) ....

akasolace:
@tychobrailleur so I tested it yesterday and it has been fixed in one place but not in the other. 
I will share animated gif this evening to make it clearer but the situation is that:
   - clock icon now appears correctly from the start
   - club logo: the default icon which is loaded by the problematic function, only appears after interaction (mouse hover)

tychobrailleur:
> @tychobrailleur so I tested it yesterday and it has been fixed in one place but not in the other.

Ah, yes, I can imagine as I didn't look at the other one.  I'll amend the PR.

weisJ:
This problem should be fixed in the darklaf `latest.intergration` version. Because darklaf uses caching and lazy evaluation for derived images (i.e. they are only loaded when needed and share the same original image for optimal quality) sometimes the images could not be available when painting as AWT loads images asynchronously.

akasolace:
@weisJ ok good to know this will be fixed upstream. Meanwhile @tychobrailleur found a ix.

PR #907, opened 2021-02-12
Trainingtable upgrade

wsbrenk:
 
   - fixes issue #__  
   
 - [ ] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -3,15 +3,13 @@ package core.db;
 import core.module.IModule;
 import core.module.ModuleManager;
 import core.module.config.ModuleConfig;
+import core.training.TrainingPerWeek;
 import core.util.HOLogger;
 import module.playeranalysis.PlayerAnalysisModule;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Base64;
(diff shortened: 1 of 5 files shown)


src/main/java/core/db/DBUpdater.java
```diff
@@ -295,19 +298,18 @@ private void updateDBv500(int dbVersion, int version) throws SQLException {
 				trainingTable.tryDeleteColumn("YEAR");
 				trainingTable.tryDeleteColumn("WEEK");
 
+				// Fill training table using information from VEREIN, PLAYER and XTRADATA (Step 2)
+				var sql = "select ACTIVATIONDATE FROM BASICS LIMIT 1";
+				rs = m_clJDBCAdapter.executeQuery(sql);
+				if (rs != null) {
+					rs.next();
+					var activationdate = rs.getTimestamp("ACTIVATIONDATE").toInstant();
+					// TODO all happens on construction. I would prefer a static method for such jobs
```
akasolace, line 307: this will only create a training week manager but then we need to call a function to store the vector inside TrainingTable

akasolace (commented):
I can merge this; but I think we still need the part where we store the vector inside TrainingTable .... correct ?

src/main/java/core/db/DBUpdater.java
wsbrenk: OK - thought that would be done implicitly

Issue #903, opened 2021-02-11
[BUG] trainweekmanager 


akasolace:
private static final Instant m_NextTrainingDate = HOVerwaltung.instance().getModel().getXtraDaten().getTrainingDate().toInstant()

in trainweekmanager will crash for freshinstall

wsbrenk:
FutureTrainingTable is broken. it uses TraininPerWeek, especially deleted methods

wsbrenk:
FutureTrainingTable changed (same structure as TRAINING). 
DBUpdater data migration needs to be done.

Statistics need refactoring of TrainingPerWeek usage.

wsbrenk:
statistic could get the week list with 
select min(hrf_id) from hrf group by  100*YEAR(datum)+WEEK(datum)

wsbrenk:
HTSkillID versus PlayerSkill should be refactored too

PR #920, opened 2021-02-16
#903 training refactor cont

wsbrenk:
 
   - fixes issue #__  
   
 - [ ] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/training/TrainingPreviewPlayers.java
+++ b/src/main/java/core/training/TrainingPreviewPlayers.java
@@ -12,6 +12,7 @@ import core.model.player.Player;
 import module.lineup.LineupPosition;
 import org.jetbrains.annotations.Nullable;
 
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/core/training/TrainingPreviewPlayers.java
```diff
@@ -112,7 +113,7 @@ public int getNextWeekTraining() {
                 nextWeekWeek++;
             }
             var hattrickDate = new HattrickDate(nextWeekSaison, nextWeekWeek);
-            nextWeekTraining = DBManager.instance().getFuturTraining(hattrickDate.toInstant());
+            nextWeekTraining = DBManager.instance().getFuturTraining(Timestamp.from(hattrickDate.toInstant()));
```
akasolace, line 116: ,does the one from training manager not good?

wsbrenk: sorry - but i didn't got it that the training manager would also load future trainings from futureTrainingTable

akasolace: no it only load upcoming training for subskill calculation.

PR #926, opened 2021-02-21
#903 debug training calculation

wsbrenk:
 
   - fixes issue #__  
   
 - [ ] have been updated
 - [ ] do not require update


akasolace:
Hello I will merge this one but I see you replaced a lot of intant by Timestamp.  I would suggest in the future we use instant everywhere. When you need a timestamp for a sql query, you do juust at that moment a call to getSQLTimestampFromInstant()

PR #927, opened 2021-02-22
continuous update of Training table

akasolace:
adds a method updateHistoricalTrainings() in training manager 

   - ensure training table gets regularly updated by introducing new method updateHistoricalTrainings() in training manager.
   - I suggest this function is called at the end of subskill calculation (to be done)

### DIFF
--- a/src/main/java/core/training/TrainingManager.java
+++ b/src/main/java/core/training/TrainingManager.java
@@ -44,14 +44,20 @@ public class TrainingManager {
 					.map(TrainingPerWeek::getTrainingDate)
 					.max(Instant::compareTo).get();
 			recentTrainings = new TrainingWeekManager(previousTrainingDate.plus(1, ChronoUnit.DAYS), false, true);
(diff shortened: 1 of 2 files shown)


wsbrenk:
>I suggest this function is called at the end of subskill calculation (to be done)

this i do not understand. calling update/store training after subskill calculation would be to late, since the calculation needs the training information. i suggest to handle this during download, when training xmls were processed and BEFORE subskill calculation is called, whenever a training had happened since previous download.

akasolace:
@wsbrenk 
in training manager there are 2 vectors;
    - historical trainings to be used for display in training module
    - recent trainings to be used for training calculation (most of the time it would be empty cause no new training reauire calculation, otherwise it should contains 1 element except for users who skipped downloads it might contains multiple entries)
    
once calculation is performed items are moved from recent training to historical trainings

wsbrenk (approved):
in my pr i used recentTrainings in download to store those in db before training calculation happpens. maybe you could change the DBManager call there to a call of this updateHistoricalTrainings.

PR #929, opened 2021-02-22
#903 store training since last download

wsbrenk:
 
  store recent trainings (since last download) in database during download.

calc subskills takes this new downloaded trainings into account.

training panel is NOT refreshed after download - the new training is not displayed
 
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/model/HOModel.java
+++ b/src/main/java/core/model/HOModel.java
@@ -8,6 +8,7 @@ import core.model.misc.Basics;
 import core.model.misc.Economy;
 import core.model.misc.Verein;
 import core.model.player.Player;
+import core.training.TrainingWeekManager;
 import module.youth.YouthPlayer;
 import core.model.series.Liga;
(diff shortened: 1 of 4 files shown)


src/main/java/core/model/HOModel.java
```diff
@@ -624,4 +626,9 @@ public void setYouthTrainings(List<YouthTraining> youthTrainings) {
                 .sorted(Comparator.comparing(YouthTraining::getMatchDate))
                 .collect(Collectors.toList());
     }
+
+    public void storeTrainingsSinceLastDownload() {
+        var recentTrainings = TrainingManager.instance().getRecentTrainings();
+        DBManager.instance().saveTrainings(recentTrainings,false);
```
wsbrenk, line 632: @akasolace i remember you had another approach to store recentTrainings. Feel free to replace this line by your approach. I use storeTrainingSinceLastDownload during download session just before subskill calculation is started.

Issue #960, opened 2021-03-06
[BUG]download does not update training lists


wsbrenk:
consequence: training effect is not calculated after download.

reproduce:

* press f11
* verify if past training and future training lists in training panel are updated correctly
* verify if training progress is displayed on player overview

akasolace:
I made a commit for this last week but then you also made a commit for the same thing so I told you I will let you decide how to deal with it. I noticed the function I created in training manager is never called ... happy to help if you want me to ...

akasolace:
@wsbrenk to clarify, I expect this should work as follow: 
your function calculate subskill should look whether or not trainingmanager.recentTrainings has entries. 
if no -> pass,  if yes -> subskill are updated then at the end of your function you should call updateHistoricalTrainings()

akasolace:
lookimng at the function recalcSubskills()  (https://github.com/akasolace/HO/blob/1fd0aa76ecd05e1801cf2c9f37049e76e53d1b91/src/main/java/core/model/HOVerwaltung.java#L116)

I am confused because it seems you are not even using trainingManager.recentTrainings .....

akasolace:
Applying the above, I get this:

- before download:

![image](https://user-images.githubusercontent.com/1136496/110207630-100d7680-7e85-11eb-9723-430e547d0eaa.png)

![image](https://user-images.githubusercontent.com/1136496/110207642-23b8dd00-7e85-11eb-9508-87d22164c558.png)


We can see that training 05.03 is nowhere .... is it what we want ? 

- after download:


![image](https://user-images.githubusercontent.com/1136496/110209314-118e6d00-7e8c-11eb-8e3a-613734fee5ae.png)

 
![image](https://user-images.githubusercontent.com/1136496/110209319-1e12c580-7e8c-11eb-84aa-cbfa08f1db05.png)


but 12.03 is not in future training

PR #962, opened 2021-03-06
#960 training download

wsbrenk:
 
   - fixes issue #960 (alternative approach)

Future training is updated too

 - [ ] have been updated
 - [x] do not require update


akasolace:
@wsbrenk I suggest we stick with my previous commit. I tested it and I think it works well. Can you split your PR and only push the one regarding Youth

wsbrenk:
@akasolace i'm not sure if your change breaks the recalc feature. recalcSubskills calls model.calcSubskills too and needs the trainings between the downloads saved in the database, not the trainings of the current hoModel.

wsbrenk:
var trainingWeeks = TrainingManager.getTrainings(previousDownload.getDate(), getNextTrainingDate) would do it. 

Proposed signature: getTrainings(fromDate, beforeDate) // beforeDate is used as exclusive date (next training is not included in the result)

akasolace:
@wsbrenk I made a couple of tests again this morning, It seems to me that master is working well. Subskills are properly calculated and past and forward trainings are properly updated.... please let me know if you still see an issue that I missed ...

wsbrenk:
Before Recalc:
beforerecalc

After Recalc (7 weeks)
afterRecalc

No trainingweeks are available for recalc loop as i mentioned above

wsbrenk:
@akasolace i do not intend to use my pr. but you should introduce the method i proposed above. this should be used in calcSubskills instead of getRecentTrainings.

akasolace:
@wsbrenk I don't really understand your request. I mean let say you want to recalculate subskill from `startDate`, you can already do something like: 

`var trainingWeeks = new TrainingWeekManager(startDate, false, true).getTrainingList()`

PR #965, opened 2021-03-07
fix HOVerwaltung.recalcSubskills

wsbrenk:
 
   - fixes recalcSubskills (my idea)
   
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/java/core/model/HOModel.java
+++ b/src/main/java/core/model/HOModel.java
@@ -502,7 +502,7 @@ public class HOModel {
      */
     public final void calcSubskills() {
 
-        var trainingWeeks = TrainingManager.instance().getRecentTrainings();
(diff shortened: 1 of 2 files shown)


akasolace:
@wsbrenk really I am completely missing the point ... the thing is the current implementation ensure consistency between training being treated for subskill calculation and database update. It seems to me that it is working just fine.  I think your PR would break this...  Why do you want to change it? What is the thing not working at the moment according to you ?
To me the only open point regards subskill recalculation that can either be access directly via trainingweekManager as discussed before or implemented in trainingManager if we agree on a defined training range like always 1 season ....

wsbrenk:
@akasolace please check if File->Recalculate (7 weeks) is working correctly with your master version. For me it is not the case.

if you have another idea to implement it, feel free to do it your way, but now it is not working.

akasolace:
@wsbrenk indeed it should not work for the moment as I did not include this but I will do a commit that should do it ...

PR #966, opened 2021-03-07
uttility function in training manager to support subskill recalculate

akasolace:
introduce methods computeHistoricalTrainingsInclMatches() and getTrainingsBetweenDates() in training manager to ensure minimum impact on existing subskill recalculation methid

 if you have initial date = currentDate - 7 weeks
we need to call once `computeHistoricalTrainingsInclMatches(initial_date)`
then we can do bunch of call as follow `getTrainingsBetweenDates(startDate, endDate)`
startdate will need obviously to be >= initial_date
the idea is to initilize the vector only once
that limit call to the database

### DIFF
--- a/src/main/java/core/training/TrainingManager.java
+++ b/src/main/java/core/training/TrainingManager.java
@@ -6,7 +6,6 @@ import core.gui.HOMainFrame;
 import core.model.HOVerwaltung;
 import core.model.player.Player;
 import core.util.HOLogger;
-
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.time.Instant;
@@ -27,7 +26,6 @@ public class TrainingManager implements PropertyChangeListener {
 	private TrainingPerWeek nextWeekTraining;        						// used to determine training bar, include upcoming game   => Created at initilization
     private TrainingWeekManager recentTrainings;     						// trainings that took place (if any null otherwise) since last entry in Training table  => Created at initilization
 	private List<TrainingPerWeek> historicalTrainings;          			// used to populate training history, no match information => Created at initilization
-	private List<TrainingPerWeek> historicalTrainingsInclMatches;           // used for subskill recalculation => Created on request
 
 
 	public static final boolean TRAINING_DEBUG = false;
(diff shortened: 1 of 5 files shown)


src/main/java/core/training/TrainingManager.java
```diff
@@ -23,12 +24,13 @@
 	// singleton class
 	private static TrainingManager m_clInstance;
 
-	private TrainingPerWeek nextWeekTraining;        // used to determine training bar, include upcoming game   => Created at initilization
-    private TrainingWeekManager recentTrainings;     // trainings that took place (if any null otherwise) since last entry in Training table  => Created at initilization
-	private List<TrainingPerWeek> historicalTrainings;         // used to populate training history, no match information => Created at initilization
+	private TrainingPerWeek nextWeekTraining;        						// used to determine training bar, include upcoming game   => Created at initilization
+    private TrainingWeekManager recentTrainings;     						// trainings that took place (if any null otherwise) since last entry in Training table  => Created at initilization
+	private List<TrainingPerWeek> historicalTrainings;          			// used to populate training history, no match information => Created at initilization
+	private List<TrainingPerWeek> historicalTrainingsInclMatches;           // used for subskill recalculation => Created on request
```
wsbrenk, line 30: i do not see any difference between both lists. Both are including match info and they are identical if computeHistoricalTrainingsInclMatches uses the same start date as the TrainingManager.instance(). so i propose:

* use historicalTrainings to filter in getTrainingsBetween
* remove the loading of matches from TrainingPerWeek constructor and make getMatches loading data on request again

getMatches(){
if (matches==null) loadmatches();
return matches;
}

wsbrenk (changes requested):

src/main/java/core/training/TrainingManager.java
akasolace: > i do not see any difference between both lists. Both are including match info

no, first list is simply the content of training table loaded into a List there are no match info .... it is very quick to load. 
the second one is made parsing HRFs, and loading all matchkurzinfo (more heavy)

we only want to load match for subskill calculation, otherwise we don't even need a training table at all  and we redo the work of parsing hrf with match info at every application start (but I believe this would be terrible on performance).

akasolace:
@wsbrenk  I made the changes I think you requested...

src/main/java/core/training/TrainingManager.java
wsbrenk: loading the training table content invokes the constructor of TrainingPerWeek which loads the matches - isn't it?

akasolace: no it does not ...

akasolace: Let's try with this PR, and see how it goes, ok for you ?

PR #967, opened 2021-03-08
youth injured, suspended

wsbrenk:
  
   - continues issue #367
   
 - [ ] have been updated
 - [x] do not require update

### DIFF
--- a/src/main/resources/sprache/English.properties
+++ b/src/main/resources/sprache/English.properties
@@ -2115,4 +2115,3 @@ ls.youth.player.training.matchtype=Match Type
 ls.youth.player.training.rating=Rating
 ls.youth.player.lastmatchdate=Last Match Date
 ls.youth.player.isSuspended=Suspended
-ls.youth.player.InjuredWeeks=Injured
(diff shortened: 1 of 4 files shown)


src/main/resources/sprache/English.properties
```diff
@@ -2110,4 +2113,6 @@ ls.youth.player.matchcount=Match Count
 ls.youth.player.trainingsum=Trained skill sum
 ls.youth.player.training.matchtype=Match Type
 ls.youth.player.training.rating=Rating
-ls.youth.player.lastmatchdate=Last Match Date
\ No newline at end of file
+ls.youth.player.lastmatchdate=Last Match Date
+ls.youth.player.isSuspended=Suspended
+ls.youth.player.InjuredWeeks=Injured
```
akasolace, line 2118: could you not use existing translated terms e.g. `ls.player.injurystatus.injured`.  ?  If we can avoid having doublons in translation file it is best especially because our translators are lagging behind.... what do you think ?

wsbrenk: of course - thanks for your hint

akasolace: similarly for suspended you can use:   ls.player.warningstatus.suspended

I haven't check others .....

Issue #987, opened 2021-03-23
[BUG] "Past trainings" imported from .hrf tend to be shifted by one week


dominiquelee:
On a fresh update, the past trainings are filled from existing .hrf but are wrong by one week
![image](https://user-images.githubusercontent.com/1204726/112152235-5f240d00-8be2-11eb-99f5-acb18ced050e.png)
Here, I did a defending training on the 1/1 but it was reported as done on the 1/8.

I believe that's because for someone that only downloads one .hrf by week (on the training day), the only .hrf "downloaded on the same week as the training" is the one that was downloaded 7 days before the training, thus, with the previous training.

**To Reproduce**
Update from 4.1 to 5.0 dev
Training / Past Trainings
Check for last training change
Compare with your lineups that week, it most likely doesn't match
After a subskill recalculation, it's also visible through Statistics/Players for trained players as the curves won't be smooth at all.

**Platform information:**
 - OS: Windows
 - Version 5.0 dev, r3557

wsbrenk:
I'm afraid the software has no real chance to decide for sure which training is the right one if before the training date training A was set and then training B. It could be both A correct, as well as B or even C, depending on what happened in the meantime in hattrick. At the moment, the algorithm chooses A. If this is incorrect, the user must correct it manually. A more complex investigation due to player developments is difficult to perform in the context.

dominiquelee:
It's clear that there's no perfect solution, but I'd argue that choosing the training of the following week might be correct more often for most people. I believe most people do their weekly sync immediately after the training rather than immediately before the training (though obviously, that's a big baseless hypothesis).

wsbrenk:
@dominiquelee i agree, and i think that could be changed easily. Let us wait on @akasolace who should decide which way we should go.

wsbrenk:
changed method is not only called on db upgrading, but also each time when HO downloads hattrick's data.

current solution gives errors on sequence:   Change training type - Training Date - HO download => download selects previously selected training type

proposed solution gives errors on sequence Training Date - Change training type - HO download => download selects future training type

which one is less likely?

PR #988, opened 2021-03-27
#987 TODO

wsbrenk:
idea how we could solve #987


akasolace:
hello @wsbrenk haven't thought about it yet. Is it really something we should do for the next version ? 
I am frankly under water at the moment at work, no time to work on HO! lately. Hope things will improve shortly ....

wsbrenk:
@akasolace there is not a second chance to do it. if the upgrade to V5.0 is done the trainings per week are fixed and need manual change if they are not correct. i could do the change and you decide if we take the current or the proposed solution. There are some days free this weekend;-)

Issue #1053, opened 2021-05-28
[BUG] Error on download


akasolace:
![image](https://user-images.githubusercontent.com/1136496/120025544-1c6f1e80-bff1-11eb-94e4-76c0d11c4c27.png)

![image](https://user-images.githubusercontent.com/1136496/120026426-5b51a400-bff2-11eb-90a1-ba16a81cde92.png)

akasolace:
same bug as reported by @poteus  #1038  (no idea how I ended up with this wrong entry during data download ....

akasolace:
@wsbrenk I can't ind any reference to that game. I never faced that opponent.
The match report mentions some of my newly acquired players. It seems to me this is  a bug on CHPP site ....

akasolace:
I reported the bug on HT forums

akasolace:
@wsbrenk also I noticed that downloading new data cause download of lineups for all game in matkurzinfo table,  should ze not li,it this to most recent game or even do lazy doznload of lineup ?

wsbrenk:
![image](https://user-images.githubusercontent.com/11821502/120068632-ec239080-c081-11eb-9134-92279cd97463.png)

this tournament matches are no longer given by match archive. this gives the errors.

wsbrenk:
@akasolace i haven't played tournament matches for quite a long time. Could you please check if you find out how long those matches are kept in ht match archive? 

my first guess is, that i don't reload those matches if they are older than one year.

wsbrenk:
> also I noticed that downloading new data cause download of lineups for all game in matkurzinfo table, should ze not li,it this to most recent game or even do lazy doznload of lineup ?

are you sure, if really ALL matches details/highlights were reloaded or not only a maximum of 10 each time you start your HO?

akasolace:
Just to understand the issue ... you are saying that tournament games are simply dropped from ht ? In that case I still don't understand why March details shows wrong match date ....

wsbrenk:
my assumption is probably wrong. I have just found tournament games from 2016 in the archive. Maybe matches were stored as htoIntegrated matches which weren't htointegrated, wrong match type or anything like this.

most of the changes of my pr should help to store with correct sourceSystem. the isObsolet method should be modified - if we know what kind of wrong matches were in the database.

akasolace:
the fake entry in my db, says the game was played the 2021-05-23 17:26:52 and I got a match report

akasolace:
it really seems a bug from CHPP

[...]

PR #1054, opened 2021-05-29
#1053 MatchKurzInfo.isObsolet

wsbrenk:
 
   - fixes issue #1053
   
 - [ ] have been updated
 - [x] do not require update

3. [Optional] suggested person to review this PR @akasolace - could you please check, if my implementation of MatchKurzInfo.isObsolet is OK (if hto match is older than one year?)

### DIFF
--- a/src/main/java/core/net/OnlineWorker.java
+++ b/src/main/java/core/net/OnlineWorker.java
@@ -260,8 +260,7 @@ public class OnlineWorker {
 		return downloadMatchData(info, refresh);
 	}
 
-	public static boolean downloadMatchData(MatchKurzInfo info, boolean refresh)
-	{
+	public static boolean downloadMatchData(MatchKurzInfo info, boolean refresh) {
 		int matchID = info.getMatchID();
 		if (matchID < 0) {
 			return false;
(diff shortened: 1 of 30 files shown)


akasolace:
@wsbrenk I will wait to have more information about HT behaviour before merging this PR, I am not sure the 365 days rule is correct, It seem to me it is from season to another and only affect division battle ... to be investiguated ...

src/main/java/core/net/OnlineWorker.java
```diff
@@ -270,23 +269,23 @@ public static boolean downloadMatchData(MatchKurzInfo info, boolean refresh)
 		showWaitInformation(1);
 		// Only download if not present in the database, or if refresh is true
 		if (refresh || !DBManager.instance().isMatchInDB(matchID)
-				|| DBManager.instance().hasUnsureWeatherForecast(matchID)
-				|| !DBManager.instance().isMatchLineupInDB(SourceSystem.HATTRICK.getValue(), matchID)
+				|| (info.getMatchStatus() != MatchKurzInfo.FINISHED && DBManager.instance().hasUnsureWeatherForecast(matchID))
+				|| !DBManager.instance().isMatchLineupInDB(info.getMatchType().getSourceSystem().getValue(), matchID)
```
wsbrenk, line 273: i think here could be the error: htointegrated matches were checked with wrong SourceSystem (HATTRICK), which led to an additional download of maybe wrong matches.

akasolace: as you can see in my various screenshot above the source system was correctly set to hto integrated

PR #1069, opened 2021-06-04
youth matchType refactoring

wsbrenk:
 
youth module is working again.
 
 - [ ] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -266,7 +266,8 @@ final class DBUpdater {
 			futureTrainingTable.tryDeleteColumn("WEEK");
 		}
 
-		if (columnExistsInTable("SourceSystem", MatchLineupTable.TABLENAME)) {
+		// SourceSystem does NOT exist in HO4, but HeimName does
+		if (columnExistsInTable("HeimName", MatchLineupTable.TABLENAME)) {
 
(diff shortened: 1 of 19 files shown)


src/main/java/core/db/DBUpdater.java
```diff
@@ -266,7 +266,8 @@ private void updateDBv500(int dbVersion) throws SQLException {
 			futureTrainingTable.tryDeleteColumn("WEEK");
 		}
 
-		if (columnExistsInTable("SourceSystem", MatchLineupTable.TABLENAME)) {
+		// SourceSystem does NOT exist in HO4, but HeimName does
+		if (columnExistsInTable("HeimName", MatchLineupTable.TABLENAME)) {
```
akasolace, line 270: @wsbrenk it means it will try the upgrade everytime we launch HO! isnt it?  (looking the PR from my iphone ...) 
We might need something like if matchtype does not exist ......... to make sure this is applied only once, no ?

wsbrenk: no  - since the column is deleted some lines lower

wsbrenk: i think it not a good idea to add the columns SourceSystem and then remove them again in the same method, if a user is starting upgrade from HO4.

Issue #1080, opened 2021-06-08
[BUG] training wrong value displayed for new player


akasolace:
I just bought a player, 

As expected in HT, I have no information about skills development

![image](https://user-images.githubusercontent.com/1136496/121245175-ffafd200-c89f-11eb-8ddf-5033d8a6fd75.png)

In HO!  it shows:

![image](https://user-images.githubusercontent.com/1136496/121245238-15bd9280-c8a0-11eb-928b-8c75fe8b8dbe.png)

whereas in reality it has PM = 3 and Scoring = 16  (values correctly displayed in Lineup tab)

wsbrenk:
@akasolace do you mean the values given in training development of ...?
these values are expected skill changes in the future.

akasolace:
I thought training development was past training (I think display can really be clarified see #1079 ).  
Still,  I bought the player with skill 16 in scoring and HO! tells me that in 2 days his skill will drop to 16 (which is also incorrect).

wsbrenk:
>  (I think display can really be clarified see #1079 ).
Still, I bought the player with skill 16 in scoring and HO! tells me that in 2 days his skill will drop to 16 (which is also incorrect).

i confirm clarification of the display. currently it is done by different colors. black letters for past skill ups, colored letter for future skill ups. Bad tagging of selected player is a problem of the dark mode.

Could you debug, why your new player shows a skill down from 17 to 16? I can't see, when your download was done and if it could be, that the player had skill 17 then!?

akasolace:
no I purchased my player with skill 16, also HO! forecast a loss o skill from 17 to 16 to happen in 2 days ... (not in the past)

akasolace:
I just updated HO! and still the same issue  (squad tab is good but Training tab predict a fake skill drop )



![image](https://user-images.githubusercontent.com/1136496/121319818-ded19600-c90c-11eb-8b26-7071ad82bc0a.png)

![image](https://user-images.githubusercontent.com/1136496/121319993-09bbea00-c90d-11eb-95cb-681d72c708f4.png)


![image](https://user-images.githubusercontent.com/1136496/121319928-f741b080-c90c-11eb-95d5-09e9972817b1.png)

akasolace:
actually I just noticed plenty of players have that bug ...

wsbrenk:
@akasolace I suspect that we mistakenly do not specify the target skill, but the initial skill on skill drops.

[...]

PR #1081, opened 2021-06-09
#1080 fix FutureTrainingManager.checkSkillChange rounding issue of sk…

wsbrenk:
…ill drops

   - fixes issue #1080
   
 - [ ] have been updated
 - [ ] do not require update

### DIFF
--- a/src/main/java/core/training/FutureTrainingManager.java
+++ b/src/main/java/core/training/FutureTrainingManager.java
@@ -255,18 +255,11 @@ public class FutureTrainingManager {
 	*/
 	private int checkSkillChange(int pos) {
 		if (finalSub[pos] >= 1) {
-//			Alternative 1: Set sub=0 after a skillup 
-//			(We will use this, until the training speed formula is optimized)
-//			finalSub[pos] = 0;
-
-//			TODO flattermann
-//			Alternative 2: Use overflow sub after a skillup
-//			(This would be more accurate. But only if the underlaying formula is exact) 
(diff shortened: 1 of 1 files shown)


akasolace:
instead of this what about my more general proposal, truncating values instead of rounding ? 
can't it help in other places in HO! ?

wsbrenk:
@akasolace i didn't fount the place where i could replace rounding by truncating.

if truncation, then we need to find out whether the user is displaying one or two precision numbers and truncate accordingly.

akasolace:
@wsbrenk If you agree this would be a more genral approach, I would close that PR and make a more general ticket...

wsbrenk:
@akasolace i agree. display is suboptimal.

Concerning calculation i don't know if the current (alternative) formula tooks skills drops into account. if so - we overestimate the skill drops of trained skills. do you know who was responsible for that formula?

akasolace:
> Concerning calculation i don't know if the current (alternative) formula tooks skills drops into account. if so - we overestimate the skill drops of trained skills. do you know who was responsible for that formula?

know sure I understand the question as I am not familiar with that part of the code. For me skill drops and training are separated formula as they do not occur at the same time, no ?

wsbrenk:
> For me skill drops and training are separated formula as they do not occur at the same time, no ?

concerning hattrick you're right - but in HO the training calculation is done in one run at training date as far as i know. don't know if we could get the skill down date from chpp and whether  it is worth doing so.

PR #1136, opened 2021-08-28
Update build.gradle

akasolace:
@wsbrenk  I guess this should fix the pull of translation from PoEditor during build. 

in case it fails, I will let @tychobrailleur advise on how to do it...


wsbrenk:
@akasolace what is probably still missing is the definition of the POEDITOR_APIKEY in the release-dev.yml. similar to and beneath           INSTALL4J_LICENSE: ${{ secrets.INSTALL4J_LICENSE }}
in line 26.

Another question about the other direction, english file sync to poeditor: this is already in place with poeditors github integration or do we need to push changes of the english language file to poeditor (if we have new keys)?

Issue #1350, opened 2022-01-20
[BUG] Not able to download / Upload Lineup in HO 5.1 Version


Achill-3s:
I have updated to the new stable version 5.1 for HO.
But when i go to lineup, i can no longer upload or download lineups for future matches.

When i scroll over it, there is a popup message saying:
"Please start by downloading new data from Hattrick. Meanwhile only lineup simulator is available"

Even afteri download data, I cannot see the upload/downlod lineup options or even select future match.
Kindly help out how to fix this issue please.

Yours Sincerely,
Achilles
![image](https://user-images.githubusercontent.com/98124706/150416807-3628ba99-34e9-44e9-a988-806127a03880.png)

wsbrenk:
i will try to reproduce this

Achill-3s:
If this helps the HO update went from Version 4.1.3310.2 to Version 5.1.3877.2 when the issue above occurred.
Any questions just let me know to help regarding reproducing the bug.

Achill-3s:
FYI both portable or installer versions of 5.1 and 6.0 have the same issue.
I needed to revert back to Portable 4.1 to get the feature working again for setting match orders.

wsbrenk:
could you upload a backup of your database? latest  zip file in %APPDATA%/HO/db

Achill-3s:
Find attached contents of: C:\Users\Achilles_\AppData\Roaming\HO
Three Teams, all work with 4.1,, but not in 5.1 version LineUp

Attachment At:
https://ufile.io/ws0p6u77

wsbrenk:
my system says, that it is not save to download this. could you please attach it to this issue - simple drag it to these notes.

Achill-3s:
File is too big to attach as a reply - 63MB

Achill-3s:
I uploaded the Roaming Folder as it is at a shared dropbox.
Let me know if the link doesnt work for any reason.

https://www.dropbox.com/sh/4o3sa56if7108p2/AACHU7aAZloqdNjE7WsChn1ea?dl=0

wsbrenk:
Download successful

wsbrenk:
there is no authorization in those databases.  i can not invoke a download to reproduce the error.

Achill-3s:
I have re-added the youth permission which was missing since i reverted to 4.1 for using HO.
Keep in mind that it maybe easier to use the portable 5.1 with the attached db below for re-producing the error.
[...]

PR #1357, opened 2022-01-26
#1350

wsbrenk:
 
   - fixes issue #1350
   
 - [x] have been updated
 - [ ] do not require update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,8 +41,8 @@ test {
 //  Version ============================================================
 // Development_stage (DEV:0  BETA:1  STABLE:2)
 def major = '5'
-def minor = '1'
-def development_stage = 2
+def minor = '2'
+def development_stage = 1
 def releaseArtefacts = true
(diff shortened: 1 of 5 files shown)


wsbrenk:
@akasolace All the time classes confuse me a bit - so i only prolonged the dataTooOld period (now 2 days instead of 1 hour).

if you would try to reproduce the error, set your time zone to cairo then try download and the download is instantly too old.

akasolace:
@wsbrenk that is strange, when I am done with my player analysis commit, I will investigate this error ....

wsbrenk:
@akasolace i will merge this to the release HO5 branch. it is a workaround to fix the download problems in different time zones. It should be a HO5.2-beta. After approx. 2 weeks of no other bug reports we should produce a new stable 5.2.

in HO6 the download is no longer disabled by time intervals. But i fear there can be other bugs concerning errors in time zone handling.

Issue #1267, opened 2021-11-07
[FEATURE] Lineup-Simulation workflow


wsbrenk:
Feedback from HT forum:

Match page with always all options available and working instead of being greyed out would be nice

- [x] increase hight of right panel to screen hight
- [x] layout of rating panel
- [x] rating panel change layout of hatstat, loddarstat, tactic and formation
- [x] new copy button in rating panel
- [x] move rating prediction type selection from rating panel to rating settings panel (below rating panel)
- [x] move match selection/banner + up-/download and order settings to new panel above rating panel
- [x] assistant panel and Database Load/Store panels in TabView below rating settings/lineup simulation

wsbrenk:
Another user; "The negative feedback for me is the lineup tab huge and in my opinion with a wrong layout."

peppermil:
It's the same for me, I liked the old lineup tab (HO4) better - more compact layout.

wsbrenk:
![image](https://user-images.githubusercontent.com/11821502/142763713-c013009d-e516-4659-9af5-be53485ff1a6.png)

peppermil:
yes that would be great ;)

wsbrenk:
progress report: 
![image](https://user-images.githubusercontent.com/11821502/143770949-c71c9ce4-cf2f-45ca-b5c0-245547f7bd4d.png)

wsbrenk:
colored dots instead of borders:
![image](https://user-images.githubusercontent.com/11821502/143788524-9aa5eb2a-c461-4f06-b76a-4e672c7f2251.png)

Cotisso:
Could you please add back the setting where we can save the line ups?
In my opinion it is very important.
Thx, have a great day!

wsbrenk:
Load/Store panels should do that job - i'm working on it.

wsbrenk:
the following database tables have been removed

* AUFSTELLUNG (info will be added to MATCHLINEUPTEAM table)
* POSITIONEN (info will be added to MATCHLINEUPPLAYER table)
* MATCHORDER (info will be added to MATCHLINEUPPLAYER table)

wsbrenk:
MATCHLINEUPTEAM table should store user defined lineups too (template lineups, not assigned to an single match)

wsbrenk:
![image](https://user-images.githubusercontent.com/11821502/146759536-0aea147c-ead8-4090-a372-fa0f2199152a.png)

move match and lineup selection not above but below rating and assistant panels

PR #1374, opened 2022-02-04
store lineup of uploaded match orders

wsbrenk:
 
   - fixes issue #__  
   
 - [ ] have been updated
 - [x] do not require update


akasolace:
Tbh, I don't really recall when these table is being used.
Is this fixing a specific open bug ? For my info  Is it to keep lineup information between the time you send information and before you download again data? Is there any other use case ?
Otherwise your code seems good to me. I can't imagine any side effect.

wsbrenk:
when uploading the lineup, it is not saved. i.e. when the HO is restarted, the previously loaded lineup is visible again and not the uploaded one. you have to download again to see the correct lineup.

i think this a residual point of #1267  - so i thought that no other ticket is necessary.

Issue #1373, opened 2022-02-04
[BUG] League statistics is broken


akasolace:
League Statistics is showing crazy value whereas Hattid data seems just fine .... 

![image](https://user-images.githubusercontent.com/1136496/152578575-e9f33eb8-e09c-4b96-b041-51cea1d15c89.png)

wsbrenk:
@akasolace : download of all matches are required. if you want to reactivate the hattid connection, i would like to have this configurable. i prefer downloading all matches and calculate the values locally.

wsbrenk:
to fix this, we alternativly could download all league matches by default - ne need to do it by hand.

akasolace:
For me Hattid was easier ... I am fine with your approach but at the moment it is broken not all matches are downloaded in HO.
So either we force download of all matches or we get data from Hattid  whatever you prefer. For the latter I am happy to reintroduce but for the former I would leave it up to you as you are more familiar with the approach.

akasolace:
What about performance ?  I would suggest we first fetch from Hattid … if
error we force download all matches from HT , what do you think ?

wsbrenk:
#995 and #1070 were the reason for changing this. i would like if connection to Hattid could be suppressed by configuration.
i used to download all matches by hand, so i don't have any problem with missing data.
But i'm interested in the approach to download all matches of the league automatically.

wsbrenk:
i'm working on the download of all matches of the latests match days

akasolace:
so all matches played in the league for current season, right ?

wsbrenk:
the plan is that all finished league matches since last download will be loaded. Older matches which were not loaded before this change have to be loaded by hand. 

In future this should only happen in case of download errors (connection loss or something like this)

Issue #1359, opened 2022-01-29
[BUG] Errors in MatchLineupTeamTable and HTDateTime


wsbrenk:
[...]

PR #1376, opened 2022-02-05
#1373

wsbrenk:
 
   - fixes issue #1373, #1359 (only MatchLineupTeamTable)
   
 - [x] have been updated
 - [ ] do not require update


akasolace:
@wsbrenk 
will it now download all match of the serie for the last day or for all current season ? 
To fix the issue, I think it should be the former, right ?

Issue #383, opened 2020-02-07
[eFEAT] auto adjustment of stamina subskill 


akasolace:
we could offer to adjust stamina subskill of player based on game results.
if player had a game in the week, similar to what Foxtrick is doing, we deduct stamina decimal from RatingEndOfGame  vs Rating, and we adjust stamina decimal accordingly.

akasolace:
@sophiastr for example this one ....

sophiastr:
Indeed, this is easier. Thank you very much.
Can you assign me this issue?

sophiastr:
Good Evening,
Sorry, I have no solution to the problem yet. I estimate that within 1 or 2 weeks I will have solved it. Sorry again. However, I have some questions. I do not understand how stamina is being calculated so far.
Can you tell me in which class the player's stamina is calculated?
I found this class related to the player's stamina.
![Στιγμιότυπο οθόνης (533)](https://user-images.githubusercontent.com/72817947/163895441-e7237935-9fc8-4d34-9ec2-177dd8b0892d.png)

wsbrenk:
@sophiastr I don't think there is already a calculation of the stamina subskill in HO

other skills are calculated in HOModel.calcSubskills

sophiastr:
Hello again, I am working on the method for calculating stamina.
Now, I understand better the project and the various variables.
Can you give me more information on how to adjust stamina? 
What exactly do you want me to do?
Also, if you want my opinion it would be useful to put more parameters in stamina calculation.
For example, the minutes a player has played, training, and injuries affect the stamina.
Do you want me to go ahead with my idea or yours? 
Finally, because I work in the class "Player", I have noticed some errors in the comments. Wrong names, many German words that confuse me, and other mistakes.
Would you mind correcting these?

Thank you,
Sophia

wsbrenk:
@sophiastr i do not really know how @akasolace planned to calculate the stamina subskill. what i find concerning stamina numbers is this: https://www88.hattrick.org/Forum/Read.aspx?t=17445814&n=6&v=0 

maybe we could conclude from player's rating at end of match to his stamina?!

sophiastr:
Okay, then I will start calculating the stamina by taking into account only the RatingEndOfGame.
I'm waiting for some feedback @akasolace.
Tell me if you want to make the calculation of stamina more complicated and realistic by adding other variables.

Thank you,
Sophia

sophiastr:
[...]

PR #1491, opened 2022-05-06
Adjust stamina

sophiastr:
 
   - fixes issue #383  
   
 - [ ] have been updated
 - [ ] do not require update

4. About the issue:

If a player has played a game during the week, stamina is adjusted according to the type.
previous stamina - ((3/m_lastMatchRating))^0.2 .
The type does not allow major changes in case of a bad game rating.
Also, I am writing a test to check my code.


sophiastr:
Hello @wsbrenk, 
I made a pull request. I'm waiting for your feedback.

akasolace:
Hello @sophiastr,
What I had in mind was not quite that but the following:
 
when a player plays a game you have from match report:
  - time spent on the field 
  - average rating
  - rating at end of game

From the difference between the 2 ratings it is possible to guess the stamina sub levels. Hence we could update them for players after each game...

The issue description might not have been clear. 

What do you think ? Also looking forward to have @wsbrenk inputs ...

sophiastr:
So, you mean that  the new stamina will be the difference: stamina - (average rating - rating at end of game) ???
Moreover, so far I was working on the Player class, now in which class should I work in order to automatically renew stamina after each game?
Sorry, it was not what you wanted.
I will have the new code soon.

wsbrenk (changes requested):
We are very grateful for the support. but I'm afraid we still have some things to adjust.

sophiastr:
Can you give me some more information? For example, in which class are the variables you want to use? In this class?
image

Also, is this the equation?
stamina - (average rating - rating at end of game)

Thank you, @akasolace @wsbrenk

wsbrenk:
@sophiastr i think Player.calcSubskills is a good place to handle the stamina subskill too.
i would introduce a new variable Player.subStamina to store the guessed stamina subskill.

sophiastr:
Hello @wsbrenk,

I made some changes.
I am waiting for your feedback.
I, also, fixed the format in Player class.
If the code is what you wanted, tell me to start the tests.

Thank you

sophiastr:
Good evening again, whenever you have time give me some feedback. I really want to be able to contribute to this project. Tell me if the implementation is ok or if I need to change something.

wsbrenk:
@sophiastr please excuse me for not coming to the review at the moment - have to harvest honey at the moment and therefore come to nothing else.

src/main/java/core/model/player/Player.java
```diff
@@ -1712,7 +1757,7 @@ public int getValue4Skill(int skill) {
             case WINGER -> m_iFluegelspiel;
             case SCORING -> m_iTorschuss;
             case SET_PIECES -> m_iStandards;
-            case STAMINA -> m_iKondition;
+            case STAMINA -> getStamina();
```
wsbrenk, line 1760: in Player.calcsubskill you mention that stamina is the subskill of stamina/kondition (german word for stamina). but here you return it as major value of stamina. major value of stamina is downloaded from hattrick and does not need to be calculated.

src/main/java/core/model/player/Player.java
```diff
@@ -2194,54 +2223,69 @@ public void setFutureTraining(FuturePlayerTraining.Priority prio, HODateTime fro
...
+                setStamina((int) m_isubStamina);
+            } else {
+                if (m_isubStamina < 1) {
+                    setStamina(1);
+                }
+                if (m_isubStamina > 10) {
+                    setStamina(10);
+                }
+            }
+        }
```
wsbrenk, line 2283: i do not understand this calculation - please explain. i thought the result of substamina should be a double value in the range from 0,00 up to 0.99.

wsbrenk (changes requested):

sophiastr:
Thank you for your feedback @wsbrenk. 
You are completely right. 
I hadn't understood so well the difference between stamina and substamina.
One last question.

Why substamina should be between 0 and 0.99? Since substamina comes from the difference between the 2 ratings (LastMatchRating average rating), it may be greater than 1. Is this true?

sophiastr:
I made some changes. When you have time, look at them and give me some feedback.
I'm not sure I understand exactly how you want stamina subskill to be calculated but I made an attempt.
If this is not what you want, give me some information and I will change it immediately.
Thank you @wsbrenk ,
Sophia

sophiastr:
Hello again @wsbrenk ,
Because I have the time and desire to write code, is there any work to be done?
For example, I can deal with the format or Javadoc comments. If there is no job, can I continue with a new issue?
Thank you,
Sophia

masterpatje:
@sophiastr if you want you can look at issue 1489 how and verify how subskills are calculated. Maybe you can find out if my problem is related to substitutes not being calculated correctly. You might look at the code and see how the calculation of subskills is done.

sophiastr:
Of course, I will deal with the issue that you suggested to me @masterpatje .
I need a couple of days to do a test for the code I delivered.
It is very important for me to be able to merge within a week. So as soon as you find time, give me some feedback. Thank you and sorry @wsbrenk .

sophiastr:
Hello @wsbrenk ,

I am really sorry to bother you, but please check my pull request when you find the time.

Thank you and sorry.
Sophia

sophiastr (commented):
Calculate Stamina Subskill

wsbrenk (changes requested):

src/main/java/core/model/player/Player.java
```diff
@@ -2261,19 +2261,10 @@ public void calcSubskills(int previousID, List<TrainingPerWeek> trainingWeeks) {
...
-                    setStamina(1);
-                }
-                if (m_isubStamina > 10) {
-                    setStamina(10);
-                }
-            }
+            double m_iavgRating = DBManager.instance().getBewertungen4Player(getPlayerID())[2];
+            double subStamina = m_lastMatchRating/m_iavgRating;
+            subStamina = subStamina/10;
+            setSubStamina(subStamina);
```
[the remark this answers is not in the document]
sophiastr: Thank you for your feedback. While I was trying to solve the issue, I used many variables, above them the last game rating, and the average rating of the last game. But all the getters in PlayerInfo return null. Have you noticed this problem? That's why I used getBewertungen4Player. Do you want to understand first why it returns null and then use those variables in the formula?

sophiastr:
Good evening @wsbrenk , do you know if LastMatch_RatingEndOfGame information is stored somewhere? Is this information somewhere? Because in all the ways I tried to get it, I did not succeed.
I tried in so many ways.
They return null or 0.
Can you tell me if you have this problem too?
If you do not have the same problem can you send me the command to have access?

Sophia

wsbrenk:
@sophiastr 

in constructor of class PlayerInfo i find this:
~~~
        this.rating = Float.parseFloat(i.get("LastMatch_Rating"));
        this.lastMatchId = Integer.parseInt(i.get("LastMatch_id"));
        this.lastMatchPosition = Integer.parseInt(i.get("LastMatch_PositionCode"));
        this.lastMatchPlayedMinutes = Integer.parseInt(i.get("LastMatch_PlayedMinutes"));
        this.lastMatchRatingEndOfGame = Float.parseFloat(i.get("LastMatch_RatingEndOfGame"));
~~~

i willl now try to find out if we could use it and where it comes from

wsbrenk:
i do not find any playe where this information is stored in the database. it looks like it is only temporary available and not used in any place for now.
class PlayerInfo is used in the team analyzer module. there is a lot to do, to get this running (database update and so on). 

i have moved this to the milestone 7. tell me if you want to analyze the things that needs to be done or if you want me to do it?
i think we have to find a solution how to get this info in the database table SPIELER (player).

sophiastr:
Hello @wsbrenk ,
Thank you very much for your help. Now that I have the information, I will try to fix the method again. I am thinking of doing this:
ratingEndOfGame - averageRatingOfLastGame
In this way, if the number is positive it means high stamina otherwise if it is negative the number means low stamina.
Is this approach correct and what do you want me to do?

Sophia

sophiastr:
Hello @wsbrenk ,
Thank you very much for your help. Now that I have the information, I will try to fix the method again. I am thinking of doing this:
ratingEndOfGame - averageRatingOfLastGame
In this way, if the number is positive it means high stamina otherwise if it is negative the number means low stamina.
If the difference is big and positive then it means very high stamina.
This is how I would approach it.
Is this approach correct and what do you want me to do?

Sophia

sophiastr:
Thanks a lot for the help @wsbrenk . Can I continue in Player class ? I will send you how I intend to do my method in a little while.

sophiastr:
Hello again @wsbrenk 
I still do not have access to the information needed.
I used the method you created in Player class but it returns null.
I also used this command: DBManager.instance().getTALatestPlayerInfo(getPlayerID()).getLastMatchRatingEndOfGame()
All the methods in PlayerInfo return Null or 0.
Moreover, I need another variable to calculate stamina subskill. I need averageRatingOfLastMatch.
This variable exists in PlayerInfo and Player class but the getters in each class return zero.

Finally, I thought of a new idea to calculate stamina subskill.
Here is my proposal:

If LastMatch_RatingEndOfGame > averageRatingOfLastMatch means an increase in stamina.
As a result i calculate stamina subskill based on this formula: (LastMatch_RatingEndOfGame -averageRatingOfLastMatch) / 10
great difference means high stamina
10-2= 6/10(great increase)
If LastMatch_RatingEndOfGame < averageRatingOfLastMatch means a decrease in stamina.
As a result, i calculate stamina subskill based on this formula: (averageRatingOfLastMatch - LastMatch_RatingEndOfGame) / 10
great difference means low stamina
 6-2=4/10(big decrease)
Finally, if there is an increase, I add the stamina subskill
If there is a decrease, I deduct the stamina subskill

It makes sense, though if you want something more complicated then tell me to try.
Tell me your opinion.
If you want me to proceed with your idea, please give me information or a formula because I do not understand how a simple subtraction can cover all the scenarios. 


Sophia

sophiastr:
Hello @wsbrenk ,

When you have time, see why the method you made returns Null so I can move on.
Thank you and sorry
Sophia

wsbrenk:
@sophiastr the values of LastMatch_RatingEndOfGame are only available in new downloaded matches. HO will not download "old" matches automatically. please check if the new methods returns null even in new downloaded matches - this should not happen!

concerning your ideas: 
* RatingEndOfGame is never greater than AverageRating => your formulas will only give "decrease" of stamina.

wsbrenk:
@sophiastr before further investigating on that stamina formula (i don't have any idea, how it should work) you could start making the three new lastmatch values available in the squad table view. this should be a little bit easier.

masterpatje:
> @sophiastr the values of LastMatch_RatingEndOfGame are only available in new downloaded matches. HO will not download "old" matches automatically. please check if the new methods returns null even in new downloaded matches - this should not happen!
> 
> 
> 
> concerning your ideas: 
> 
> * RatingEndOfGame is never greater than AverageRating => your formulas will only give "decrease" of stamina.
> 
> 

Actually it is possible in very rare occassions that the RatingEndOfGame is higher than average rating. I don't know if this still happens and what conditions triggers it. Maybe changed weather condition during the match in combination with a specialty that was penalised by certain weather conditions that are no longer present? If I remember correctly hattrick shows this by a different colour in the stars... might be red that indicated the gain in ratings and the red stars represent the increase.

Suggestion: add line of code that in these rare events the rating end of game is adjusted to / equal to the average rating.

Best of luck.

PR #1631, opened 2022-08-16
Main Class Change Recommended

BeChris100:
# Some changes recommended

First of all, I didn't check out the branch "release/6" because one line was missing something out. I described the situation below.

## Change in the main class (core/HO.java)
 
   - fixes issue of parsing command line arguments in the Main Class (core/HO.java)
      - change the line from `if (args != null)` to `if (args.length != 0)`

## Why this change?

Java never sets the parameter "args" to null if no program arguments has been passed. Instead, Java Runtime counts the Program Arguments from first to end. In case of passing no arguments, Java does not set the parameter "args" to null. For this case where no arguments has been passed, Java still initializes the parameter "args" as `new String[0];`.

## How do I know this?

Few months ago, I was experimenting with Java, which is my usual thing to do. At first, I wanted to check the CLI Arguments if there are no passed, so I put `System.out.println(args != null);` in the main function. It returned as true. So I knew that Java does not initialize `args` as null, but sets it as `new String[0]`.


wsbrenk (changes requested):
thanks for your input.

release/6 branch is not planned to be merged to master branch. Please checkout the master branch and create a new pull request.

Issue #1593, opened 2022-07-22
[FEATURE] Replace DBManager's EscapeSequences methods


wsbrenk:
i think the existing insert- and deleteEscapeSequences are error prone.

for example they will deliver wrong results if the original strings contained '#' or '§' characters.

To get rid of escaping, the use of prepared statements should be considered.

wsbrenk:
@tychobrailleur what are your thoughts concerning this?

wsbrenk:
#1592 found this escapes here. maybe they are the reason for that issue!?

tychobrailleur:
Yes, I would agree prepared statements are preferable, safer and clearer. I think it would be a lot of work to convert all SQL statements into prepared statements, though?

wsbrenk:
@tychobrailleur work is in progress. the crucial thing is how to migrate all the strings in database with escaped characters. Up to now i have no good idea concerning this. i fear this will become a very long running migration.

wsbrenk:
![image](https://user-images.githubusercontent.com/11821502/181372682-a41f1f1f-1574-4523-9005-2d9b4163d249.png)

PR #1637, opened 2022-09-20
#1593 Feature/db prepared statements

wsbrenk:
 
   - fixes issue #1593
   
 - [x] have been updated
 - [ ] do not require update


src/main/java/core/db/AbstractTable.java
```diff
@@ -3,74 +3,168 @@
...
+			valuePlaceholders.append(sep).append("?");
+			ret.append(sep).append(c.getColumnName());
+			sep = ",";
+		}
+		ret.append(")").append(valuePlaceholders).append(")");
+		return ret.toString();
+	}
+	protected int executePreparedInsert(Object ... values){
+		if ( preparedInsertStatementBuilder==null){
+			preparedInsertStatementBuilder=createPreparedInsertStatementBuilder();
```
tychobrailleur, line 92: I must admit I don't quite understand the added complexity of using `PreparedStatementBuilder`, which AFAICT don't do much (other than creating the `PreparedStatement` and caching it).  I thought this could have been because they get overridden in subclasses, but that doesn't seem to be the case either.  Would it not be simpler to call directly `JDBCAdapater#createPreparedStatement`?  Am I missing something?

Also, I am worried that those statements are left open without seemingly being closed, which could cause some DB issues.   Is that the case?  A while back I had started some work to ensure all statements and result sets were getting closed properly, but hadn't had the chance to finish...  Except now it looks like it would be straightforward, as all the executions seem to be localised to this class.

src/main/java/core/db/AbstractTable.java
```diff
@@ -3,74 +3,168 @@
...
+	protected PreparedInsertStatementBuilder createPreparedInsertStatementBuilder(){
+		return new PreparedInsertStatementBuilder(this);
+	}
+	private String createInsertStatement() {
+		var ret = new StringBuilder("INSERT INTO ");
+		ret.append(getTableName());
+		var valuePlaceholders = new StringBuilder(" VALUES ");
+		var sep = "(";
+		for (var c : columns) {
+			valuePlaceholders.append(sep).append("?");
```
tychobrailleur, line 83: There is a bit of obfuscation here, you have to follow carefully to see that the `sep` `(` will become the opening paren for the `VALUES` clause, and then becomes `,` for separating column names.

I like streams for that, something like (untested):


```
		var ret = new StringBuilder("INSERT INTO ");
		ret.append(getTableName());

		ret.append(" (")
				.append(Arrays.stream(columns)
				.map(ColumnDescriptor::getColumnName)
				.collect(Collectors.joining(",")));
		ret.append(") VALUES (")
				.append(String.join(",", Collections.nCopies(columns.length, "?")))
				.append(")");

		return ret.toString();
```

And even that is not necessarily easy to follow,,,

src/main/java/core/db/DBUpdater.java
```diff
@@ -76,6 +77,46 @@ void updateDB(int DBVersion) {
...
+	private PreparedStatement getMigrateEscapesStatement(String table, String column, String where){
+		var sql = "Update "+ table + " SET " + column + "=REPLACE("+column+", ?, ?) " + where;
+		var ret = migrateStatements.get(sql);
+		if ( ret == null ){
+			ret = new DBManager.PreparedStatementBuilder(sql).getStatement();
+			migrateStatements.put(sql, ret);
+		}
+		return ret;
+	}
+	private void migrateEscapes(String table, String ... columns){
```
tychobrailleur, line 99: There is code duplication here, this method can probably be written using the other one.

So replacements are:
- `§` -> `\`
- `#` -> `'`

Correct?

src/main/java/core/db/MatchesOverviewQuery.java
```diff
@@ -72,27 +60,36 @@ static int getMatchesKurzInfoStatisticsCount(int teamId, int matchtype, int stat
 		return tmp;
 	}
 
+//	private static final HashMap<String,PreparedStatement> preparedStatements = new HashMap<>();
```
tychobrailleur, line 63: You can remove, I guess

src/main/java/core/db/EconomyTable.java
```diff
@@ -68,38 +70,58 @@ protected String[] getCreateIndexStatement() {
...
-							+ economy.getCostsSum() + ", " + economy.getExpectedWeeksTotal() + ", " + economy.getLastIncomeSponsors() + ", " + economy.getLastIncomeSponsorsBonus() + "," + economy.getLastIncomeSpectators() + ", "
-							+ economy.getLastIncomeFinancial() + ", " + economy.getLastIncomeTemporary() + ", " + economy.getLastIncomeSum() + ", " + economy.getLastCostsPlayers() + ","
-							+ economy.getLastCostsStaff() + ", " + economy.getLastCostsArena() + ", " + economy.getLastCostsYouth() + ", " + economy.getLastCostsFinancial() + ", "
-							+ economy.getLastCostsTemporary() + ", " + economy.getLastCostsSum() + ", " + economy.getLastWeeksTotal() + ", " + economy.getExpectedCash() + ", "
-							+ economy.getIncomeSoldPlayers() + ", " + economy.getIncomeSoldPlayersCommission() + ", " + economy.getCostsBoughtPlayers() + ", " + economy.getCostsArenaBuilding() + ", "
-							+ economy.getLastIncomeSoldPlayers() + ", " + economy.getLastIncomeSoldPlayersCommission() + ", " + economy.getLastCostsBoughtPlayers() + ", " + economy.getLastCostsArenaBuilding()
-							+ ")";
-			adapter.executeUpdate(statement);
+			executePreparedDelete( hrfId);
+			executePreparedInsert(
```
tychobrailleur, line 81: Neat! :-)

src/main/java/core/db/StatisticQuery.java
```diff
@@ -8,31 +8,68 @@
...
 import java.sql.ResultSet;
 import java.sql.Timestamp;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
+import java.util.stream.Collectors;
 
 
 public class StatisticQuery {
 
+	private static String getPlaceholders(int anzahl){
```
tychobrailleur, line 21: That could be put in a util class  and used in that sample code I posted above!

tychobrailleur: Also, could you please rename `anzahl` to `count`?  I get what it means, but not everyone would! ;-)

src/main/java/core/db/TransferTypeTable.java
```diff
@@ -19,23 +19,17 @@ protected void initColumns() {
...
-        	adapter.executeUpdate("insert into " + getTableName()
-                                                          + " (PLAYER_ID, TYPE) values ("
-                                                          + playerId + "," + type + ")"); 
-        }
+    void setTransferType(int playerId, int type) {
+        executePreparedDelete(playerId);
+        executePreparedInsert(
+                playerId,
+                type
+        );
```
tychobrailleur, line 28: You seem to have gone for this pattern in multiple places of first deleting the existing entry of same ID, and then inserting the new one, rather than updating – which means that this operation is no longer atomic.  This could potentially lead to some troubles.  Any concern there?

src/main/java/core/model/match/MatchKurzInfo.java
```diff
@@ -196,7 +196,7 @@ public final void setGuestTeamGoals(int m_iGastTore) {
 	 * 
 	 * @return Value of property m_iGastTore.
 	 */
-	public final int getGuestGuestGoals() {
+	public final int getGuestTeamGoals() {
```
tychobrailleur, line 199: :+1:

src/main/java/tool/injury/InjuryDetailPanel.java
```diff
@@ -110,17 +109,20 @@ final int getTSIPre() {
         }
     }
 
+
+    private static final DBManager.PreparedStatementBuilder preStatementBuilder = new DBManager.PreparedStatementBuilder(
+            "select marktwert from SPIELER where spielerid=? and verletzt=-1 order by DATUM desc");
+    private static final DBManager.PreparedStatementBuilder postStatementBuilder = new DBManager.PreparedStatementBuilder(
+            "select marktwert from SPIELER where spielerid=? and verletzt>-1 order by DATUM desc");
```
tychobrailleur, line 116: We should push those SQL queries out of the view.  Not sure where the best place is, but grouping those SQL queries in a controller-type class would be preferable.  Maybe in `Player`?

src/main/java/module/transfer/history/PlayerDetailPanel.java
```diff
@@ -253,7 +254,8 @@ private void refreshPlayerTable(List<PlayerTransfer> values) {
      */
     private void updatePanel() {
         if (playerId > 0) {
-            if (DBManager.instance().getIsSpielerFired(playerId)) {
+            var playerNotes = DBManager.instance().loadPlayerNotes(playerId);
```
tychobrailleur, line 257: `loadPlayerNotes` should probably a `Player` method.

tychobrailleur (commented):
I haven't been able to go through the whole review unfortunately, but I think I looked at the important locations.  Some great cleanups in there, and prepared Statements are definitely a great change, thanks for doing this!!

I have added some comments, hopefully they make sense.  Let me know if you have questions.

src/main/java/core/db/AbstractTable.java
wsbrenk: @tychobrailleur  thank you very much for your input!!!

concerning Builder: I introduced the builder when it became annoying for me to type the same three lines over and over again. later on i found it good to create specialized Builder with standard sqls for inserting, updating and deleting in abstractTable. Overriding is delegated to the standard Create-Builder methods provided in AbstractTable.

concerning Close: I hope they don't need any closing, since their usage is reusing. If I'm wrong, we now have a central place to handle this, when ho is terminated.

src/main/java/core/db/AbstractTable.java
wsbrenk: good suggestion, i will change that accordingly

src/main/java/core/db/DBUpdater.java
wsbrenk: you're right - changed this.

src/main/java/core/db/MatchesOverviewQuery.java
wsbrenk: is removed.

src/main/java/core/db/EconomyTable.java
wsbrenk: there were only a few places that used a sql update and i thought it would be a good idea to make it unique.  do you think that it is better to use the update ?

src/main/java/core/db/StatisticQuery.java
wsbrenk: moved it to DBManager.

src/main/java/core/db/TransferTypeTable.java
wsbrenk: the advantage i see for this pattern is, that it works in both cases: initial creation of the record and its update. 
checking if records exists and doing either an update or an insert depending on the result sounds more complicated - but maybe, this is a mistake?!

src/main/java/tool/injury/InjuryDetailPanel.java
wsbrenk: changed to: 
~~~
           String tsi = player.getLatestTSINotInjured();
...
            tsiPre.setText(tsi);

            tsi = player.getLatestTSIInjured();
~~~

src/main/java/module/transfer/history/PlayerDetailPanel.java
wsbrenk: Player.getNotes is already available. 

Refactoring of module transfer is another project. i don't want to start it now. do not understand yet if players can appear here whose id is known, but who are not stored in the table SPIELER!?!?

src/main/java/core/db/EconomyTable.java
tychobrailleur: I would be more comfortable with update, yes.  My worry is that if for some reason, HO is killed between delete and insert, the record will be lost, and this may cause further troubles.

src/main/java/core/db/AbstractTable.java
tychobrailleur: Ok, it would be good to keep an eye on all these cached preparedstatements, and see if that causes issues further down the line.

src/main/java/core/db/AbstractTable.java
tychobrailleur: Thanks!

src/main/java/tool/injury/InjuryDetailPanel.java
tychobrailleur: Definitely much better, thanks!

src/main/java/module/transfer/history/PlayerDetailPanel.java
tychobrailleur: FAir enough, not a big deal, something to keep in mind.

src/main/java/core/db/EconomyTable.java
wsbrenk: OK  - what would you suggest then

- checking existance of record before inserting or updating (should be as slow as deleting;-)
-  saving status in model classes when loading record from db (isAvailableInDb)

wsbrenk: a day of reflection, brings me to the idea to implement the 2nd plan times

wsbrenk: latest commit implements the update in SPIELERTABLE (delete is removed).

one thing, that is not so good: the great similarity of update and insert method.

wsbrenk: with latest commit the methods store, update and insert are in AbstractTable and shorter than a limerick.

wsbrenk:
@tychobrailleur now refactoring of all tables is done. deleting of records is only done now when list of records have to be updated. would be nice if you could look over it again and give me some hints where I should document or change something.

PR #1863, opened 2023-04-29
Fix click on “Last Match” column when it has been moved.

tychobrailleur:
This fixes the click on “Last Match” column if the column has been moved.

The “Last Match” column is clickable, and opens:
- The match in Hattrick when shift-clicking,
- The match in HO when double-clicking.

This commit also adds some comment to clarify the behaviour.

3. `src/main/resources/release_notes.md` ...

 - [x] has been updated
 - [ ] does not require update

4. [Optional] suggested person to review this PR @wsbrenk


wsbrenk:
@tychobrailleur Hi Sebastién, i'm so happy to see you here again.

I have failed to set up the master branch for the next HO version until today. That would be HO8. For bug fixes and minor changes, there is the release/7 branch, which already builds beta 7.1.

if you want to include this to version 8, we should change build.gradle accordingly. I will do this before the requests are merged. 
if you want to include this to version 7.1, please change your request to merge into release/7

wsbrenk:
My rule of thumb regarding minor/major changes is: As soon as the database needs to be changed, it is most likely a major change and should no longer be included in the 7 release. Unless there is a bug in the DB that needs to be fixed.

tychobrailleur:
Hi Sebastién, i'm so happy to see you here again.

I always lurk here, even though I don't always have time to work on HO. ;-)

> if you want to include this to version 8, we should change build.gradle accordingly. I will do this before the requests are merged. if you want to include this to version 7.1, please change your request to merge into release/7

Sure I'll do that, I think it's a small low risk change, so the release branch should be ok.

wsbrenk:
the release notes needs some merging

tychobrailleur:
> the release notes needs some merging

Apologies, I hadn't noticed.

PR #1867, opened 2023-04-30
Minor UI improvement in teamAnalyzer.

tychobrailleur:
This adds a bit of padding around the check boxes on the teamAnalyser panel, and centers the automatic / manual radio buttons.

Screenshot from 2023-04-30 18-49-55

It also cleans up code a bit, in particular in lineup assistant.  Probably more effort to come on that front.

Cf. above.

 - [ ] has been updated
 - [x] does not require update


tychobrailleur:
@wsbrenk While working on this, I came across this piece of code that IntelliJ says is unreachable (the whole `for`):
https://github.com/akasolace/HO/blob/master/src/main/java/module/lineup/assistant/LineupAssistant.java#L535 because `zusPos` remains always empty, therefore `zusPos.size()` is always 0.  Is that expected, or is there something missing here?

wsbrenk:
@tychobrailleur IntelliJ is in the right - as almost always! I rummaged through the git history and found no clue that would speak against deleting the for loop.

Issue #1782, opened 2023-01-05
[FEATURE] Schum rating prediction


wsbrenk:
this is from [Schum's thread](https://www89.hattrick.org/Forum/Read.aspx?t=17404127&n=55&v=4&mr=0)

Midfield rating multipliers.

As usual, let's start with the coordinate system.

If you remove all players in the installer, then all ratings will be visually displayed as 0.75
In reality, it is 0, or "absent".

Next, I will give a table

Display | Value
-- | --
0.75 | 0.00
1.00 | (0.00 ; 0.25)
1.25 | [0.25 ; 0.50)
1.50 | [0.50 ; 0.75)
1.75 | [0.75 ; 1.00)
2.00 | [1.00 ; 1.25)
2.25 | [1.25 ; 1.50)
2.50 | [1.50 ; 1.75)
2.75 | [1.75 ; 2.00)
3.00 | [2.00 ; 2.25)

and so on, ad infinitum...

And now the values of the PM multipliers:

Type | Value
-- | --
Away | 1.00
Derby | 1.10
Home | 1.20
PIC | 0.84
Normal | 1.00
MOTS | 1.10(6)*
CA | 0.93
LS | 0.96

*the exact value of the MOTS multiplier is 83/75=1.10(6)
The values of the remaining multipliers do not contain more than 2 decimal places.

Away - away match
Derby - derby away, national team at home
Home - Home Match
PIC - Half Strength Game
MOTS - Match of the Season
CA - Counter-Attack Tactics
LS - Tactics "Long Strikes"

The unwritten manual has a similar table [17236276.15](https://www89.hattrick.org/Forum/Read.aspx?t=17236276&v=4&a=1&n=15)
Now the built-in rating calculator allows you to find refined values that are very close.

As a follow-up to the previous post, I'll quote downgrade defense and attack ratings for different tactics:

**Protection ratings:**
0.85 - The tactic "Attack on the flank" reduces the rating of the defense in the center by 15%
0.85 - "Attack in the center" tactic reduces the ratings of the defense on the flanks by 15%
0.93 - The "Play Creative" tactic reduces all protection ratings by 7%

**Attack ratings:**
0.96 - Long-Range Strike Tactics Reduce All Attack Ratings by 4%

The values are precise, not rounded. There are no decimal places below.

**Retreat to the defense.**

From the Rules:
[...]

PR #1920, opened 2023-09-24
Feature/refactor rating prediction

wsbrenk:

   - fixes issue #1782 (wip)

All rating calculation is now in one class (RatingPredictionModel), which implements the formula presented by Schum.
Next I would like to implement the possibility to override it's calculation methods by groovy scripts. 
The effect of retreat event is also still missing.


src/main/java/core/model/HOModel.java
```diff
@@ -179,14 +181,14 @@ public void storeLineup(MatchLineupTeam matchLineupTeam) {
...
 
-    /**
-     * returns the lineup (setRatings is called)
-     */
-    public final MatchLineupTeam getCurrentLineupTeamRecalculated() {
-        getCurrentLineupTeam();
-        m_clAufstellung.getLineup().setRatings();
-        return m_clAufstellung;
-    }
+//    /**
```
tychobrailleur, line 184: May as well remove.

src/main/java/core/model/player/Player.java
```diff
@@ -312,12 +310,13 @@ public class Player extends AbstractTable.Storable {
     private Integer m_lastMatchRating;
     private Integer lastMatchRatingEndOfGame;
 
+    // TODO this should not be an attribute of Player. Move it to lineup (MatchLineupPosition or something like this)
```
tychobrailleur, line 313: Agreed!!

src/main/java/core/model/player/Player.java
```diff
@@ -906,47 +904,111 @@ public void setHrfDate() {
...
+            referenceKeeper.setSkillValue(PASSING, 1);
+            referenceKeeper.setSkillValue(WINGER, 1);
+            referenceKeeper.setSkillValue(SCORING, 1);
+            referenceKeeper.setSkillValue(EXPERIENCE, 1);
+        }
+        return referencePlayer;
+    }
+
+
+//    /**
```
tychobrailleur, line 949: Can we just remove, and the same below?

src/main/java/core/rating/RatingCalculationCache2.java
```diff
@@ -0,0 +1,26 @@
+package core.rating;
+
+import java.util.HashMap;
+
+abstract class RatingCalculationCache2<T1, T2> {
```
tychobrailleur, line 5: What is the purpose of the `RationCalculationCache` classes?  I wonder if we could not use generic methods instead of introducing new classes for each?

src/main/java/core/rating/RatingPredictionManager.java
tychobrailleur, line 1: Should this class be removed altogether?

src/main/java/core/rating/RatingPredictionModel.java
```diff
@@ -0,0 +1,1419 @@
...
+        if (  roleId == KEEPER){
+            reference = Player.getReferenceKeeper();
+        }
+        else {
+            reference = Player.getReferencePlayer();
+        }
+        return  getPlayerRating(p, roleId, behaviour, minute) /  getPlayerRating(reference, roleId, behaviour, minute);
+    }
+
+    RatingCalculationCache2<Player, Integer> playerTaticStrengthCache = new RatingCalculationCache2<>() {
```
tychobrailleur, line 998: Ah I see it now.

src/main/java/core/rating/RatingPredictionParameter.java
```diff
@@ -108,7 +108,7 @@ public double getParam (String section, String key, double defVal) {
     	if (allProps.containsKey(section)) {
     		Properties props = allProps.get(section);
     		String propString = props.getProperty(key, "" + defVal);
-    		if (propString != "") {
+    		if (!Objects.equals(propString, "")) {
```
tychobrailleur, line 111: or `!propString.equals("")`

src/main/java/module/lineup/Lineup.java
```diff
@@ -101,7 +97,15 @@ public void setPlayers(List<MatchLineupPosition> matchLineupPositions) {
 		}
 	}
 
-	private static class Settings {
+	public long getRatingRevision() {
+		return ratingRevision;
+	}
+
+    public void setSubstitutionsUpdated() {
+		ratingRevision++;
```
tychobrailleur, line 105: What is `ratingRevision` for?

tychobrailleur (approved):
Wow, fantastic PR!!  Lots of cleanup!  Lots of commented out stuff that can still be removed, but that's great work

src/main/java/core/rating/RatingPredictionManager.java
wsbrenk: yes, I kept it for the time being to be able to look something up

src/main/java/core/rating/RatingPredictionParameter.java
wsbrenk: changed to if (!propString.isEmpty()) {...

this class also might be removed, if we remove RatingPredictionManager. It could be useful if we implement the groovy api to override the new RatingPredictionModel. One candidate of that could be an implementation of the current rating prediction.

src/main/java/module/lineup/Lineup.java
wsbrenk: Some of the new rating caches register the lineup's rating revision when they calculate. Changing of skill values and other properties, that changes the results of those calculations increments the revision number. Then the caches are resetted.

wsbrenk: comment added

PR #1951, opened 2023-11-22
Add Kotlin support to project.

tychobrailleur:
Also add a unit test acting as a smoke test, and experimenting with some Kotlin features.

Cf. above

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/test/java/core/util/KotlinBehaviourTest.kt
+++ b/src/test/java/core/util/KotlinBehaviourTest.kt
@@ -1,0 +1,88 @@
+package core.util
+
+import core.file.xml.XMLManager
+import core.model.match.MatchKurzInfo
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+
+class KotlinBehaviourTest {
+
+    @Test
+    fun testMapsAddition() {
+        val map1 = mapOf("1" to 42, "2" to 53)
+        val map2 = mapOf("1" to 1, "2" to 666)
+        val map = map1 + map2
+
+        Assertions.assertEquals(1, map["1"])
+        Assertions.assertEquals(666, map["2"])
+    }
+
+    @Test()
+    fun testDoubleBangBehaviour() {
+        val doc:Document? = XMLUtils.createDocument("<example><node val=\"42\" /></example>")
+        val root = doc!!.documentElement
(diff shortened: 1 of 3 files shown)


src/test/java/core/util/KotlinBehaviourTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import core.model.match.MatchKurzInfo
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+
+class KotlinBehaviourTest {
+
+    @Test
```
wsbrenk, line 13: @tychobrailleur what do you think about starting with HO unit or integration testing? This would help not only for the kotlin port but also for further development...

tychobrailleur: You mean starting adding unit tests to code we write?  That's definitely something I'd be behind, it's been on my mind.  That's something we could do on  a ad-hoc basis for old code, and more systematically for new code.

wsbrenk: what do you think about creating a test for a class or module before starting the kotlin port to verify that after porting the software, the function is still OK?

wsbrenk (approved):

src/test/java/core/util/KotlinBehaviourTest.kt
tychobrailleur: Oh I see.  I think this is going to be a monumental task to write unit tests for all, and probably overkill.  There is also the difficulty that we currently have to write unit tests, which is that we use singletons everywhere, so as soon as you touch some significant code, you pull in the database etc. which makes things quite complex.  Maybe that's something we need to look into, seeing how to rely less on singletons?

I'll write unit tests for some critical portions, like significant calculations (the Schum stuff comes to mind), but probably not for everything, that's too much work. :-)

tychobrailleur: (Case in point: I was trying to do this for this seemingly innocuous [method](https://github.com/ho-dev/HattrickOrganizer/blob/1f95ec66516600c5483972bafe4772b1fbb0bf92/src/main/java/core/constants/player/PlayerAbility.java#L153) in `PlayerAbility`.  The init of this class causes `ITEMS` to be initialised, calling `getNameForSkill`, which in turn calls `Helper.getNumberFormat`, whose initialisation loads `HOVerwaltung.instance()` which requires loading data from DB)

PR #1952, opened 2023-11-22
Cleanup of `Player`: rename attributes and methods.

tychobrailleur:
Because of Hungarian notation in naming attributes, the `Player` class does not conform to Java Bean conventions, preventing an easy conversion from Java to Kotlin.  This commit ensures that the attributes are renamed in a consistent way (also translating from German to English appriopriate).  Getters and setters were also renamed according to Java Bean conventions.

Also, fixed a bug where reference keeper returned was incorrect.

Cf. above

 - [ ] has been updated
 - [x] does not require update


wsbrenk (approved):
very good!

now i learned that hat trick is not a trick with a hat but three goals in one match;-)

PR #1961, opened 2023-11-26
Change order of Gradle tasks.

tychobrailleur:
The `clean` task is called after `compileKotlin`, and before `compileJava`, causing the output of the Kotlin compilation to be wiped, and therefore not made available to the Java compilation.

This commit re-orders tasks to get `clean` done before the other tasks for `compileJava`.  Hopefully, this should solve the build issues encountered with Kotlin in the Github workflow...

This commit also verifies the presence of the `INSTALL4J_HOME` env var to avoid compile issue when checking out the project the first time.

cf. above

 - [ ] has been updated
 - [x] does not require update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -208,8 +208,11 @@ distributions {
 }
 
 install4j {
-    if (install4jHomeDir != null && install4jHomeDir != "") installDir = file(install4jHomeDir)
-    else installDir = file(System.getenv('INSTALL4J_HOME'))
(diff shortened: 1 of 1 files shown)


tychobrailleur:
@wsbrenk Do you see any side effects for this change?

wsbrenk (approved):
I'm not the expert to decide that - I trust you completely

Issue #816, opened 2020-11-30
[BUG]Layout of the special events tab


wsbrenk:
Feedback from wettnetzer (HO 4.1)

In the Special Events tab, you have made the line height significantly higher in one of the previous versions, which makes it more difficult to keep track, or can you set this as a user.
Perhaps a suggestion on the special events. I think it would be great if you could sort the table with a click on the column header, as is the case in the player overview.

PR #2004, opened 2023-12-22
#816

wsbrenk:

  wip:
  
refactor Special event table

@tychobrailleur  what to do concerning icons:
- replace png by svg icons
- create a getIcon that can handle both kinds (current solution doesn't work)

TODO: 
- [x] icon handling
- [x] layout save and restore 

 `src/main/resources/release_notes.md` ...

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/gui/comp/entry/ColorLabelEntry.java
```diff
@@ -236,15 +236,39 @@ private void setGraphicalChangeValue(double number) {
...
     }
 
+    private void initAdd() {
+        var isInitDone = this.getLayout() != null && this.getLayout().getClass() == BoxLayout.class;
+        if (!isInitDone){
+            this.setLayout(new BoxLayout(this, BoxLayout.X_AXIS));
+            var icon = this.getIcon();
+            if (icon!=null){
+                super.setIcon(null);
+                this.add(new JLabel(icon));
```
tychobrailleur, line 246: Doesn't this get called one time too many the first time in `addIcon`?  I must admit I don't follow the logic. :)

tychobrailleur:
The approach we had decided on with akasolace was to convert to SVG, much easier to deal with.

tychobrailleur:
There is a lot of commented out code that can be removed!

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
wsbrenk: this should handle the case if setIcon was called before addIcon is called. This icon will be added to the list of components. The added icon is then appended. (that's the plan)

wsbrenk:
currently I'm looking for svg replacements to the existing png icons. Unfortunately, Hattrick itself still uses PNG icons and e.g. the Rotate.png is nowhere to be found as SVG.

wsbrenk:
next commit will remove the comments.

Issue #1242, opened 2021-10-18
[FEATURE] Allow users to edit colors ....


wsbrenk:
Feedback from peppermill at HT forum:

Squad:
Can you take another color here for dark green, green and blue is hard to distinguish 

![image](https://user-images.githubusercontent.com/11821502/137686230-e58c5e39-0a2d-4ecc-b469-5c919902f87d.png)

akasolace:
@wsbrenk I renamed this one cause I think we can do a more general feature ... what do you think ?

wsbrenk:
![image](https://github.com/ho-dev/HattrickOrganizer/assets/11821502/7ac75756-80e0-4b44-b06e-b0a26c40a138)

wsbrenk:
Now i will add a color picker

tychobrailleur:
Nice :)

wsbrenk:
The following actions i plan to implement:

* editing the color reference column will show a combo box with all color names
* selecting a new color reference will change the color value accordingly (the referenced color will be displayed and the color value editing will be disabled)
* if no color reference is selected editing the color value cell should invoke a color picker to set a new color  value
* the user settings will be stored in the new color database table which override the standard settings on HO startup

Currently I'm struggling with the empty combo box entry in color reference combo box, which is a very slim entry which is not easily selectable...

wsbrenk:
Now it is ready for review

TODOs (new ticket candidate): There were 200 color names defined in HO. Some of them are removed by this merge request. It would be useful to remove more of them and name them more uniquely to make editing of colors more easier.

PR #2017, opened 2024-02-02
#1242 color editor

wsbrenk:

   - fixes issue #1242

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/db/HOColorTable.java
```diff
@@ -0,0 +1,39 @@
...
+                ColumnDescriptor.Builder.newInstance().setColumnName("NAME").setGetter((o) -> ((HOColor) o).getName()).setSetter((o, v) -> ((HOColor) o).setName((String) v)).setType(Types.VARCHAR).isNullable(false).setLength(255).build(),
+                ColumnDescriptor.Builder.newInstance().setColumnName("THEME").setGetter((o) -> ((HOColor) o).getTheme()).setSetter((o, v) -> ((HOColor) o).setTheme((String) v)).setType(Types.VARCHAR).isNullable(false).setLength(255).build(),
+                ColumnDescriptor.Builder.newInstance().setColumnName("COLOR_REFERENCE").setGetter((o) -> ((HOColor) o).getColorReference()).setSetter((o, v) -> ((HOColor) o).setColorReference((String) v)).setType(Types.VARCHAR).isNullable(true).setLength(255).build(),
+                ColumnDescriptor.Builder.newInstance().setColumnName("VALUE").setGetter((o) -> ((HOColor) o).getValue()).setSetter((o, v) -> ((HOColor) o).setValue((Integer) v)).setType(Types.INTEGER).isNullable(true).build(),
+        };
+    }
+
+    @Override
+    protected String createSelectStatement() {
+        return createSelectStatement("WHERE THEME=?");
```
tychobrailleur, line 33: Do we need an index on `THEME`?

src/main/java/core/option/UserColorsPanel.java
```diff
@@ -0,0 +1,340 @@
...
+	 * @return Image Panel
+	 */
+	private JPanel getTopPanel() {
+		JPanel panel = new ImagePanel();
+		var themes = ThemeManager.instance().getRegisteredThemes();
+		var names = themes.stream().map(Theme::getName).toArray();
+		skins = new JComboBox(names);
+		var selected = ThemeManager.getCurrentThemeName();
+		skins.setSelectedItem(selected);
+		skins.addActionListener(this);
```
tychobrailleur, line 113: These days I prefer to use a lambda here rather than making the panel an `ActionListener`, but up to you.

src/main/java/core/option/UserColorsPanel.java
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	/**
+	 * Compare two color settings
+	 * @param c1 Color setting, may be null
+	 * @param c2 Color setting, may  be null
+	 * @return True, if the settings are different
+	 */
+	private boolean AreDifferentColors(HOColor c1, HOColor c2) {
```
tychobrailleur, line 304: Can you please lowercase this method and the next one, and push into HOColor?

src/main/java/core/option/UserColorsPanel.java
```diff
@@ -0,0 +1,340 @@
...
+		return !o1.equals(o2);
+	}
+
+	/**
+	 * Get the color value of the color setting
+	 * If no value is set, the corresponding reference is searched in the internal setting list.
+	 * @param currentColor Color setting
+	 * @return Color value
+	 */
+	public Color getColor(HOColor currentColor) {
```
tychobrailleur, line 327: This also probably belongs in `HOColor`.

src/main/resources/sprache/English.properties
```diff
@@ -2276,3 +2276,158 @@ ls.copyright.authoroftheformulas=The author of the formulas
...
+ls.color.tableentry_bg=Table entry
+ls.color.tableentry_fg=Table entry font color
+ls.color.background_container=Container background
+ls.color.tableentry_improvement_fg=Table improvement entry font color
+ls.color.tableentry_decline_fg=Table decline entry font color
+ls.color.rating_border_below_limit=Rating border below limit
+ls.color.rating_border_above_limit=Rating border above limit
+ls.color.skillentry2_bg=2nd skill entry font color
+ls.color.player_pos_bg=Player position background
+ls.color.player_subpos_bg=Player position behaviour
```
tychobrailleur, line 2319: US English seems to be the language of choice (cf. “color” & “gray”) so for consistency this should be “behavior.”

tychobrailleur (changes requested):
As I said before, this is a really nice feature -- and I wonder if this should be extended to other theme features, including Font, font size, etc.

src/main/java/core/db/HOColorTable.java
wsbrenk: I haven't thought about it yet. 
However, I assume that the normal case will be that only a single value will be stored in the column.
Isn't an index more of a hindrance?

src/main/java/core/option/UserColorsPanel.java
wsbrenk: good point - changed this.

src/main/java/core/option/UserColorsPanel.java
wsbrenk: done

src/main/java/core/option/UserColorsPanel.java
wsbrenk: this method works with the copied colors of the panel. it has to stay in the UserColorsPanel.

src/main/resources/sprache/English.properties
wsbrenk: thanks - learned again.
But i see, that there are a lot of other cases, where this is wrong...

src/main/java/core/option/UserColorsPanel.java
tychobrailleur: Ok, I must misunderstand what `colorReference` does then.

This method could still be split into 2, one that's in `HOColor`, which returns the `getColor` colour, and a second argument which is a reference colour by default if there is no `getColor`, and then the panel-specific code, which I believe is that `colors.stream()...` part.  But it's only a suggestion, to segregate responsibilities.  I'll leave it to your judgment. ;-)

src/main/resources/sprache/English.properties
tychobrailleur: Very possibly, I think this hasn't been super-consistent over the years.  Maybe we need to decide which one we pick, and create a ticket to update (I can take care of it).  My preference would be UK english, being in Ireland and all, and making more sense in the context of football.

src/main/java/core/db/HOColorTable.java
tychobrailleur: A hindrance, in what way?  

If we don't expect that table to contain many records, then no need for an index.

src/main/java/core/option/UserColorsPanel.java
wsbrenk: if I understand your suggestion correctly, streaming would have to be done in any case, even if the color in currentColor should have a value.
I don't think that's so good.

src/main/resources/sprache/English.properties
wsbrenk: Now I got it

src/main/java/core/option/UserColorsPanel.java
wsbrenk: Now I got it

tychobrailleur (approved):

Issue #1941, opened 2023-11-12
[8.0 R530] File>Update crashes


Yani-ABDESSELAM:
File>Update crashes

**Additional information**

File>Update crashes
V8.0 DEV (r530)

**java.lang.NullPointerException: Cannot invoke "Object.toString()" because "preparedStatement" is null**

_java.lang.NullPointerException: Cannot invoke "Object.toString()" because "preparedStatement" is null
	at core.db.JDBCAdapter.executePreparedUpdate(JDBCAdapter.java:134)
	at core.db.AbstractTable.executePreparedInsert(AbstractTable.java:370)
	at core.db.AbstractTable.insert(AbstractTable.java:111)
	at core.db.AbstractTable.store(AbstractTable.java:99)
	at core.db.MatchesKurzInfoTable.storeMatchKurzInfos(MatchesKurzInfoTable.java:84)
	at core.db.DBManager.storeMatchKurzInfos(DBManager.java:1177)
	at core.net.OnlineWorker.getMatches(OnlineWorker.java:531)
	at core.net.DownloadDialog.startDownload(DownloadDialog.java:322)
	at core.net.DownloadDialog.actionPerformed(DownloadDialog.java:87)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:297)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6626)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3389)
	at java.desktop/java.awt.Component.processEvent(Component.java:6391)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5001)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
[...]

PR #2018, opened 2024-02-03
#1941 repair failed V5 database update

wsbrenk:

   - fixes issue #1941

 - [x] has been updated
 - [ ] does not require update


tychobrailleur:
Did it fail for just 500 updates, or also subsequent ones?  Do you think there are other databases out there with similar issues?

This reminds me that a long time ago, I had started a [HO DB Repair](https://github.com/tychobrailleur/hodb-repair) tool which was trying to solve that kind of problems more generically, but it is way out of date, at this point.

wsbrenk:
In case of #1242, i only found missing database components of the v5 update. This change will help the reporter to repair his database - i have no idea if there are other users with the same issue.
So a more generic solution of this issue would be helpful and would override this changes.

What do you think about a solution which describes the database table and a database init method, that will compare the existing table with this description and invokes the necessary changes?

wsbrenk:
Forget it - this idea won't work in all circumstances. Some historical changes could be lost that will lead to not acceptable data loss.

tychobrailleur:
> What do you think about a solution which describes the database table and a database init method, that will compare the existing table with this description and invokes the necessary changes?

Another approach could be to ignore versions for database schema updates, and apply them at startup if they haven't been yet, regardless of version (not sure what startup penalty this would cause, however).  That and make sure changes are backward compatible (i.e. no table or column drop, no rename, etc.).

A long time ago I had suggested to use database migrations (like Flyway or liquibase) to deal with those database changes, this could be another approach, but that's perhaps a bit overkill.

tychobrailleur (approved):

Issue #2020, opened 2024-02-05
[FEATURE] Improve team analyzer tab


tychobrailleur:
- Solve the weird button locking when downloading details
- Add some information panel that gives some details about the selected team
- Make Special Events optional through a module parameter (on by default)

Ongoing work there: https://github.com/ho-dev/HattrickOrganizer/tree/feature/improve_ta

tychobrailleur:
Here is the current look, still working to get a better UX:

![Screenshot from 2024-02-25 13-27-14](https://github.com/ho-dev/HattrickOrganizer/assets/114285/befb0825-7a40-4629-a497-b16a19722790)


Things to note:
- Current streak will be shown with bullets of colour, like in the table in the "Series" tab
- Position will be correct :-)  For this and the previous item to work, there will be additional requests required
- I have moved the Update button beside the Analyse one 
- I am trying to figure out why not all columns in the match table are populated.
- Also, there is a number (called "apperance") beside each player's name, but I don't know what it does; I'll try to figure it out.


Feedback and comments more than welcome!

wsbrenk:
> Also, there is a number (called "apperance") beside each player's name, but I don't know what it does; I'll try to figure it out.

I think this is the number of matches the player played within the list of selected matches (when the average row is selected).

wsbrenk:
@tychobrailleur do you want to continue working on this ticket? If so - shall we release it with HO8 or should we postpone it to new tickets in milestone for HO9?

tychobrailleur:
> @tychobrailleur do you want to continue working on this ticket? If so - shall we release it with HO8 or should we postpone it to new tickets in milestone for HO9?

Yes, I want to continue on this one, but I'd like to have some users' feedback.  Maybe we can close this one -- and release with HO8 as is, and create a new one for further improvements for HO9.  How does that sound?

wsbrenk:
@tychobrailleur sounds good to me. So let us start the beta phase and maybe there will be some feedback and we could decide if we realize that in HO8.0, HO8.1 or HO9.

PR #2037, opened 2024-03-09
Improve layout of Team Analyzer slightly

tychobrailleur:
As per conversation at #2020, this is the first pass at improving the layout of the Team Analyzer, and add some info about the team selected in the dropdown (see screenshot at #2020).  I have removed the streak for now, as it requires to download more information. This PR is to get initial feedback and comments on these changes.

I have made a conscious effort of adding comments where appropriate to describe the existing Team Analyzer code, hopefully, it'll be easier to go through that code in the future. :)

   - fixes issue #2020

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/net/MyConnector.java
```diff
@@ -579,10 +579,9 @@ public String fetchRegionID(int teamId) {
 	}
 
 
-	public String fetchTeamDetails(int teamId)
-	{
+	public String fetchTeamDetails(int teamId) {
```
wsbrenk, line 582: @tychobrailleur do we need both methods get- and fetchTeamDetails? Is there a chance to handle it with one of them?

src/main/java/module/teamAnalyzer/ht/HattrickManager.java
```diff
@@ -1,54 +1,54 @@
...
+     *
+     * <p>Recent tournament are added if on manual, or if they are wanted, in addition to
      * the number specified.
      *
-     * @param teamId teamid to download matches for
+     * @param teamId ID of team to download matches for
      * @param filter the match filter object.
      */
     public static void downloadMatches(final int teamId, Filter filter) {
+        System.out.println("Current Thread: " + Thread.currentThread().getName()); // FIXME Remove printout
```
wsbrenk, line 41: @tychobrailleur when do you want to fix this?

src/main/java/module/teamAnalyzer/ui/FilterPanel.java
```diff
@@ -139,39 +134,76 @@ private void jbInit() {
...
 						"UPDATE for Team " + SystemManager.getActiveTeamId());
-				//HattrickManager.downloadPlayers(SystemManager.getActiveTeamId());
+
 				// Load squad info of all teams
-				for ( var team  : TeamManager.getTeams()){
-					HattrickManager.downloadPlayers(team.getTeamId());
+				try {
+					for (var team : TeamManager.getTeams()) {
+						Map<String, String> teamDetails = HattrickManager.getTeamDetails(team.getTeamId());
+						System.out.println(teamDetails);
```
wsbrenk, line 150: @tychobrailleur should be removed too?

wsbrenk (changes requested):
some minor things

src/main/java/core/net/MyConnector.java
tychobrailleur: Wow, well spotted!  I have collapsed the two, now.

src/main/java/module/teamAnalyzer/ui/FilterPanel.java
tychobrailleur: Well that's embarrassing. :-)  Thanks for noticing, now fixed.

src/main/java/module/teamAnalyzer/ht/HattrickManager.java
tychobrailleur: Fixed too

tychobrailleur:
> some minor things

Any comment/suggestion on the UI?  Is  the flow clear enough?
Screenshot from 2024-03-10 09-23-48
One thing I was wondering is whether we should just remove “Analyze” and trigger the analysis automatically, once the update is complete.

wsbrenk:
@tychobrailleur UI looks very good to me. Concerning triggering the automatic update i was thinking about it in the past too.

Since the update is very expansive due to a number of downloads from hattrick, which are only necessary when new matches happened, i would not invoke it on each analyze invocation. I run several analyzes each time i change my own lineup. We could make it smarter if on analyze we check if new matches happened since last analyze and start the automatic download (update) only in that case, when matches had taken place since then.

wsbrenk (approved):

tychobrailleur:
> @tychobrailleur UI looks very good to me. Concerning triggering the automatic update i was thinking about it in the past too.
> 
> Since the update is very expansive due to a number of downloads from hattrick, which are only necessary when new matches happened, i would not invoke it on each analyze invocation. I run several analyzes each time i change my own lineup. We could make it smarter if on analyze we check if new matches happened since last analyze and start the automatic download (update) only in that case, when matches had taken place since then.

Ah yes, I think I vaguely remember we've had that same conversation in the past.  I'll have a think about your suggestion, this definitely sounds reasonable.

When you say “expensive,” are you referring to the time it takes to make those downloads, or are you concerned about the actual number of calls made to HT?  Are you concerned about the load on CHPP?

wsbrenk:
> When you say “expensive,” are you referring to the time it takes to make those downloads, or are you concerned about the actual number of calls made to HT? Are you concerned about the load on CHPP?

A little bit of both. But I would be more urgent about the first point, although I had only recently noticed a stupid error that must have caused a very heavy load on the chpp machines. We should avoid becoming the focus of HTs through something like this.

tychobrailleur:
>We should avoid becoming the focus of HTs through something like this.

Agreed.  We should remain good citizens of the Hattrick community.
This is partly the intent of caching there: https://github.com/ho-dev/HattrickOrganizer/blob/679816a1e3aa6a7d123ad27774799136eb920a27/src/main/java/module/teamAnalyzer/ht/HattrickManager.java#L26

Also, noticed that we potentially download the same match details in that method: https://github.com/ho-dev/HattrickOrganizer/blob/679816a1e3aa6a7d123ad27774799136eb920a27/src/main/java/core/net/OnlineWorker.java#L247  There would be room for optimisation there, I believe.

wsbrenk:
I like caching a lot - sometimes more than it's good.

In this case we have to consider the situations when i started and updated HO before i start watching the league matches  and want to analyze again, when the matches are finished. HO should be smart enough to download the new matches after they were finished.

wsbrenk:
@tychobrailleur  are you still working on this issue or will you fix the conflicts and merge this pr in the next time - my approvement ist now 2 weeks old;-)

tychobrailleur:
Thanks, I have now resolved the conflicts, and merged.  Let's see if we get more feedback from users that way!

Issue #1471, opened 2022-04-18
[FEATURE] Right-mouse button copy-paste - Authorization code


masterpatje:
**Is your feature request related to a problem? Please describe.**
This feature request isn't important, because you have to do it once. But now that I am testing multiple versions, I noticed it and thought to notify it.

When I want to copy the code from the hattrick CHPP Authorize page I have to do this with for example the Windows ctrl+v combination. It would be easy if I can do it with the mouse. If I click right mouse button I can copy the code from the hattrick CHPP Authorize page, but I can't paste it with the Right Mouse Button into the field in the application.

**Describe the solution you'd like**
It would be great if I can right-mouse button copy paste the authorisation code. 

**Additional context**
![image](https://user-images.githubusercontent.com/102049050/163848448-d26e87e8-4db5-499d-938f-fcb29e462e6b.png)

PR #2041, opened 2024-03-10
#1471 enable standard popup menus in oauth dialog's text fields

wsbrenk:

   - fixes issue #1471

 - [x] has been updated
 - [ ] does not require update

3. [Optional] suggested person to review this PR @tychobrailleur : Do you think it is a good idea replacing JTextfields with awt.Textfield?


tychobrailleur:
It is usually not a good idea to mix AWT and Swing components, in particular because of the difference in look and feel, but it seems [this article](https://www.oracle.com/technical-resources/articles/java/mixing-components.html) says they should work together.  So I guess worth a try?

tychobrailleur (approved):

Issue #752, opened 2020-10-27
changes in AutoLinueup checkboxes should (if possible) not be made via mouseListener


akasolace:
see #744

maybe add something in ListSelectionListener of LineupPanel, or see how to cascade tableEvent rom sorter to model in AsutellungSpielerTable

from @weisJ 
> Also it like TableSorter is a custom table model for HO. So if sorting events aren’t propagated then there needs to be changes made to TableSorter itself.

tychobrailleur:
I am not sure I understand what is needed here.  Is this still relevant?

akasolace:
Yes it is still relevant but really not something urgent as it is working well as it is. I opened this ticket if someday there is really nothing else to do in HO! :-)

Issue #744, opened 2020-10-21
[BUG] Best position and AutoLineup are not in sync when using darklaf theme


johari75:
**Describe the bug**
Activating or deactivating the autoLineUp flag for a player causes activation or deactivation the autoLineUp flag for another player.

**To Reproduce**
see clip in zip-file

**Screenshots**
[HO! - Hattrick Organizer 4.0 r3266 - Unterhose Kairo - 14.0.1 2020-10-21 15-11-27.zip](https://github.com/akasolace/HO/files/5416048/HO.-.Hattrick.Organizer.4.0.r3266.-.Unterhose.Kairo.-.14.0.1.2020-10-21.15-11-27.zip)

**Platform information:**
 - OS: [Windows 10 pro (10.0.19041 Build 19041)]
 - Version (Hattrick Organizer 4.0 r3266)

**Additional context**

akasolace:
@johari75 I can't reproduce. Something should be wrong with your database. Can you share it?

akasolace:
@johari75 yes please as a zip file. You only have one team, right ?

johari75:
@akasolace yes, I have only one team. My db folder is to big. 27 MB zipped

johari75:
@akasolace Here you are
[20201022_UnterhoseKairo_db.zip](https://github.com/akasolace/HO/files/5421203/20201022_UnterhoseKairo_db.zip)

akasolace:
@tychobrailleur @weisJ  I checked this one and the error comes from darklaf. The mouse pressed events are not detected correctly (at least for boolean cells I am not sure for other). Events are not always detected sometime and the reported value is incorrect: https://github.com/akasolace/HO/blob/733405bb6404efeec3eeddd5a09045885bcaad88/src/main/java/module/lineup/AustellungSpielerTable.java#L51


**with nimbus:**  best position and autolineup are in sync across the app

![EVdgy7YZA1](https://user-images.githubusercontent.com/1136496/96901114-20ebbb80-1493-11eb-83c0-4d24efcd147b.gif)


**with darklaf:**   best position and autolineup are not in sync across the app

![GeXRhbrYAp](https://user-images.githubusercontent.com/1136496/96901284-5b555880-1493-11eb-9864-4ef4e8ab8274.gif)

weisJ:
[...]

PR #2049, opened 2024-04-01
#752 LineupPlayerTable uses TableModelListener instead of MouseListener

wsbrenk:

   - fixes issue #752

 - [ ] has been updated
 - [x] does not require update

### DIFF
--- a/src/main/java/core/gui/comp/table/HOTableModel.java
+++ b/src/main/java/core/gui/comp/table/HOTableModel.java
@@ -157,8 +157,8 @@ public abstract class HOTableModel extends AbstractTableModel {
 	/**
 	 * return value
 	 * 
-	 * @param row
-	 * @param column
+	 * @param row Row number
+	 * @param column Column number
 	 * 
 	 * @return Object
 	 */
@@ -176,12 +176,7 @@ public abstract class HOTableModel extends AbstractTableModel {
 		return (m_clData != null) ? m_clData.length : 0;
 	}
 
-	@Override
-	public boolean isCellEditable(int row, int col) {
-		return false;
-	}
-
-	@Override
+    @Override
 	public final Class<?> getColumnClass(int columnIndex) {
 		final Object obj = getValueAt(0, columnIndex);
 
--- a/src/main/java/core/gui/comp/table/HOTableModel.java
+++ b/src/main/java/core/gui/comp/table/HOTableModel.java
@@ -218,6 +218,7 @@ public abstract class HOTableModel extends AbstractTableModel {
 	@Override
 	public void setValueAt(Object value, int row, int column) {
 		m_clData[row][column] = value;
+		fireTableDataChanged();
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


wsbrenk:
@tychobrailleur I really have no idea if a TableModelListener is better than a MouseListener. So I need your advice.

tychobrailleur:
> @tychobrailleur I really have no idea if a TableModelListener is better than a MouseListener. So I need your advice.

For sorting, a `MouseListener` would be more appropriate, as it is invoked in reaction to a click.  A `TableModelListener` would be react to any change to the model backing the table, so it could be anything, even things other than a click.

If I remember correctly, sorting in HO is done in a very custom way, which doesn't use the `JTable` facilities (i.e. [`TableRowSorter`](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/table/TableRowSorter.html)), but I think this is because of the very complex things like the split table players' name / rest of attributes.

What problem are you trying to solve here?


> 
> In version 9 we could really think about whether we can bring the confusingly many different tables of the HO to a common (Kotlin) solution.

Yes, agreed, we could do with a bit of consistency, and unify into a generic component that would make maintenance a bit easier!

wsbrenk:
@tychobrailleur In https://github.com/ho-dev/HattrickOrganizer/issues/744#issuecomment-715419003 Jannis recommends using a tableModelListener  - so i tried my very best, but i'm not sure if it is a better solution than the previous.

tychobrailleur:
> @tychobrailleur In [#744 (comment)](https://github.com/ho-dev/HattrickOrganizer/issues/744#issuecomment-715419003) Jannis recommends using a tableModelListener - so i tried my very best, but i'm not sure if it is a better solution than the previous.

Ah fair enough, if the expert says so! ;-)

Let me have a closer look, then...  But I sympathise with your “Widerstreben”!

tychobrailleur:
Sorry for the delay, I'm looking into this tonight.

tychobrailleur:
I have had a closer a look at the thread at #744 , and I think it makes sense to use a `TableModelListener` here: I had not fully grasped that the click was not a click to act on the table itself (for sorting), but for a value in the table (the state of that unselectable feature).  So the way you have it is good.

One thing I am not sure is the behaviour of the Best Position feature. In the Squad tab, I pick two consecutive players and mark them both as Keeper in the override, but only one appears as manually overridden in the Lineup tab

optimized

Probably a separate issue...

tychobrailleur (approved):

wsbrenk:
> Probably a separate issue...

i can not reproduce this - Are you sure you selected the same player?

wsbrenk:
Now i think i see, what you were doing. The second player you selected is your trainer i think. This player is not displayed in the lineup panel  if he can no longer play in the team.

tychobrailleur:
> Now i think i see, what you were doing. The second player you selected is your trainer i think. This player is not displayed in the lineup panel if he can no longer play in the team.

Should he appear in the table of players in the Lineup tab, then, if that's the case?  Or should we make it “unselectable“ by default, without a way to override?

PR #2105, opened 2024-07-13
fix: property 'window.about.text': adding homogeneous line break afte…

sgcr:
I saw in the UI in the "Help"/"About" dialog that the text pointing to the start of the project started looks different in other languages.
To make it homogeneous like in the base language English I added a line break to all other languages where it was missing.

Additionally I corrected the years for danish because there were different years,

### DIFF
--- a/src/main/resources/sprache/Catalan.properties
+++ b/src/main/resources/sprache/Catalan.properties
@@ -1414,7 +1414,7 @@ ls.menu.help.about = Quant a
 # About window from Top menu Help->About:  Title
 window.about.title = Crèdits
 # About window from Top menu Help->About:  Text
-window.about.text = El desenvolupament va començar el 2003 per part de Thomas Werth & Volker Fischer. Des de 2006, aquest projecte és de codi obert i desenvolupat per diferents desenvolupadors.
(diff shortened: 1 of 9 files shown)


wsbrenk (changes requested):
Good idea,  But these files are overwritten by the poeditor files when building by the github action.

So these changes need to be included in the poeditor project.

Maybe we can also make two entries out of it and hardcode the line break.

sgcr:
> Good idea,  But these files are overwritten by the poeditor files when building by the github action.
> 
> So these changes need to be included in the poeditor project.
> 
> Maybe we can also make two entries out of it and hardcode the line break.

I will have a look poeditor.

sgcr:
> > Good idea,  But these files are overwritten by the poeditor files when building by the github action.
> > So these changes need to be included in the poeditor project.
> > Maybe we can also make two entries out of it and hardcode the line break.
> 
> I will have a look at poeditor.

I signed up for an account at poeditor and waiting for approval for HO.

sgcr:
Waiting for approval for the other language I want to modify...

sgcr:
The same translations in that PR were made in poeditor.

I think this PR can be closed now.

Issue #2140, opened 2024-08-26
[FEATURE] Arena sizer: Add the general information about the current stadium


sgcr:
**Is your feature request related to a problem? Please describe.**
As a new contributor for HO I was looking for a small side project.
I saw the "Arena sizer" within the "Tools" and was missing the general information about the current stadium.
When clicking the UI, I recognized that clicking on the name of my stadium shows the same content when clicking on "Arena".
**So it might be a good idea to show in the tab with the stadium name the details about my stadium.**

**Describe the solution you'd like**
Please look at the screenshot to have an impression for the proposal.
3 group boxes from left to right containing the current, expansion and future of your stadium.
Each group box contains 3 columns: place type, places and the rate (or change) in percent.
the number of terraces, basic seating, seats under roof, seats in VIP boxes and and the total seats.
In the last line there is space for group box specific information.

The left group box **"Current"** shall show the normal information and of the stadium and in the last line is shown when the stadium was last improved or if there is a construction in progress.

The middle group box **"Expansion"** shall show the expansion of seats if there is any.
The number in the "Places" column should be with positive or negative sign in the specific color for that.
The percentage in third column "Change in %" shall show change in percent of each category (compared to the current places). The number in percent shall be shown with positive or negative sign in the specific color for that.
The last line shows when the current expansion is completed and an additional label with the remaining duration that is updated each second.

The last group box **"Future"** shall show the future of the stadium if there is an expansion in progress.
The future is the sum of current and expansion places for each type. The rate in percent shows how the rate of your stadium looks like after the expansion.
The last line shows the day when the future is reality (the next after completion).

**Proposal**
![Proposal_AreaInfoPanel](https://github.com/user-attachments/assets/92840f85-faac-463b-b6a0-2542dc4aebb9)

tychobrailleur:
Nice!  And welcome!  I like the proposal, and great contribution (in particular the tests!!).  I have added some comments to the review.

PR #2141, opened 2024-08-26
[FEATURE]: #2140: Add ArenaInfoPanel

sgcr:
   - fixes issue #2140

 - [x] has been updated


src/main/java/hattrickdata/Arena.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Getter;
+import lombok.ToString;
+
+import java.util.Optional;
+
+@Getter
+@Builder
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
```
tychobrailleur, line 15: Probably a bit late to the party, but I am not super keen on adding a dependency to lombok in HO.  Could this be turned into Java records, or better still use Kotlin?  We have been trying to move to Kotlin in recent times, and the language would naturally offer those facilities.

build.gradle
```diff
@@ -28,6 +28,7 @@ plugins {
     id "com.github.breadmoirai.github-release" version "2.5.2"
     id "com.install4j.gradle" version "9.0.6"
     id "org.barfuin.gradle.taskinfo" version "2.1.0"
+    id "io.freefair.lombok" version "8.10"
```
tychobrailleur, line 31: Probably a bit late to the party, but I am not super keen on adding a dependency to lombok in HO. Could this be turned into Java records, or better still use Kotlin? We have been trying to move to Kotlin in recent times, and the language would naturally offer those facilities.

Generally speaking we are trying to keep new dependencies to a minimum, to keep HO size in check, but also because it's proved difficult to maintain over the years.

src/main/java/core/file/hrf/HRFStringBuilder.java
```diff
@@ -79,8 +88,39 @@ public void createArena(Map<String, String> arenaDataMap) {
...
+        appendKeyValue(arenaStringBuilder, "expandingSitt", arena.getExpandedCapacity().map(Capacity::getBasic).orElse(0));
+        appendKeyValue(arenaStringBuilder, "expandingTak", arena.getExpandedCapacity().map(Capacity::getRoof).orElse(0));
+        appendKeyValue(arenaStringBuilder, "expandingVIP", arena.getExpandedCapacity().map(Capacity::getVip).orElse(0));
+        appendKeyValue(arenaStringBuilder, "expandingSseatTotal", arena.getExpandedCapacity().map(Capacity::getTotal).orElse(0));
+        appendKeyValue(arenaStringBuilder, "isExpanding", arena.getExpandedCapacity().isPresent());
+        appendKeyValue(arenaStringBuilder, "RebuiltDate", arena.getCurrentCapacity().getRebuiltDate().map(HRFStringBuilder::localDateTimeToHrfString).orElse(""));
+        appendKeyValue(arenaStringBuilder, "ExpansionDate", arena.getExpandedCapacity().map(ExpandedCapacity::getExpansionDate).map(HRFStringBuilder::localDateTimeToHrfString).orElse(""));
+    }
+
+    // TODO: duplicates from HODateTime
```
tychobrailleur, line 117: Yup :-)

src/main/java/hattrickdata/Arena.java
tychobrailleur, line 15: See above if this could be Java records, or Kotlin.  Nice thing with Kotlin is that it mixes seemlessly with Java code.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
```diff
@@ -0,0 +1,174 @@
...
+                    labelExpansionFinished.setText(EMPTY);
+                }
+        );
+    }
+
+    private static String getLanguageString(String key) {
+        return HOVerwaltung.instance().getLanguageString(key);
+    }
+
+    private static BigDecimal calculateDaysToNow(HODateTime hoDateTime, int scale) {
```
tychobrailleur, line 142: Could this somehow be pushed into `HODataTime`?  Or is there an equivalent method there, maybe?

src/main/java/tool/arenasizer/ArenaInfoPanel.java
```diff
@@ -0,0 +1,174 @@
...
+        return BigDecimal.valueOf(hoDateTime.getInstant().toEpochMilli() - HODateTime.now().getInstant().toEpochMilli())
+                .divide(BigDecimal.valueOf(1000L * 86400L), scale, RoundingMode.HALF_DOWN);
+    }
+
+    private static String toDurationString(HODateTime hoDateTime) {
```
tychobrailleur, line 147: This method has two concerns that should probably be separated, the conversion of time till now to day/hour/min/sec (which should probably be pushed to `HODateTime`), and the building of the string to display in the panel.

src/main/java/tool/arenasizer/CapacityPanel.java
```diff
@@ -0,0 +1,185 @@
...
+        labelVipBoxes.setToolTipText(getLanguageString("ls.club.arena.seatsinvipboxes"));
+        if (LONG_MODE) {
+            labelVipBoxes.setText(labelVipBoxes.getToolTipText());
+        }
+
+        labelTotal.setToolTipText(getLanguageString("Gesamtgroesse"));
+        if (LONG_MODE) {
+            labelTotal.setText(labelTotal.getToolTipText() + " " + SUM_SIGN);
+        } else {
+            labelTotal.setText(SUM_SIGN);
```
tychobrailleur, line 166: I wonder if the sum sign should be part of the localised/translation strings: I am thinking in particular of languages that are RTL rather LTR, this may cause some strange things?

src/main/java/tool/arenasizer/LabelWithSignedNumber.java
```diff
@@ -0,0 +1,84 @@
+package tool.arenasizer;
```
tychobrailleur, line 1: There are a set of global components under `core.gui`, do you think this should be long there?  In case we'd need that component somewhere else?

src/main/resources/language/German.properties
```diff
@@ -2259,4 +2261,21 @@ ls.player.lastlineup = Position letzter Einsatz
 ls.player.motherclub.name = Heimatverein
 ls.player.matchescurrentteam = Tore für aktuellen Verein
 ls.about.database.folder = Datenbank-Verzeichnis
-ls.about.logs.folder = Logging-Verzeichnis
\ No newline at end of file
+ls.about.logs.folder = Logging-Verzeichnis
```
tychobrailleur, line 2264: @wsbrenk Have we moved away from poeditor?

sgcr: I am not @wsbrenk but I will answer:

No, I think not.

I wanted to persist the change and made the change already additionally in POEditor for Spanish.

This PR has to be deleted if it works.

build.gradle
sgcr: Hi :-)

I see. ;-)

I am not familiar with Kotlin but with Java and Lombok.

The question is how to make to code better maintainable and understandable in general.

One thing is that whenever I see a getter and setter that is the standard implementation it may be annotated with the annotations `@Getter` and `@Setter` from Lombok.
This makes the code easy to read and prevents superfluous code.

I see the use of Lombok in that case as a step in between to the shift to Kotlin.
It makes it clear that the getter and setter can be moved to Kotlin easily and in the time between and it reduces superfluous code to read and maintain while making it not necessary to move everything to Kotlin in one step.

Beside that there are 42 files in Kotlin and 888 in Java.
So Kotlin is only 4,5% of the files (not lines).
When do you think is the move to Kotlin finished? Is that a long term goal?

Does the size of the library really matters?

I have to think about it...

src/main/java/core/file/hrf/HRFStringBuilder.java
sgcr: I have currently no idea to fit that in a nice way but it is on my list.
It bothers me. :-)

build.gradle
tychobrailleur: This is probably a very subjective matter, I do not think getters and setters are that much of a readability issue vs annotations.  It is a tradeoff, and between adding a new dependency and living with boilerplate code, my preference goes to the latter.  This is mostly due to issues we've previously had with third-party dependencies maintenance.  In any case Java records are a good Java alternative for the same problem, I think. And they come with yummy immutability. ;-)

On the Kotlin question, we have tried to implement new features using Kotlin to make that transition more progressive.  I had started the big bang approach of [converting the code to Kotlin](https://github.com/ho-dev/HattrickOrganizer/commits/feature/kotlin_support/), but @wsbrenk suggested that we should add unit tests as part of the migration, which obviously made things more difficult: as you probably noticed, testability is a major pain point in HO, due to these massive singletons that are everywhere.  This is something I had started to look into, but never really had the time to address further...

src/main/resources/language/German.properties
tychobrailleur: If I remember correctly (and @wsbrenk, keep me honest here!), new strings get added to `English.properties`, which then gets published to the poeditor website.

Translators go into the poeditor website, and provide the translations for their respective language there.  At build time, HO then pulls the translation files along with the new translations, and package them along with the rest of the code.

So if you need to add/change a German translation, this needs to be done on poeditor website.

There were however recently issues with that project, and @wsbrenk  floated the idea of changing the translation files directly in Github – which is not great for translators, who may not be tech savvy.

Hope this makes sense!

wsbrenk: @tychobrailleur decribes the translation workflow correctly. Again i tried to contact akasolace asking him to move the poeditor project. otherwise it will be archived again very soon. In the meantime, it occurred to me that we could create a new project at poeditor and import the property files of the current project there. The translators would then all have to register again.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
wsbrenk: Duration = HODateTime.between(HODateTime from, HODateTime to)

src/main/java/hattrickdata/CurrentCapacity.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Getter
+@SuperBuilder()
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@ToString(callSuper = true)
+public class CurrentCapacity extends Capacity {
+
+    private LocalDateTime rebuildDate;
```
wsbrenk, line 21: I would be happy, if we could change this to HODateTime.  

HODateTime was the replacement of a lot of different DateTimes in HO code. Please do not introduce those again.

src/main/java/hattrickdata/ExpandedCapacity.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Getter
+@SuperBuilder
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@ToString(callSuper = true)
+public class ExpandedCapacity extends Capacity {
+
+    private LocalDateTime expansionDate;
```
wsbrenk, line 20: HODateTime

src/main/java/hattrickdata/HattrickDataInfo.java
```diff
@@ -0,0 +1,24 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode
+@ToString
+public class HattrickDataInfo {
+
+    private String fileName;
+    private String version;
+    private int userId;
+    private LocalDateTime fetchedDate;
```
wsbrenk, line 23: HODateTime

src/main/java/tool/arenasizer/ArenaInfoPanel.java
```diff
@@ -0,0 +1,174 @@
...
+
+        setTranslation();
+    }
+
+    private static HODateTime nextDay(HODateTime hoDateTime) {
+        var localDate = hoDateTime.getLocalDateTime().toLocalDate();
+        var nextDay = localDate.plusDays(1);
+        var instant = Instant.from(nextDay.atStartOfDay(HODateTime.DEFAULT_TIMEZONE));
+        return new HODateTime(instant);
+    }
```
wsbrenk, line 105: Looks like a good method of HODateTime

wsbrenk (changes requested):
Hi @sgcr, very good job.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
sgcr: Yes.

Committed as 'Review Finding 1'

sgcr: I saw your new comment too late.
Will be fixed tomorrow.

src/main/java/hattrickdata/CurrentCapacity.java
sgcr: Yes, I saw it right after PR creation too.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
sgcr: I decided to name it `daysBetween`, `daysToNow` and `daysToNow`.

build.gradle
tychobrailleur: (BTW, I meant to say, IntelliJ is your friend for both Lombok -- converting back and forth between annotations and accessors -- and Kotlin from and to Java)

src/main/java/tool/arenasizer/ArenaInfoPanel.java
sgcr: Review Finding 2:

Added `HumanDuration`

src/main/java/tool/arenasizer/CapacityPanel.java
sgcr: Review Finding 3: Add sum sign to translated texts

src/main/java/tool/arenasizer/ArenaInfoPanel.java
sgcr: Review Finding 4: Move 'nextDay' to 'HoDateTime.nextLocalDay'

src/main/java/hattrickdata/CurrentCapacity.java
sgcr: Refactor: Review Finding 5: Use HODateTime instead of LocalDateTime

src/main/java/hattrickdata/HattrickDataInfo.java
sgcr: Refactor: Review Finding 5: Use HODateTime instead of LocalDateTime

src/main/java/hattrickdata/ExpandedCapacity.java
sgcr: Refactor: Review Finding 5: Use HODateTime instead of LocalDateTime

src/main/java/tool/arenasizer/LabelWithSignedNumber.java
sgcr: Well, every time when you need to show a number (or a percentage) as a label that shall be displayed with a plus or minus sign and additionally green for a positive value an red for a negative value.

sgcr: Review Finding 6: Move LabelWithSignedNumber to package 'core.gui'

src/main/java/core/util/HumanDuration.java
```diff
@@ -0,0 +1,67 @@
...
+            final var unit = getLanguageString("ArenaInfoPanel.hours_abbreviation");
+            strings.add(String.format(DURATION_SUB_FORMAT, hours, unit));
+        }
+        if (minutes != 0) {
+            final var unit = getLanguageString("ArenaInfoPanel.minutes_abbreviation");
+            strings.add(String.format(DURATION_SUB_FORMAT, minutes, unit));
+        }
+        if (seconds != 0) {
+            final var unit = getLanguageString("ArenaInfoPanel.seconds_abbreviation");
+            strings.add(String.format(DURATION_SUB_FORMAT, seconds, unit));
```
wsbrenk, line 59: Suggestion: replace "ArenaInfoPanel" by "Duration"

build.gradle
wsbrenk: I don't dare to decide this here because I'm neither our biggest java nor kotlin expert. However, I like Sébastien's suggestion to port to kotlin, because I can imagine that it could be possible to support androids in the medium term.

wsbrenk (approved):
For me it is fine now - I don't want to decide about lombok or not...

sgcr:
> For me it is fine now - I don't want to decide about lombok or not...

Good :)

See also my comment regarding Java records with the last commit.

The pull requests: #2145 and #2143 should be reviewed and merged first because they are partly included here but separate general things.

I would like to rebase this PR here then.
#2145 is clearly a bug.
#2143 seems to be good idea due to testing.

sgcr:
Can be merged.

PR #2150, opened 2024-09-15
refactor: Standardize utility classes

sgcr:
Standardizing means that the utility class has a private constructor that throws UnsupportedOperationException to make clear that there is never the intension to have an instance of that class.

 - [x] does not require update


tychobrailleur (approved):
Thanks for that!

Issue #2119, opened 2024-08-01
[FEATURE] Save players list sort


tristan-bellosta:
Is there a way to keep players ordered by number?
Everytime I restart HO, I need to sort the table again.

It would be nice if last sort we choose is applied at restart.

PR #2154, opened 2024-10-03
#2119

wsbrenk:

   - fixes issue #2119 (see getRowOrderSettings, setRowOrderSettings)

WIP: Todos
- [x] code cleanup
- [x] documentation
- [x] release notes
- [ ] porting of some classes to kotlin is already prepared
- [x] Bug in player analysis table
- [x] Sort keys of lineup player table is not stored
- [x] Auto lineup column is missing in lineup player table
- [x] Error in match table

Since I want to port the main classes of this request to kotlin I created this draft to make review easier.

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/gui/comp/table/HOTableModel.java
```diff
@@ -43,6 +51,15 @@ public abstract class HOTableModel extends AbstractTableModel {
...
 	protected int instance;
 
+	public TableRowSorter<HOTableModel> getRowSorter() {
+		if ( table != null) return (TableRowSorter<HOTableModel>) table.getRowSorter();
+		if ( fixedColumnsTable != null) return fixedColumnsTable.getTableRowSorter();
+		return null;
+	}
+
+	private JTable table;
+	private FixedColumnsTable fixedColumnsTable;
```
tychobrailleur, line 61: This feels like a kludge to have the two fields here: if I understand correctly, this could be one or the other.

I wonder if this could be solved by getting `initTable(FixedColumnsTable table)` to call `initTable(JTable table)` with the nested table in `FixedColumnsTable`? and then in `getRowSorter` get the sorter from `table` as normal.  Would that work?

src/main/java/module/lineup/LineupPanel.java
```diff
@@ -145,82 +137,50 @@ private LineupRatingAssistantPanel getLineupRatingAssistantPanel() {
...
+//					if (player != null) {
+//						HOMainFrame.instance().selectPlayer(player);
+//					}
+//				}
+//			}
+//		};
+
+//		this.lineupPlayersTable.getSelectionModel().addListSelectionListener(lsl);
+//		this.lineupPlayersTableNameColumn.getSelectionModel().addListSelectionListener(lsl);
+//	}
```
tychobrailleur, line 183: Just remove :)

src/main/java/core/gui/comp/table/FixedColumnsTable.java
```diff
@@ -1,11 +1,12 @@
...
+import core.gui.comp.renderer.HODefaultTableCellRenderer;
+import core.model.HOConfigurationIntParameter;
 
 import javax.swing.*;
 import javax.swing.event.ListSelectionListener;
 import javax.swing.table.TableCellRenderer;
-import java.awt.*;
+import javax.swing.table.TableRowSorter;
 
 public class FixedColumnsTable extends JScrollPane {
```
tychobrailleur, line 11: It is possibly because this is probably not the right inheritance here that we're having the issue above.  In all logic this should be a JTable which “decorates” (in the design pattern sense) a JTable in a scrollpane.  Maybe we'll want to refactor this in the future.

src/main/java/module/lineup/LineupTableModel.java
```diff
@@ -1,139 +1,129 @@
...
+import core.gui.model.PlayerOverviewTableModel;
 import core.gui.model.UserColumnController;
-import core.gui.model.UserColumnFactory;
-import core.model.player.Player;
-import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayList;
-import java.util.List;
 
+public class LineupTableModel extends PlayerOverviewTableModel {
```
tychobrailleur, line 7: Is there any point to this class at all now?  or is it needed to set the id, and therefore save the state?

tychobrailleur (commented):
Wow, a lot of work in there!  it looks like now we have a solid generalization of tables, which is great.  Plenty of commented out code to remove, which is great.  As indicated in the comments, I am not convinced by that `FixedColumnsTable` inheritance (which I know was introduced quite a while back), it seems to cause some contorsions that could be avoided to make things simpler.

src/main/java/core/gui/comp/table/HOTableModel.java
wsbrenk: Thank you for your review - I'm also not happy about this switching of table types. I will check if I could handle it, the way you are suggesting,  making the FixedColumnClass an extension of Jtable.

wsbrenk: FixedColumnsTable is alone now.

src/main/java/core/gui/comp/table/FixedColumnsTable.java
wsbrenk: extends JTable now
The surounding ScrollPane is returned in method getContainerComponent.

src/main/java/module/lineup/LineupPanel.java
wsbrenk: I will so so - just want to double check that i didn't miss any feature

src/main/java/core/gui/comp/table/FixedColumnsTable.java
wsbrenk: It's not as easy as initially thought. There are some methods to override and some it doesn't work at all.

tychobrailleur (approved):
I think I had reviewed most of it last time, and changes look good to me.  Still a lot of commented out code, but I am sure you'll take care of it! ;-)

src/main/java/module/lineup/LineupTableModel.java
wsbrenk: Removed this class

PR #2177, opened 2024-11-04
Porting ho table model to kotlin

wsbrenk:

   - fixes issue #___

My first steps in kotlin programming. 
I'm really excited if this can lead to us being able to support Android in the near future (HO10 or HO11 !?).
I don't think it's so great that we lose the history of the changes on this tour.

 - [x] has been updated
 - [ ] does not require update


tychobrailleur:
> I don't think it's so great that we lose the history of the changes on this tour.

I experimented with this a bit, and one way of approaching this is to `git mv` from java to kt in one commit, and then the next commit converts the content to Kotlin.  I don't recommend it, though, because (a) on certain commits we will have broken builds, and (b) after the conversion, the git history will probably be garbage anyway.

tychobrailleur:
I'll have a look when I get a chance, there is a lot to unpack here :)

src/main/java/core/gui/comp/table/HOTableModel.kt
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    /**
+     * Initialize the table object with data from the model
+     * Todo: Think about making HOTableModel supporting only FixedColumnsTable (JTable==FixedColumnsTable(0 fixed columns))
+     * @param table Table object
+     */
+    fun initTable(table: JTable) {
+        this.table = table
+        if (table !is FixedColumnsTable) {
```
tychobrailleur, line 299: Well, didn't know that was possible, `is`! :D

src/main/java/core/gui/comp/table/HOTableModel.kt
```diff
@@ -0,0 +1,402 @@
...
+     * Get row order from user columns and restore it to the given row sorter
+     *
+     * @param rowSorter Row sorter
+     */
+    private fun getRowOrderSettings(rowSorter: RowSorter<HOTableModel>) {
+        // Restore row order setting
+        val sortKeys = ArrayList<RowSorter.SortKey>()
+        val sortColumns = Arrays.stream(this.columns).filter { i: UserColumn -> i.sortPriority != null }.sorted(
+            Comparator.comparingInt { obj: UserColumn -> obj.getSortPriority() }).toList()
+        if (!sortColumns.isEmpty()) {
```
tychobrailleur, line 357: This can become `sortColumns.isNotEmpty()`

src/main/java/core/gui/comp/table/HOTableModel.kt
```diff
@@ -0,0 +1,402 @@
...
+                changed = true
+            }
+        }
+        return changed
+    }
+
+    companion object {
+        @Serial
+        private val serialVersionUID = -207230110294902139L
+    }
```
tychobrailleur, line 401: I am still not sure whether there is any point to `serialVersionUID`.  My understanding was that this was used in serialization/deserialization, and to ensure that the changes made to a serializable class were compatible so that older version of a serialized instance can still be deserialized (I remember a loooong time this mattered for example in RMI), but I don't believe we make sure changes are compatible, or if serialization is even used here.

src/main/java/core/gui/model/PlayerOverviewTableModel.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Initializes the lineup only
+     */
+    fun reInitData() {
+        val tmpDisplayedColumns = getDisplayedColumns()
+        for (i in players!!.indices) {
+            val currentPlayer = players!![i]
+            for (j in tmpDisplayedColumns.indices) {
+                if (tmpDisplayedColumns[j].id == UserColumnFactory.NAME || tmpDisplayedColumns[j].id == UserColumnFactory.LINEUP || tmpDisplayedColumns[j].id == UserColumnFactory.BEST_POSITION || tmpDisplayedColumns[j].id == UserColumnFactory.SCHUM_RANK_BENCHMARK || tmpDisplayedColumns[j].id == UserColumnFactory.GROUP) {
```
tychobrailleur, line 210: Would you mind breaking up into multiple lines for readability?

tychobrailleur (commented):
When we discussed introducing Kotlin, I remember we had proposed adding unit tests as part of the migration.  Is that still something you're keen on doing?  For Swing components this may be tricky, but for something like `HOConfigurationIntParameter` or `HOConfigurationParameter` this can have some value.

So, how do you find Kotlin? :)

wsbrenk (commented):
Merged my current state (#2184) to this branch and include your suggestions.

src/main/java/core/gui/model/PlayerOverviewTableModel.kt
wsbrenk: done

tychobrailleur (approved):

Issue #2061, opened 2024-04-29
[FEATURE] Youth Tab enhancements


vomme2020:
**Is your feature request related to a problem? Please describe.**
I would like to have some more details on the youth tab.

**Describe the solution you'd like**
I would like to have the player number in the list of attributes.
Would make it possible to sort by this as well.

I would like to have an additional field - prefferably for each match.
This would make it possible to add personal observations and the trainer comments for that player for any given match.

Finally, it would be great if it could be configured, what columns to have under the player player tab.
Could be added to the existing preferenses->columns page with an additional dropdown value
![image](https://github.com/ho-dev/HattrickOrganizer/assets/112803830/58c088ce-b161-4464-ac08-87397d7c1b7e)

hvbun:
I would like to be able to add some sort of categorization to the player attributes
Others than that I agree to this suggestion.

wsbrenk:
![image](https://github.com/user-attachments/assets/22437d60-fc94-423e-b5b2-c9a431b7128b)

wsbrenk:
new columns in youth player overview table: 

![image](https://github.com/user-attachments/assets/76d02497-5f0e-4a3a-8140-bbb2cf7ac87b)

wsbrenk:
Added further columns: 
* injury status
* career goals
* career hattricks
* season goals

Hattricks have fixed the error of missing owner notes in the chpp file.

PR #2187, opened 2024-11-17
#2061

wsbrenk:

   - fixes issue #2061 and some errors in FixedColumnsTable

 - [x] has been updated
 - [ ] does not require update


src/main/java/module/teamAnalyzer/ui/RecapPanelTableModel.java
```diff
@@ -334,6 +334,7 @@ private String formatTacticSkill(TeamLineup lineup) {
 
     public TeamLineup getTeamMatchReport(int minSelectionIndex) {
         if ( teamReport != null){
+            var table = getTable();
```
tychobrailleur, line 337: Is this used at all?

src/main/java/core/gui/comp/table/HOTableModel.java
```diff
@@ -454,4 +455,13 @@ private boolean setRowOrderSettings(RowSorter<HOTableModel> sorter) {
 		}
 		return changed;
 	}
+
+	/**
+	 * Returns the primary table
```
tychobrailleur, line 460: I must admit I don't quite get the intent here.  I understand the model is attached to a list of tables, but I don't follow that idea of a primary table?

tychobrailleur (approved):

src/main/java/core/gui/comp/table/HOTableModel.java
wsbrenk: Until now i found one case, where the same model is used in two different tables. It is in the transfer panel. 
If one of those tables will change the setting (column order, width or row order) it will be used for both tables. If the user will change both table settings of the transfer pane, the first of them is the winner.

All other usages of the HOTableModel will be used only by one single table (i called it the primary table).

An alternative solution to this would be to create two independant models for these transfer tables.

src/main/java/module/teamAnalyzer/ui/RecapPanelTableModel.java
wsbrenk: Thanks a lot for your review!

the table is used in the next line to examine the index in the data model.

PR #2192, opened 2024-12-07
#2151

wsbrenk:

   - fixes issue #2151

 - [x] has been updated
 - [ ] does not require update


src/main/java/module/transfer/XMLParser.java
```diff
@@ -44,14 +44,15 @@ public static List<PlayerTransfer> getAllPlayerTransfers(int playerId) {
 
     	final String xml = MyConnector.instance().getTransfersForPlayer(playerId);
         final List<PlayerTransfer> transferList = new Vector<>();
+        if (xml.isEmpty()) return transferList;
```
sgcr, line 47: I suggest to always use curly braces and never omit them.
I think you get better readability and it is less error prone.

src/main/java/module/transfer/XMLParser.java
```diff
@@ -44,14 +44,15 @@ public static List<PlayerTransfer> getAllPlayerTransfers(int playerId) {
...
 
         final Document doc = XMLManager.parseString(xml);
+        if (doc == null) return transferList;
```
sgcr, line 50: I suggest to always use curly braces and never omit them.
I think you get better readability and it is less error prone.

src/main/java/module/transfer/XMLParser.java
```diff
@@ -60,7 +61,7 @@ public static List<PlayerTransfer> getAllPlayerTransfers(int playerId) {
 
                 // Get the player info
                 final Element playerElement = (Element) container.getElementsByTagName("Player").item(0); //$NON-NLS-1$
-                final int playerid = XMLManager.xmlIntValue(playerElement, "PlayerID"); //$NON-NLS-1$
+                playerId = XMLManager.xmlIntValue(playerElement, "PlayerID"); //$NON-NLS-1$
```
sgcr, line 64: I don't get the benefit from reusing the parameter of the function and overwrite the value.
Can you explain?
But yes, the name of variable should be camel case.

src/main/java/module/transfer/XMLParser.java
sgcr, line 47: I also suggest to use `StringUtils.isEmpty` instead to call `empty()` on the string that might be `null`.

wsbrenk: Changed this - thanks

src/main/java/module/transfer/XMLParser.java
wsbrenk: Changed

src/main/java/module/transfer/XMLParser.java
wsbrenk: Reason for the change was to adjust the spelling. I don't know if reusing the parameter has some disadvantages. If parameter and the return value of the xml parsing differs an error on HT side occured. They should always be identical.

src/main/java/module/transfer/XMLParser.java
wsbrenk: Changed

src/main/java/module/transfer/XMLParser.java
sgcr: Okay. We should refactor this later when the parsing code gets improved - okay from my side.

src/main/java/module/transfer/XMLParser.java
```diff
@@ -44,14 +44,15 @@ public static List<PlayerTransfer> getAllPlayerTransfers(int playerId) {
 
     	final String xml = MyConnector.instance().getTransfersForPlayer(playerId);
```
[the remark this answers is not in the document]
sgcr: Well, the PR fixes the bug but I would like to know what the exact XML is that cannot be parsed.

sgcr (approved):

sgcr (approved):

src/main/java/module/transfer/XMLParser.java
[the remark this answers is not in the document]
wsbrenk: As far as i understood the issue this is a followup of an http request error - maybe because ht was down for a while. Then an empty xml string is returned from the MyConnector function.

sgcr (approved):

Issue #2014, opened 2024-01-20
[FEATURE] Bigger fonts


mathias213:
**Is your feature request related to a problem? Please describe.**
Font size can be increased up to 16 (dots?) which is fine is most cases.

Unfortunately, on very dense display panel (like a laptop using 4K) fonts, even in 16, are still very smalls.

**Describe the solution you'd like**
If we would be allowed to increase fonts until something bigger (let's say 32 to be sure to please everyone), that should do.

**Additional context**
Nothing, HO is still my favorite (and only) tool : )

PR #2211, opened 2024-12-28
#2014  Increase maximum font size option

wsbrenk:

   - fixes issue #2014

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/core/gui/comp/panel/DoubleLabelPanel.java
+++ b/src/main/java/core/gui/comp/panel/DoubleLabelPanel.java
@@ -26,7 +26,7 @@ public class DoubleLabelPanel extends JPanel {
         layout.setVgap(0);
         setLayout(layout);
         // Fix right label width to avoid components moving when values change.
-        var width = 60;
+        var width = UserParameter.instance().fontSize * 5;
         var height = UserParameter.instance().fontSize;
         rightLabel.setMinimumSize(new Dimension(width, height));
         rightLabel.setMaximumSize(new Dimension(width, height));
--- a/src/main/java/module/matches/TeamsRatingPanel.java
+++ b/src/main/java/module/matches/TeamsRatingPanel.java
@@ -5,6 +5,7 @@ import core.gui.comp.CustomProgressBar;
 import core.gui.comp.panel.LazyImagePanel;
 import core.gui.theme.HOColorName;
 import core.gui.theme.ThemeManager;
+import core.model.UserParameter;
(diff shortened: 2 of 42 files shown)


src/main/java/core/gui/comp/panel/DoubleLabelPanel.java
```diff
@@ -26,7 +26,7 @@ private void addLabels() {
         layout.setVgap(0);
         setLayout(layout);
         // Fix right label width to avoid components moving when values change.
-        var width = 60;
+        var width = UserParameter.instance().fontSize * 5;
```
tychobrailleur, line 29: Not sure whether this helps (actually, most definitely not here, but maybe in other classes), but `stringWidth` in `FontMetrics` can help finding the size of a string in a given font. See for example [here](https://github.com/ho-dev/HattrickOrganizer/blob/master/src/main/java/core/icon/TextIcon.java#L33).

src/main/java/module/matches/TeamsRatingPanel.java
```diff
@@ -98,14 +97,15 @@ private void addListeners() {
...
+		var fontSize = UserParameter.instance().fontSize;
 
 		f =  new JLabel("").getFont();
 		f = f.deriveFont(f.getStyle() | Font.BOLD);
 
 		bars = new JProgressBar[14];
 		for (int i = 0; i < 14; i++) {
 			bars[i] = new JProgressBar(0, 100);
-			bars[i].setPreferredSize(new Dimension(200, 20)); //25 if nimbus
+			bars[i].setPreferredSize(new Dimension(20 * fontSize, 2 * fontSize)); //25 if nimbus
```
tychobrailleur, line 108: Am I right in saying the comment at the end of the line is no longer valid/relevant?

tychobrailleur (approved):
Sweet, a lot of good stuff in there!  I have added some comments, but none blocking IMHO.

src/main/java/core/gui/comp/panel/DoubleLabelPanel.java
wsbrenk: Good idea for future refactorings. For time being, i only wanted to replace any constant by font size related values.

src/main/java/module/matches/TeamsRatingPanel.java
wsbrenk: no idea, what is the reason for that comment. i will remove it.

PR #2221, opened 2024-12-30
Fix / 0 when stadium under construction, and existing type is 0.

### LINKED ISSUE #2219: [BUG] Error opening Arena Sizer
Hi,

I'm updating my stadium, this morning (whit the r.725) i can open the arena size module, after I start the upgrade and this night I try to open arena sizer (whit r.727) an appears an error.

I don't know if the problem is the update or the arena sizer when you are upgrading a stadium.

image

The logs don't show  more info, only the stacktrace.

tychobrailleur:
This fixes #2219.  This also renames the fields in the stadium class to English for non-German speakers' understanding.

(per recent discussions, I turned on auto-formatting on IntelliJ, so apologies the PR contains lots of formatting changes as a result...)

   - fixes issue #2219 

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/tool/arenasizer/ArenaInfoPanel.java
+++ b/src/main/java/tool/arenasizer/ArenaInfoPanel.java
@@ -17,125 +17,148 @@ import static org.apache.commons.lang3.StringUtils.EMPTY;
 
 public class ArenaInfoPanel extends JPanel {
 
-    private final CapacityPanel currentCapacityPanel;
-    private final CapacityPanel expandedCapacityPanel;
-    private final CapacityPanel futureCapacityPanel;
+	private final CapacityPanel currentCapacityPanel;
+	private final CapacityPanel expandedCapacityPanel;
+	private final CapacityPanel futureCapacityPanel;
 
-    public ArenaInfoPanel() {
-        setLayout(new FlowLayout());
+	public ArenaInfoPanel() {
+		setLayout(new FlowLayout());
 
-        currentCapacityPanel = new CapacityPanel();
-        add(currentCapacityPanel);
+		currentCapacityPanel = new CapacityPanel();
+		add(currentCapacityPanel);
 
-        expandedCapacityPanel = new CapacityPanel();
-        add(expandedCapacityPanel);
+		expandedCapacityPanel = new CapacityPanel();
+		add(expandedCapacityPanel);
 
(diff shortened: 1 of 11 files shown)


src/main/java/tool/arenasizer/ArenaInfoPanel.java
```diff
@@ -17,125 +17,148 @@
...
+		updateValues();
+
+		Timer timer = new Timer(1000, actionEvent -> updateExpansionFinishedLabel());
+		timer.start();
+	}
+
+	private BigDecimal calculatePercentExpansion(int numberOfSeats, int expansion) {
+		if (numberOfSeats == 0) {
+			return null; // Ideally we'd have a way to represent ∞, maybe future update.
+		} else {
```
tychobrailleur, line 45: This is the heart of the fix, checking that existing seating is 0.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
```diff
@@ -17,125 +17,148 @@
...
+			);
+			expandedCapacityPanel.labelPercentageSeatsUnderRoof.setPercentNumber(
+				calculatePercentExpansion(stadium.getUnderRoofSeating(), stadium.getUnderRoofSeatingUnderConstruction())
+			);
+			expandedCapacityPanel.labelPercentageVipBoxes.setPercentNumber(
+				calculatePercentExpansion(stadium.getVipBox(), stadium.getVipBoxUnderConstruction())
+			);
+			expandedCapacityPanel.labelPercentageTotal.setPercentNumber(
+				calculatePercentExpansion(stadium.getTotalSize(), stadium.getTotalSizeUnderConstruction().orElse(0))
+			);
```
tychobrailleur, line 94: Here simplified greatly the code, and wrapped in a unique `if` statement.

src/main/java/tool/arenasizer/ArenaInfoPanel.java
wsbrenk: Good fix - alternativly we could refer to the current value. then 100% would be the displayed value in this case.

if ( currentValue > 0 ) percentage = expansion / currentValue
else if (expasion !=0) percentage = -100
else percentage = 0

wsbrenk (approved):

Issue #163, opened 2019-07-01
[eFEAT] Improve Lineup Assistant manager


akasolace:
Current algorithm is far from being optimal.
See if it feasible to have a real optimization in acceptable time <5s ?

@brokenelevator  I mean optimization as per current spec. For sure it does not intend to do more than simply maximizing total overall rating.

tychobrailleur:
I still think there is an issue with the most recent change:  I had pre-placed 3 players in midfield (my trainees), but one of the trainees was automatically set by the assistant as a winger as well.

![Screenshot From 2025-01-14 19-48-31](https://github.com/user-attachments/assets/f90eab07-9333-42bb-bcc5-f8883a31d5a2)

PR #2227, opened 2025-01-05
#163 Refactor lineup assistant

wsbrenk:

   - fixes issue #163

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/module/lineup/assistant/LineupAssistant.java
+++ b/src/main/java/module/lineup/assistant/LineupAssistant.java
@@ -8,8 +8,6 @@ import core.model.player.IMatchRoleID;
 import core.model.player.MatchRoleID;
 import core.model.player.Player;
 import core.rating.RatingPredictionModel;
-import org.jetbrains.annotations.NotNull;
-
 import java.util.*;
 import java.util.stream.Collectors;
 
@@ -379,33 +377,6 @@ public class LineupAssistant {
 		return getBestPlayerForPosition(position, considerForm, ignoredInjury, ignoreRedCarded, playersIdealPositionOnly, positions);
 	}
 
-	protected final void doReserveSpielerAufstellen(byte position, boolean mitForm,
-													boolean ignoreVerletzung, boolean ignoreSperre, List<Player> vPlayer,
-													List<MatchLineupPosition> positionen) {
-		MatchRoleID pos;
-		Player player;
-
-		for (int i = 0; (positionen != null) && (vPlayer != null) && (i < positionen.size()); i++) {
-			pos = positionen.get(i);
-
-			// Ignore already assigned positions and non substitute position
-			if ((pos.getPlayerId() > 0) || !IMatchRoleID.aSubstitutesMatchRoleID.contains(pos.getId())) {
-				continue;
-			}
-
-			// nur exacte Pos
-			if (pos.getPosition() == position) {
-				player = getBestPlayerForPosition(position, mitForm, ignoreVerletzung, ignoreSperre,
-						vPlayer, positionen);
-
-				// position besetzen
-				if (player != null) {
-					pos.setPlayerIdIfValidForLineup(player.getPlayerId());
-				}
-			}
-		}
-	}
-
 	protected final void doReserveSpielerAufstellenIdealPos(byte position, boolean mitForm,
 															boolean ignoreVerletzung, boolean ignoreSperre, List<Player> vPlayer,
 															List<MatchLineupPosition> positionen) {
@@ -427,42 +398,6 @@ public class LineupAssistant {
 						ignoreSperre, vPlayer, positionen);
 
 				// position besetzen
-				if (player != null) {
-					pos.setPlayerIdIfValidForLineup(player.getPlayerId());
-				}
-			}
-		}
-	}
-
-	/**
-	 * automatic lineup
-	 *
-	 * @param position:         current position being optimized
-	 * @param mitForm:          whether or not to consider the form
-	 * @param ignoreVerletzung: whether or not to align the injured player
-	 * @param ignoreSperre:     whether or not to align the red-carded player
-	 * @param vPlayer:          current position being optimized
-	 * @param positionen:       list of position to be filled
-	 */
-	protected final void doSpielerAufstellen(byte position, boolean mitForm,
-											 boolean ignoreVerletzung, boolean ignoreSperre, List<Player> vPlayer,
-											 List<MatchLineupPosition> positionen) {
-		Player player;
-		if ((positionen == null) || (vPlayer == null)  ) return;
-
-		for (var pos : positionen) {
-
-			//Ignore already assigned positions and substitutes
-			if ((pos.getPlayerId() > 0) || (pos.getId() >= IMatchRoleID.startReserves)) {
-				continue;
-			}
-
-			// position found => get the best player or that position
-			if (pos.getPosition() == position) {
-				player = getBestPlayerForPosition(position, mitForm, ignoreVerletzung, ignoreSperre,
-						vPlayer, positionen);
-
-				// fill the position
 				if (player != null) {
 					pos.setPlayerIdIfValidForLineup(player.getPlayerId());
 				}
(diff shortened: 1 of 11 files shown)


src/main/java/module/lineup/assistant/LineupAssistant.java
```diff
@@ -53,159 +53,250 @@ public final boolean isPlayerInStartingEleven(int spielerId, Vector<MatchLineupP
...
-	public final void doLineup(List<MatchLineupPosition> lPositions, List<Player> lPlayers,
-							   byte sectorsStrengthPriority, boolean bForm, boolean idealPosFirst, boolean bInjured,
-							   boolean bSuspended, Weather weather) {
+	public final void doLineup(List<MatchLineupPosition> allPositions, List<Player> lPlayers,
+							   byte sectorsStrengthPriority, boolean bForm, boolean idealPosFirst, boolean ignoreInjured,
+							   boolean ignoreSuspended, Weather weather) {
 
-		lPositions = filterPositions(lPositions);
+		var lPositions = allPositions.stream().filter(i -> i.isFieldMatchRoleId() && isSelectedByAssistant(i)).toList();
+		ArrayList<Player> players = new ArrayList<>(lPlayers.stream().filter(i -> (i.getInjuryWeeks() < 1 || ignoreInjured) && (!i.isRedCarded() || ignoreSuspended)).toList());
```
tychobrailleur, line 70:
```
List<Player> players = lPlayers.stream()  ... .toList();
```
should be enough (i.e. without additional `new ArrayList<>`) – then you just need to change the signature of `optimizeLineup` to take a `List<Player> players`.

tychobrailleur: (or replace `toList()` with `.collect(Collectors.toList())` for a mutable list)

src/main/java/module/lineup/assistant/LineupAssistant.java
```diff
@@ -53,159 +53,250 @@ public final boolean isPlayerInStartingEleven(int spielerId, Vector<MatchLineupP
...
+		for (var pos : lineupPositions) {
+			if (pos.getPlayerId() == 0 &&   // there isn't already a player at this position
+					requestedPositions.contains(pos.getPosition())) {
+
+				var behaviours = new ArrayList<>(MatchLineupPosition.getBehaviours(pos.getRoleId()));
+				if (behaviours.isEmpty()) {behaviours.add(IMatchRoleID.NORMAL);}
+				var maxRating = -1.0;
+				Player bestPlayer = null;
+				byte bestBehaviour = 0;
+				for (var behaviour : behaviours) {
```
tychobrailleur, line 216: Wow!!  So this computes the best attitude for the player in the position?  That's pretty cool!

src/main/java/module/lineup/assistant/LineupAssistant.java
```diff
@@ -53,159 +53,250 @@ public final boolean isPlayerInStartingEleven(int spielerId, Vector<MatchLineupP
...
+						}
+					}
+				}
+
+				// fill the position
+				if (bestPlayer != null) {
+					pos.setPlayerIdIfValidForLineup(bestPlayer.getPlayerId());
+					pos.setBehaviour(bestBehaviour);
+					if (!pos.isBackupsMatchRoleId()) {
+						players.remove(bestPlayer);
```
tychobrailleur, line 232: One thing that troubles me, and you're probably better placed here to comment, is that we seem to give precedence to the order in which we find players in the list, which may not be the optimal one (another player further down the list could  be better in an earlier position, and the resulting line-up could overall be better).  Is this a a fair thing to say?  

Obviously exploring multiple orders would be expensive from a resource perspective, but I wonder if that could be optimised somehow.

src/main/java/module/lineup/assistant/LineupAssistant.java
```diff
@@ -53,159 +53,250 @@ public final boolean isPlayerInStartingEleven(int spielerId, Vector<MatchLineupP
...
-		var fieldPlayerPositionOrder = new ArrayList<Byte>();
-		var reservePositionOrder = new ArrayList<Byte>();
-		fieldPlayerPositionOrder.add(IMatchRoleID.KEEPER);
-		reservePositionOrder.add(IMatchRoleID.KEEPER);
-		// nun reihenfolge beachten und unbesetzte füllen
+		var fieldPlayerPositionOrder = new ArrayList<List<Byte>>();
+		var reservePositionOrder = new ArrayList<List<Byte>>();
+		fieldPlayerPositionOrder.add(List.of(IMatchRoleID.KEEPER));
+		reservePositionOrder.add(List.of(IMatchRoleID.KEEPER));
 		switch (sectorsStrengthPriority) {
```
tychobrailleur, line 99: Do you think it'd be possible some day to compute all 6 combinations in parallel, and identify the best approach for all 3 sectors?  Or better even let the user select _a posteriori_ what combination to pick for whatever sector they want maximised.

That would have a significant impact on computation, but also on UI, I think.

src/main/java/module/lineup/assistant/LineupAssistant.java
```diff
@@ -53,159 +53,250 @@ public final boolean isPlayerInStartingEleven(int spielerId, Vector<MatchLineupP
...
-	 * @param bInjured:                whether or not to consider injured player
-	 * @param bSuspended:              whether or not to advanced suspended player
+	 * @param ignoreInjured:                whether or not to consider injured player
+	 * @param ignoreSuspended:              whether or not to advanced suspended player
 	 * @param weather:                 Actual weather
 	 */
-	public final void doLineup(List<MatchLineupPosition> lPositions, List<Player> lPlayers,
-							   byte sectorsStrengthPriority, boolean bForm, boolean idealPosFirst, boolean bInjured,
-							   boolean bSuspended, Weather weather) {
+	public final void doLineup(List<MatchLineupPosition> allPositions, List<Player> lPlayers,
```
tychobrailleur, line 65: Here I am thinking of working myself on a unit test framework that would allow us to test these things.

src/main/java/module/lineup/assistant/LineupAssistant.java
tychobrailleur: Another thing I notice is that the attitude selection maximises the player rating at a given position, but doesn't necessarily maximise the expected sector: I have an example of a winger that has a higher rating as defensive, but when set to normal increases midfield slightly, and decreases defence a fair bit.  This makes total sense, but I wonder this is a behaviour the user would expect?

There could be a separate option in the future that aims to maximise selected sector (maybe based on the choice AW/MF/ST?), but probably not relevant at this point.

src/main/resources/release_notes.md
```diff
@@ -35,6 +35,7 @@
 
 ### Lineup
 
+* Refactoring of lineup assistant (#163)
```
tychobrailleur, line 38: I think it's more than refactoring the assistant, it's adding an additional feature, right?

src/main/java/module/lineup/assistant/LineupAssistant.java
wsbrenk: mutable list is required - i will try your 2nd suggestion - Thanks!

src/main/java/module/lineup/assistant/LineupAssistant.java
wsbrenk: Attitude or Behaviour? What would be the better name for it?

src/main/java/module/lineup/assistant/LineupAssistant.java
wsbrenk: The order of the player list has no impact to the optimized lineup - only the order of the positions/sectors.

The calculated player rating is quite similar to the hatstat calculation. I fear it would be very hard to find further improvements. With the help of the sector preference setting, the desired sector is already preferred.

src/main/java/module/lineup/assistant/LineupAssistant.java
wsbrenk: Add "Auto" Entry to the selection list  - good idea

src/main/java/module/lineup/assistant/LineupAssistant.java
wsbrenk: Arguments form and weather are currently not used.

src/main/resources/release_notes.md
wsbrenk: Changed this.

src/main/java/module/lineup/assistant/LineupAssistant.java
tychobrailleur: > Attitude or Behaviour? What would be the better name for it?

Hattrick calls them “individual orders.” Behaviour is fine!

tychobrailleur (approved):
Thanks a lot for this change, really cool improvement!

tychobrailleur:
Following this change, I see locally some of the substitutes that are benched are not only players that are in the starting 11, but also placed in multiple positions.
Screenshot From 2025-01-07 09-41-41

wsbrenk:
@tychobrailleur can not reproduce this - can it be, that you are working with a not mutable player list?

List<Player> players = lPlayers.stream()  ... .toList();

tychobrailleur:
> @tychobrailleur can not reproduce this - can it be, that you are working with a not mutable player list?
> 
> List players = lPlayers.stream() ... .toList();

I don't believe I made any change to the code, but I'll check later on.

wsbrenk:
now i updated my productive version and saw a similar error - although i only saw duplicates on the backup positions, the lineup was not accepted by hattrick, telling about duplicate player positions - even if i fixed it by clearing the backup positions.

tychobrailleur:
> now i updated my productive version and saw a similar error - although i only saw duplicates on the backup positions, the lineup was not accepted by hattrick, telling about duplicate player positions - even if i fixed it by clearing the backup positions.

I cleaned and re-compiled, and the issue went away -- apologies for the noise, it must have been some cached classes from my experiments.
I do also see duplicates in the backup line, but Hattrick accepts mine:

Screenshot From 2025-01-08 20-19-39

wsbrenk:
yes - today my upload was fine too - maybe a hattrick glitch

PR #2250, opened 2025-03-22
Update dependencies

tychobrailleur:
Only dependency left untouched is install4j, as I am not sure what the impact of upgrading to a major version will be.  I have run the app for a couple of hours, without issues.  Let me know if you spot any funnies.

cf. above

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,13 +22,13 @@ buildscript {
 plugins {
     id "java"
     id "groovy"
-    id "org.jetbrains.kotlin.jvm" version "1.9.21"
+    id "org.jetbrains.kotlin.jvm" version "2.1.20"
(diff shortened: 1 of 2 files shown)


wsbrenk (approved):
Thanks - tested it successfully

@tychobrailleur what do you think about moving the remaining v9 tickets to the next release (v10) and releasing beta state of v9 now?

tychobrailleur:
> Thanks - tested it successfully
> 
> @tychobrailleur what do you think about moving the remaining v9 tickets to the next release (v10) and releasing beta state of v9 now?

Yes, I am ok with that.  Any ticket you'd absolutely want addressed for v9?

Issue #2256, opened 2025-04-05
[BUG] Build of v9.0 beta failed


wsbrenk:
Run gradle/wrapper-validation-action@v2
Error: Gradle Wrapper Validation Failed!
  See https://github.com/gradle/wrapper-validation-action#reporting-failures
✗ Other validation errors:
  Expected to find at least 1 Gradle Wrapper JARs but got only 0

PR #2257, opened 2025-04-05
#2256 Fix gradle github action

wsbrenk:

   - fixes issue #2256

 - [ ] has been updated
 - [x] does not require update


src/main/resources/truststore.jks
tychobrailleur, line 1: This is used by okhttp, which was used by the league promotion prediction feature, so I guess ok to remove.  IT may come back in the future for the purpose of the bug reporting feature, FYI.

tychobrailleur (approved):

Issue #2283, opened 2025-05-15
[FEATURE] Add experience skill to all training module tabs


wsbrenk:
Make development of players' experience skill visable in training module tabs.

wsbrenk:
@tychobrailleur Changes of translations in the english properties file did not change the translations in poeditor. The synchronisation seems only handle the keys of the english properties file.

wsbrenk:
player's training development table isn't refreshed too

wsbrenk:
After downloading new players are not shown in training tables - only after restarting HO.

PR #2341, opened 2025-09-28
#2283 Add experience skill to training tables

wsbrenk:

Training table are refactored usind FixedColumnTable (kotlin) as base type. 

4 table classes are still left for refactoring.

 - [ ] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/module/training/ui/EffectTableModel.java
+++ b/src/main/java/module/training/ui/EffectTableModel.java
@@ -20,6 +20,7 @@ import java.util.List;
  */
 public class EffectTableModel extends HOTableModel {
 
+    private static int nextId = 0;
     /**
      * Creates a new EffectTableModel object.
      *
@@ -29,7 +30,7 @@ public class EffectTableModel extends HOTableModel {
         super(columnModelId, "ls.module.training.effect");
 
         columns = new ArrayList<>(List.of(
-                new TrainingColumn("Season", 60) {
+                new TrainingColumn(nextId++, "Season", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         var htWeek = new HODateTime.HTWeek(entry.getHattrickSeason(), entry.getHattrickWeek());
@@ -37,7 +38,7 @@ public class EffectTableModel extends HOTableModel {
                         return new ColorLabelEntry(HODateTime.toEpochSecond(trainingDate), String.valueOf(trainingDate.toLocaleHTWeek().season), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("Week", 60) {
+                new TrainingColumn(nextId++, "Week", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         var htWeek = new HODateTime.HTWeek(entry.getHattrickSeason(), entry.getHattrickWeek());
@@ -45,37 +46,37 @@ public class EffectTableModel extends HOTableModel {
                         return new ColorLabelEntry(HODateTime.toEpochSecond(trainingDate), String.valueOf(trainingDate.toLocaleHTWeek().week), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("TotalTSI", 60) {
+                new TrainingColumn(nextId++, "TotalTSI", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         return new ColorLabelEntry(entry.getTotalTSI(), String.valueOf(entry.getTotalTSI()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("AverageTSI", 60) {
+                new TrainingColumn(nextId++, "AverageTSI", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         return new ColorLabelEntry(entry.getAverageTSI(), String.valueOf(entry.getAverageTSI()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("ls.player.tsi", 70) {
+                new TrainingColumn(nextId++, "ls.player.tsi", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         return new ColorLabelEntry(entry.getTSIIncrease(), String.format("+%d / %d", entry.getTSIIncrease(), entry.getTSIDecrease()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("DurchschnittForm", 70) {
+                new TrainingColumn(nextId++, "DurchschnittForm", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         return new ColorLabelEntry(entry.getAverageForm(), String.format("%.2f", entry.getAverageForm()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("ls.player.form", 70) {
+                new TrainingColumn(nextId++, "ls.player.form", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         return new ColorLabelEntry(entry.getFormIncrease(), String.format("+%d / %d", entry.getFormIncrease(), entry.getFormDecrease()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("Skillups", 70) {
+                new TrainingColumn(nextId++, "Skillups", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainWeekEffect entry) {
                         var icon  = TrainingLegendPanel.getSkillupTypeIcon(entry.getTrainingType(), entry.getAmountSkillups());
--- a/src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
+++ b/src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
@@ -16,8 +16,12 @@ import javax.swing.*;
  */
 public class TrainingSettingsTableModel extends HOTableModel {
 
-	protected List<TrainingPerWeek> o_TrainingsPerWeek;
-
+    private static final int COLUMN_INDEX_INTENSITY = 4;
+    private static final int COLUMN_INDEX_STAMINA_SHARE = 5;
+    private static final int COLUMN_INDEX_COACH_LEVEL = 6;
+    private static final int COLUMN_INDEX_ASSISTANT_LEVEL = 7;
+    private static int nextId=0;
+    protected List<TrainingPerWeek> o_TrainingsPerWeek;
     protected TrainingModel trainingModel;
 
     /**
@@ -27,7 +31,8 @@ public class TrainingSettingsTableModel extends HOTableModel {
         super(columnModelId, name);
 
         columns = new ArrayList<>(List.of(
-                new TrainingColumn("ls.youth.player.training.date", 150) {
+                // 0
+                new TrainingColumn(nextId++, "ls.youth.player.training.date", 150) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         return new ColorLabelEntry(entry.getTrainingDate(), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
@@ -38,21 +43,24 @@ public class TrainingSettingsTableModel extends HOTableModel {
                         return false;
                     }
                 },
-                new TrainingColumn("Season", 60) {
+                // 1
+                new TrainingColumn(nextId++, "Season", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         var trainingDate = entry.getTrainingDate();
                         return new ColorLabelEntry(HODateTime.toEpochSecond(trainingDate), String.valueOf(trainingDate.toLocaleHTWeek().season), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("Week", 60) {
+                // 2
+                new TrainingColumn(nextId++, "Week", 60) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         var trainingDate = entry.getTrainingDate();
                         return new ColorLabelEntry(HODateTime.toEpochSecond(trainingDate), String.valueOf(trainingDate.toLocaleHTWeek().week), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
                     }
                 },
-                new TrainingColumn("ls.team.trainingtype", 140) {
+                // 3
+                new TrainingColumn(nextId++, "ls.team.trainingtype", 140) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         return new ColorLabelEntry(core.constants.TrainingType.toString(entry.getTrainingType()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
@@ -60,7 +68,8 @@ public class TrainingSettingsTableModel extends HOTableModel {
                     @Override
                     public boolean isEditable() {return true;}
                 },
-                new TrainingColumn("ls.team.trainingintensity", 70) {
+                // 4
+                new TrainingColumn(nextId++, "ls.team.trainingintensity", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         return new ColorLabelEntry(entry.getTrainingIntensity(), String.valueOf(entry.getTrainingIntensity()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
@@ -68,7 +77,8 @@ public class TrainingSettingsTableModel extends HOTableModel {
                     @Override
                     public boolean isEditable() {return true;}
                 },
-                new TrainingColumn("ls.team.staminatrainingshare", 70) {
+                // 5
+                new TrainingColumn(nextId++, "ls.team.staminatrainingshare", 70) {
                     @Override
                     public IHOTableCellEntry getTableEntry(TrainingPerWeek entry) {
                         return new ColorLabelEntry(entry.getStaminaShare(), String.valueOf(entry.getStaminaShare()), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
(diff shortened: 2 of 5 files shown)


src/main/java/core/gui/comp/entry/ColorLabelEntry.java
```diff
@@ -54,9 +61,23 @@ public ColorLabelEntry(double sortIndex, String text, Color foreground, Color ba
         number = sortIndex;
         m_clFGColor = foreground;
         m_clBGColor = background;
+        setBorderWidth(8);
```
tychobrailleur: Should this be configurable?  Also, there should be a constant for this value.

src/main/java/core/gui/comp/table/PlayersTable.kt
```diff
@@ -0,0 +1,87 @@
...
+                    firePropertyChanged(oldSelection, player)
+                }
+            }
+
+            fun getSelectedPlayer() : Player?{
+                return selectedPlayer
+            }
+        }
+
+    var players: MutableList<Player?> = HOVerwaltung.instance().getModel().getCurrentPlayers()
```
tychobrailleur: Thinking about this one, this should belong in the tableModel.  Here a generic HOPlayerTableModel, that returns the list of players, as well as tracks the selected players would make sense I think.

src/main/java/core/gui/comp/table/UserColumn.java
```diff
@@ -146,13 +145,19 @@ public String toString(){
...
 	 */
 	public boolean canBeDisabled(){
 		return true;
 	}
 
+	/**
+	 * Column is not visible (width is reduced to zero).
+	 * @return boolean
+	 */
+	public boolean isHidden(){return false;}
```
tychobrailleur, line 157: So if I understand this correctly, instead of not adding the column, we set its size to 0.  This means that the column number is always the total number of columns.  I believe this is relevant somewhere else, where we no longer worry about displayed column count.

src/main/java/core/training/FutureTrainingManager.java
```diff
@@ -53,7 +59,8 @@ public FuturePlayer previewPlayer(int numberOfWeeks) {
 
 		this.futureSkillChanges = new ArrayList<>();
 
-		for (int i = 0; i < 8; i++) {
+		for (int i = 0; i < SKILL_INDEX.length; i++) {
```
tychobrailleur, line 62: Some unit tests for this logic would be useful.

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
```diff
@@ -119,6 +119,7 @@ open class FixedColumnsTable @JvmOverloads constructor(
             rightScrollBar.addAdjustmentListener(adjustmentListener)
             rightScrollPane.verticalScrollBar.model = fixedScrollPane.verticalScrollBar.model
             val splitPane = JSplitPane(JSplitPane.HORIZONTAL_SPLIT, fixedScrollPane, rightScrollPane)
+            splitPane.dividerSize = 2
             if (width == 0) width = 60
```
tychobrailleur, line 153: Here also could do with configurability & extract constants.

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
tychobrailleur, line 1: I believe you were wondering whether this should be a `JScrollPane` – from a pure object-oriented, I think a JTable is better and makes more sense, as instances of this class would be used as table, and you're more likely to need to ”interact” with the table methods.

src/main/java/core/gui/model/PlayerOverviewTableModel.kt
```diff
@@ -87,6 +87,7 @@ class PlayerOverviewTableModel(id: ColumnModelId, name: String) : HOTableModel(i
         assert(this.columns.size == columns.size)
     }
 
+    // TODO: table column model should control isEditable
```
tychobrailleur, line 90: That makes sense!

src/main/java/module/training/ui/AnalyzerPanel.java
```diff
@@ -354,20 +297,29 @@ private void updateFilterPanel() {
...
 
-	private void selectPlayerFromModel() {
+    private boolean disableListSelectionListener = false;
+
+    private void selectPlayerFromModel() {
+        if (this.disableListSelectionListener) return;
+        this.disableListSelectionListener = true;
 		this.changesTable.clearSelection();
-		Player player = this.model.getActivePlayer();
+		Player player = PlayersTable.Companion.getSelectedPlayer();
```
tychobrailleur, line 306: This should come from the table model.

src/main/java/module/training/ui/EffectPanel.java
```diff
@@ -78,16 +52,20 @@ private void initComponents() {
 		JPanel mainpanel = new ImagePanel();
 		mainpanel.setLayout(new BorderLayout());
 
-		this.effectTable = new JTable();
-		JScrollPane effectPane = new JScrollPane(this.effectTable);
-		effectPane.setOpaque(false);
-		mainpanel.add(effectPane, BorderLayout.CENTER);
-
+		this.effectTable = new FixedColumnsTable(UserColumnController.instance().getTrainingEffectTableModel(), 2);
```
tychobrailleur, line 55: Nice!!!

src/main/java/module/training/ui/EffectTableModel.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Creates a new EffectTableModel object.
+     *
+     * @param columnModelId UserColumnController.ColumnModelId.
+     */
+    public EffectTableModel(UserColumnController.ColumnModelId columnModelId) {
+        super(columnModelId, "training.effect");
+
+        columns = new ArrayList<>(List.of(
+                new TrainingColumn("Season", 60) {
```
tychobrailleur: The magic number should be replaced with meaningful constants

src/main/java/module/training/ui/AnalyzerPanel.java
```diff
@@ -51,20 +41,19 @@ public class AnalyzerPanel extends LazyPanel implements ActionListener {
 	private static final String CMD_CLEAR_ALL = "clearAll";
 	private ButtonModel oldPlayers;
 	private JPanel filterPanel;
-	private JTable changesTable;
+	private FixedColumnsTable changesTable;
```
tychobrailleur, line 44: Ah now all this makes sense! That's cool!

src/main/java/core/gui/comp/table/PlayersTable.kt
```diff
@@ -0,0 +1,87 @@
...
+                    val oldSelection = selectedPlayer
+                    selectedPlayer = player
+                    firePropertyChanged(oldSelection, player)
+                }
+            }
+
+            fun getSelectedPlayer() : Player?{
+                return selectedPlayer
+            }
+        }
```
tychobrailleur, line 41: Those companion object methods bother me, but I think this would be too much work to fix at this point.  We can add a TODO to re-visit.

src/main/java/module/training/ui/PlayerDetailPanel.java
```diff
@@ -73,7 +72,7 @@ protected void update() {
     }
 
     private void addListeners() {
-        this.model.addModelChangeListener(change -> setNeedsRefresh(true));
+        PlayersTable.Companion.addPropertyChangeListener(this);
```
tychobrailleur, line 75: See comment about companion object methods on PlayersTable.

src/main/java/module/training/ui/TrainingPanel.java
```diff
@@ -11,36 +15,32 @@
...
  */
 public class TrainingPanel extends JPanel implements TrainingConstants {
 
-	/** Future trainings table model */
-	private FutureTrainingsTableModel futureTrainingsTableModel;
-	/** Past trainings table model */
-	private PastTrainingsTableModel pastTrainingsTableModel;
-
-	private JTable futureTrainingsTable;
+private TrainingSettingsTable futureTrainingsTable;
```
tychobrailleur: small formatting nit.

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
tychobrailleur, line 1: Screenshot From 2025-10-25 13-51-50

I seem to get this error when changing the order of the columns in the Past Trainings table, and trying to sort on that moved column

src/main/java/module/training/ui/model/FutureTrainingEntry.java
tychobrailleur, line 1: Screenshot From 2025-10-25 13-49-14

Some outdated training types, I believe that comes from `core.constants.TrainingType.ITEMS`

src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
```diff
@@ -0,0 +1,173 @@
...
+        // Column index is from the not fixed table part
+        if ( row >= 0 && row < this.o_TrainingsPerWeek.size() && column >= 3 && column < 8){
+            var entry = this.o_TrainingsPerWeek.get(row);
+            var value = this.getValueAt(row, column);
+            if ( value != null) {
+                if ( value instanceof CBItem cbItem )
+                   entry.setTrainingType(cbItem.getId());
+                else {
+                    switch (column) {
+                        case 4:
```
tychobrailleur: Those hardcoded column indices, would that cause the issue about sorting moved columns?

src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
```diff
@@ -0,0 +1,173 @@
...
+    protected TrainingModel trainingModel;
+
+    /**
+     * Creates a new AbstractTrainingsTableModel object.
+     */
+    public TrainingSettingsTableModel(UserColumnController.ColumnModelId columnModelId, String name) {
+        super(columnModelId, name);
+
+        columns = new ArrayList<>(List.of(
+                new TrainingColumn("ls.youth.player.training.date", 150) {
```
tychobrailleur: Some more values that probably need to be extracted as constants?

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
tychobrailleur, line 1: Screenshot From 2025-10-25 14-11-06

There is a disconcerting behaviour in all the `FixedColumnTable` of the training panel, the scrollbar appears at the bottom of the panel, and when trying to scroll it to the left, it does nothing and bounces back to the end.  I suspect it “things” the fixed column are part of the total width of the right hand side table part

src/main/java/module/training/ui/EffectTableModel.java
tychobrailleur, line 1: Any reason you do not use Kotlin for new classes?

tychobrailleur (commented):
Hi @wsbrenk, I have added some comments to this review, hope this is useful!

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
wsbrenk: Good point - what about making it depending on font size. In my case (font size is 12pt), i would suggest border width = font size * 2 / 3.

src/main/java/core/gui/comp/table/PlayersTable.kt
wsbrenk: Got your point - currently PlayersTable is used in different places using different table models, for instance TrainingProgressTableModel. I have to check, if i could change this.

src/main/java/core/gui/comp/table/UserColumn.java
wsbrenk: Former developers of HO included model information into table columns, for instance the player's id. Those column were hidden. With HOTableModel we now have an option to get rid of those columns.

src/main/java/core/training/FutureTrainingManager.java
wsbrenk: Because of my change or basically when scanning a bunch of skills?

src/main/java/module/training/ui/AnalyzerPanel.java
wsbrenk: See above. The table models used in combination with players table are quiet different and are of course related to players, in some cases it is not a 1 to 1 relation, for instance the ChangesTableModel handle a list of skill changes of all players, so each player appears multiple times in the table models values.

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
wsbrenk: Today i tried to reproduce this - until now without success. What i did is moving one of the integer columns to the place of the training type column and hit those table header to sort it. No error message appears, when i do so.

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
wsbrenk: Scroll bars in the FixedColumnTable were a very tricky thing. Until now i didn't find a way to remove them synchronously when they were no longer necessary.

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
wsbrenk: Really? This is the thickness of the divider of the fixed and the not fixed part of the table. The default value (10?) seems to be much to large.

src/main/java/core/gui/comp/table/PlayersTable.kt
wsbrenk: It is my approach to replace the PlayerSelectionListener. I want all player tables in HO to inherit from this class to synchronize their player selection. Now these are the training tables. In a next step the player overview should be added. Then selecting a player in any table will select the same player in each other player table.

src/main/java/module/training/ui/EffectTableModel.java
wsbrenk: I'm with you to walk towards kotlin step by step - and I'm really learning especially by using intellijs conversion function.

But this class is not new - it is only moved from another package to this one....

src/main/java/module/training/ui/EffectTableModel.java
wsbrenk: Do you mean 60? This is no magic - it is the initial width of the column.

src/main/java/module/training/ui/TrainingPanel.java
wsbrenk: Fixed this ;-)

wsbrenk (commented):
Thanks a lot for your review - Due to your error report i tested again the trainings settings and just found an error, which i want to fix in the afternoon

Now i have to clean a bunch of my beehive frames.

src/main/java/module/training/ui/EffectTableModel.java
tychobrailleur: Ah sorry, didn't realise that, thanks for clarifying!

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
tychobrailleur: Yeah, you're right, probably overkill.

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
tychobrailleur: Actually, this isn't related to moving the columns after all.

The repro is:
- Go into the training tab
- In the Training Intensity column of past trainings, change one of the values to 4
- Sort column Training intensity by clicking on header.
- Error above pops up

wsbrenk: Now I found it - thanks!

wsbrenk: Table needs a refresh after editing - then it works.

Hopefully it is fixed now.

src/main/java/module/training/ui/model/FutureTrainingEntry.java
wsbrenk: I hope this is only a matter of your display strings (translation settings).

src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
wsbrenk: Not here, but in another place i forgot the mapping of model column index and view index

src/main/java/core/gui/comp/entry/ColorLabelEntry.java
tychobrailleur: Yes, it looks fixed now!  Thanks.

src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
wsbrenk: Again, this is only the initial width of this column. Don't think that we should use constants for this.

src/main/java/core/gui/model/PlayerOverviewTableModel.kt
wsbrenk: This should be done, when player overview table model is refactored.
It is too much for this request.

src/main/java/core/gui/comp/table/FixedColumnsTable.kt
wsbrenk: @tychobrailleur Today I learned how to synchronize both horizontal scroll bars of the FixedColumnTable. Hope this will fix the issue you reported.

src/main/java/module/training/ui/model/FutureTrainingEntry.java
wsbrenk:
```
ls.team.trainingtype.setpieces=Set pieces
ls.team.trainingtype.defending=Defending
ls.team.trainingtype.scoring=Scoring
ls.team.trainingtype.crossing=Crossing (Winger)
ls.team.trainingtype.shooting=Shooting
ls.team.trainingtype.shortpasses=Short passes
ls.team.trainingtype.playmaking=Playmaking
ls.team.trainingtype.goalkeeping=Goalkeeping
ls.team.trainingtype.throughpasses=Through passes
ls.team.trainingtype.defensivepositions=Defensive positions
ls.team.trainingtype.wingattacks=Wing attacks
```

These are the english properties. I don't know if they will be overriden by poeditor translations or the other way round.

wsbrenk (commented):
@tychobrailleur  Next item I want to handle is your suggestion concerning the new PlayersTable.
The reason why I introduced this class was to handle player selections in all HO player tables in one common way - replacing all the different approaches in the modules.
Maybe i have to move part of the new PlayersTable to a new HOPlayersTableMdel inheriting from HOTableModel.

src/main/java/module/training/ui/model/FutureTrainingEntry.java
tychobrailleur: I don't think this is just a translation thing.  In `TrainingType`:

```
public final class TrainingType {
	public static final int SET_PIECES 			= 2;
	public static final int DEFENDING 			= 3;
	public static final int SCORING 			= 4;
	public static final int CROSSING_WINGER 	= 5;
	public static final int SHOOTING 			= 6;
	public static final int SHORT_PASSES 		= 7;
	public static final int PLAYMAKING 			= 8;
	public static final int GOALKEEPING 		= 9;
	public static final int THROUGH_PASSES 		= 10;
	public static final int DEF_POSITIONS 		= 11;
    public static final int WING_ATTACKS 		= 12;
```
    
Unless I am mistaken, `SHORT_PASSES` and `THROUGH_PASSES` no longer exist, right?  Do we keep for backward compatibility with older HRF files?  Then what does Passing training correspond to in the dropdown, `SHORT_PASSES` or `THROUGH_PASSES`?
    
(To answer your question, I believe `English` is the basis for poeditor, so it doesn't get overridden)

wsbrenk: Those training types still exists.
SHORT_PASSING is the normal passing for midfielder, winger and scorers)
THROUG_PASSES is passing for midfielder, winger and defenders

Here are the german translations of hattrick itself; 


Image

src/main/resources/language/English.properties
```diff
@@ -413,14 +418,14 @@ ls.team.teamattitude_short.matchoftheseason=MOTS
...
+ls.team.trainingtype.shortpasses=Passing
 ls.team.trainingtype.playmaking=Playmaking
-ls.team.trainingtype.goalkeeping=Goalkeeping
-ls.team.trainingtype.throughpasses=Through passes
-ls.team.trainingtype.defensivepositions=Defensive positions
-ls.team.trainingtype.wingattacks=Wing attacks
+ls.team.trainingtype.goalkeeping=Keeper
+ls.team.trainingtype.throughpasses=Passing (Defenders + Midfielders + Wingers)
+ls.team.trainingtype.defensivepositions=Defending (Defenders, Keepers + All Midfielders)
+ls.team.trainingtype.wingattacks=Winger (Wingers + Attackers)
```
wsbrenk, line 428: @tychobrailleur  This will fix the training type translations if poeditor will override the translations with this changes.

wsbrenk (commented):
Fix training type translations/mapping in english properties

src/main/java/core/gui/comp/table/PlayersTable.kt
wsbrenk: @tychobrailleur  players list is now moved to a new class HOPlayersTableModel. Up to now this class is a very tiny one.

src/main/java/module/training/ui/PlayerDetailPanel.java
wsbrenk: Intention of this companion approach is to get one common place for all players tables to synchronize their player selections.

src/main/java/module/training/ui/model/TrainingSettingsTableModel.java
wsbrenk: Replaced this numbers with constants.

tychobrailleur (approved):

PR #2349, opened 2025-11-02
Add exit confirmation dialog and improve UTF-8 support for Greek tran…

eftihia5:
- Added exit confirmation dialog to prevent accidental closure
- Updated localization strings for the new dialog in all supported languages

 - [ ] has been updated
 - [x] does not require update

### DIFF
--- a/src/main/java/core/model/UserParameter.java
+++ b/src/main/java/core/model/UserParameter.java
@@ -460,6 +460,7 @@ public final class UserParameter extends Configuration {
         map.put("downloadTournamentPlayoffMatches", String.valueOf(downloadTournamentPlayoffMatches));
         map.put("downloadDivisionBattleMatches", String.valueOf(downloadDivisionBattleMatches));
         map.put("showHRFSaveDialog", String.valueOf(showHRFSaveDialog));
+        map.put("confirmOnExit", String.valueOf(confirmOnExit));
         map.put("spieleranalyseVertikal", String.valueOf(spieleranalyseVertikal));
         map.put("statistikAlleBeschriftung", String.valueOf(statistikAlleBeschriftung));
         map.put("statistikAlleErfahrung", String.valueOf(statistikAlleErfahrung));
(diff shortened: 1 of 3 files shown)


src/main/resources/language/English.properties
```diff
@@ -570,6 +570,8 @@ ls.menu.file.database=Database
 ls.menu.file.database.dbuseradministration=User administration
 ls.menu.file.database.databasecleanup=Database cleanup
 ls.menu.file.fullscreen=Full screen
+confirm.exit.title=Confirm Exit
+confirm.exit.message=Do you really want to quit Hattrick Organizer?
```
wsbrenk, line 575: Only this english properties are used to synchronize with poeditor.
All other languages has to be edited later in poeditor.

src/main/java/core/gui/HOMainFrame.java
```diff
@@ -726,8 +726,18 @@ public void windowClosed(WindowEvent windowEvent) {
 			 */
 			@Override
 			public void windowClosing(WindowEvent windowEvent) {
-				HOLogger.instance().info(getClass(), "shutting down HO... ");
-				shutdown();
+				int choice = JOptionPane.showConfirmDialog(
```
wsbrenk, line 739: In the issue, you descibed to handle only the menu option to quit HO.

I fear, that this implementation interrupt also the closing of HO with the windows close cross.
I would not like to click two times to stop the application. If that is intented i would suggest to add a check box in the dialog: "Never ask again", which would disable the confirmation dialog.
The setting of this check box could be edited in the options dialog.

src/main/java/core/model/UserParameter.java
```diff
@@ -85,6 +85,9 @@ public final class UserParameter extends Configuration {
      */
     public boolean showHRFSaveDialog = true;
 
+    // If true, shows a confirmation dialog when exiting the application
+    public boolean confirmOnExit = true;
+
```
wsbrenk, line 90: The check box state is not reloaded, when HO is restartet. I think the initialization of the reloaded value in UserParameter::setValues is missing.

wsbrenk:
@eftihia5 Are you able to merge https://github.com/wsbrenk/HO/tree/fix/ConfirmOnExit-item-listener into this pull request. I hope this could fix the error on reloading the confirmOnExit state.

src/main/java/core/option/GeneralSettingsPanel.java
```diff
@@ -1,6 +1,8 @@
 package core.option;
 
 import core.datatype.CBItem;
+import core.db.DBManager;
+import core.gui.HOMainFrame;
```
wsbrenk, line 5: These imports are not necessary, i think

src/main/resources/language/English.properties
```diff
@@ -567,6 +567,7 @@ ls.menu.file.subskillrecalculation=Subskill recalculation
 ls.menu.file.subskillrecalculation7weeks=Subskill recalculation (7 weeks)
 ls.menu.file.preferences=Preferences
 ls.menu.file.database=Database
+options.misc.currency= Currency
```
wsbrenk, line 570: Praise: Good catch.

eftihia5:
It was my pleasure to help with the project.
Thank you!!!

Issue #2296, opened 2025-07-05
[FEATURE] Team Hall of Fame


datatramp:
as posted in ht HO! forum [post=17633079.51]

Add a Hall of Fame feature from available chhp data
that lists / presents the details from Hattrick team's HoF page.

As a second step add editing HoF for supporters, if possible.

PR #2385, opened 2026-01-11
fix: Use UTF-8 encoding for German umlauts

sgcr:
Fixed a German umlaut form latin-1 to UTF-8 in the translation file for German.
Additionally the file `encodings.xml` from IDEA is stored in the repository prevent using the wrong encoding again accidentally in the future.

- [ ] has been updated
- [x] does not require update

### DIFF
--- a/src/main/resources/language/German.properties
+++ b/src/main/resources/language/German.properties
@@ -2261,7 +2261,7 @@ ls.player.ratingend = Bewertung am Ende des Spiels
 ls.player.lastminutes = Einsatzdauer
 ls.player.lastlineup = Position letzter Einsatz
 ls.player.motherclub.name = Heimatverein
-ls.player.matchescurrentteam = Spiele für aktuellen Verein
+ls.player.matchescurrentteam = Spiele fÃ¼r aktuellen Verein
 ls.about.database.folder = Datenbank-Verzeichnis
 ls.about.logs.folder = Logging-Verzeichnis
 # ArenaInfoPanel
--- a/.editorconfig
+++ b/.editorconfig
@@ -71,3 +71,7 @@ tab_width = 4
 indent_style = space
 indent_size = 4
 tab_width = 4
+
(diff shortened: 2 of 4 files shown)


src/main/resources/language/German.properties
```diff
@@ -2261,7 +2261,7 @@ ls.player.ratingend = Bewertung am Ende des Spiels
 ls.player.lastminutes = Einsatzdauer
 ls.player.lastlineup = Position letzter Einsatz
 ls.player.motherclub.name = Heimatverein
-ls.player.matchescurrentteam = Spiele f�r aktuellen Verein
+ls.player.matchescurrentteam = Spiele für aktuellen Verein
```
wsbrenk, line 2264: I fear this file was created by poeditor. Could you check if the poeditor is responsible?

wsbrenk (approved):

src/main/resources/language/German.properties
sgcr: How do I check that?

The line was committed with e0d8e4370ef8c9ef1fec1e1d3a2ac1c1e2e5d895
that originates from #2296 Hall of fame (#2366).

The other umlauts in POEditor look fine and also this one.
How do we get translations from POEditor into the repository?

How do I identify if it was from POEditor?

sgcr: I've used my API token but I get after the first 200 a 403 in the message.
403 means:
`The resource you are trying to access is restricted for this account. (wrong project ID)`

Do I need any additional rights from the original HO project owner from POEditor to be able to pull translations?

sgcr: Manual export is correct and works from web login.

wsbrenk: You're right - it was me, who created this german entry. Damn

wsbrenk (approved):

PR #2390, opened 2026-01-16
fix: Prevent class cast exception by using AmountOfMoney instead of BigDecimal

sgcr:
There was a class cast exception when scrolling in the `Player Analysis` / `Player Comparison` table to the right (on potential visibility of the `Wage` column).

- [x] does not require update

### DIFF
--- a/src/main/java/module/playeranalysis/skillcompare/PlayerTableModel.java
+++ b/src/main/java/module/playeranalysis/skillcompare/PlayerTableModel.java
@@ -194,7 +194,7 @@ public class PlayerTableModel extends DefaultTableModel{
             data[counter][35] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD);
             data[counter][36] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD_DEF);
             data[counter][37] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD_TOWING);
-            data[counter][38] = tmpPlayer.getWages().toLocale();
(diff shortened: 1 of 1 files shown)


src/main/java/module/playeranalysis/skillcompare/PlayerTableModel.java
```diff
@@ -194,7 +194,7 @@ public class PlayerTableModel extends DefaultTableModel{
             data[counter][35] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD);
             data[counter][36] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD_DEF);
             data[counter][37] = tmpPlayer.getPositionCompareAsString(IMatchRoleID.FORWARD_TOWING);
-            data[counter][38] = tmpPlayer.getWages().toLocale();
+            data[counter][38] = tmpPlayer.getWage();
```
wsbrenk: Is getWage the amount of money in swedish krona or did you create a new getter that returns the locale currency? 
The display should be the locale currency. If getWage does return locale currency i would suggest to rename the method to getLocaleWage of something like this.

wsbrenk (changes requested):

src/main/java/module/playeranalysis/skillcompare/PlayerTableModel.java
sgcr: No, I accidentally fixed the typo and reverted it back but this line.
The fix of the typo comes with another PR.

We use `getWages` for this fix.

PR #2391, opened 2026-01-16
refactor: Rename MyTableCellRenderer to PlayerTableCellRenderer

sgcr:
Rename `MyTableCellRenderer` to `PlayerTableCellRenderer` because it is not my and not your table cell renderer but it's a table cell renderer for players.

- [x] does not require update


src/main/java/module/playeranalysis/skillcompare/PlayerTable.java
```diff
@@ -55,7 +55,7 @@ public Component getTableCellRendererComponent(JTable table, Object value,
 	    {
 	    	int width = 65;
 	    	col = this.getColumnModel().getColumn(qq);
-	    	//col.setCellRenderer(new MyTableCellRenderer(iHoMiniMod));
+	    	//col.setCellRenderer(new PlayerTableCellRenderer(iHoMiniMod));
```
wsbrenk: Should be removed?!

sgcr: Removed.

PR #2392, opened 2026-01-16
chore: Make it possible to use Lombok in Kotlin code

sgcr:
Make it possible to use Lombok in Kotlin code.

- [x] does not require update

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,12 +23,13 @@ plugins {
     id "java"
     id "groovy"
     id "org.jetbrains.kotlin.jvm" version "2.2.0"
+    id "org.jetbrains.kotlin.plugin.lombok" version "2.2.0"
     id "application"
     id "org.kordamp.gradle.markdown" version "2.2.0"
     id "com.github.breadmoirai.github-release" version "2.5.2"
     id "com.install4j.gradle" version "12.0"
     id "org.barfuin.gradle.taskinfo" version "2.2.0"
-    id "io.freefair.lombok" version "8.13"
+    id "io.freefair.lombok" version "8.13.1"
 }
 apply plugin: de.jansauer.poeditor.POEditorPlugin
 


build.gradle
```diff
@@ -23,12 +23,13 @@ plugins {
...
     id "groovy"
     id "org.jetbrains.kotlin.jvm" version "2.2.0"
+    id "org.jetbrains.kotlin.plugin.lombok" version "2.2.0"
     id "application"
     id "org.kordamp.gradle.markdown" version "2.2.0"
     id "com.github.breadmoirai.github-release" version "2.5.2"
     id "com.install4j.gradle" version "12.0"
     id "org.barfuin.gradle.taskinfo" version "2.2.0"
-    id "io.freefair.lombok" version "8.13"
+    id "io.freefair.lombok" version "8.13.1"
```
wsbrenk, line 32: Is that necessary? I'm far from being a Kotlin expert, but from what I've seen so far, Kotlin is very close to Lombok features.

I would prefer to get rid of some of our dependencies here.

Two of them are tagged as vulnerable - I'm looking for replacements.

sgcr: Well, I would say that I am an expert in Java. ;-)

It should be possible to use Lombok-generated methods/members from Kotlin as well, not only the other way around.
That’s the reason I added `org.jetbrains.kotlin.plugin.lombok`.

I do not prefer Kotlin over Java for this project, because I currently don’t see a clear benefit.
I know that Sébastien Le Callonnec likes Kotlin very much.

As you know, we have a limited number of contributors. That number is unlikely to increase if we block the use of Lombok from Kotlin.

For now, I think it’s better if we can live in symbiosis.

Regarding mobile applications in the future: Java doesn’t prevent us from building mobile apps.

tychobrailleur: For the record, I do not prefer Kotlin over Java: one of the goals of the HO project is to migrate to Kotlin, in line with @wsbrenk's goal of porting to Android. This has been agreed a few years ago by the project team, and generally speaking new code should be in Kotlin + have unit tests.

I am also with @wsbrenk, I do not see the benefit of bringing lombok to Kotlin, as Kotlin has its own way of defining accessors.  I am still not clear either what the benefit of lombok is for HO in general: this is a convenience library, not a must-have.  Using lombok or not in the project  is IMHO not what drives contributions to this project.

sgcr: Thanks for the clarification — I was wrong to assume you prefer Kotlin over Java. Sorry about that.

Could you point me to where the decision to migrate to Kotlin was made (issue/PR/thread/meeting notes), and who was involved in that decision?

I have no doubt that “Android” is a valid long-term goal, but I’m currently missing a concrete plan/roadmap for it. Is there something I should read?
From my perspective, migrating to Kotlin alone will not make an Android port happen.

Also, I’m trying to understand what is driving the Kotlin migration in practice: @wsbrenk mentioned 
> I’m far from being a Kotlin expert,

and you wrote

> I do not prefer Kotlin over Java

either - and I also don’t prefer Kotlin over Java for this project.
So what are the concrete benefits we expect, and who is taking ownership of that direction?

Separately, I see several areas where we could improve the project regardless of language choice:
- Unit tests are missing in multiple areas, and some tests appear fragile around translations/i18n.
- Translation keys and consistency are sometimes painful.
- Inconsistent naming (e.g., German class/variable names) increases friction.
- Formatting/style is inconsistent in places.
- functionality is sometimes in the UI but should be separated
...

But back to Lombok: the topic here is not “using Lombok in Kotlin” or encouraging Lombok-style Kotlin code.
It’s about interoperability: enabling Kotlin code to *consume* Lombok-generated members (e.g., getters/setters) from existing Java code.

That’s why I added `org.jetbrains.kotlin.plugin.lombok`: so Kotlin can see Lombok-generated accessors and we don’t create avoidable friction in a mixed Java/Kotlin codebase.
Interoperability — not more and not less.

wsbrenk (approved):

wsbrenk:
I really have to say that I am still not a fan of using Lombok. In my opinion, it doesn't make the code any clearer and currently causes a flood of analyzer warnings in IntelliJ, which makes reviewing quite inconvenient.

wsbrenk:
Warning:(12, 7) Class 'Injury' may use Lombok @Getter

I don't like those analyzer warnings.

Issue #2288, opened 2025-06-08
[BUG] Currency displayed does not reflect in-game settings.


ChefToni:
**Describe the bug**
Although in the game preferences I had selected to use Euro currency, in HO it does not reflect as the currency appears based per Country. In my particular case, it appears as BGN (Bulgaria).

**To Reproduce**
1. Go to 'Statistics -> Finances'
2. Go to 'Transfers -> History'
3. Go to 'Misc'

**Screenshots**

![Image](https://github.com/user-attachments/assets/7902d439-4abe-48f5-90a3-ae5506f59b4c)

![Image](https://github.com/user-attachments/assets/50c171e6-6aef-4934-bc10-a90cb779b941)

![Image](https://github.com/user-attachments/assets/a79323b7-6bc9-40f6-b646-b928d822385f)

**Platform information:**
 - OS: [Windows 10]
 - Version 10 DEV r764

wsbrenk:
Until now there is no chpp api function, that returns the currency setting of the team. Only the currency settings of the leagues or countries are available.

In next HO version we can implement the option to set the user's currency choice in the settings dialog. Since this requires a database model update, we will do it in HO 10.

PR #2397, opened 2026-01-18
fix: Correct average wage calculation of team summary

sgcr:
The average wage of the team is not correct displayed when using the currency Euro with a factor 10 lower than Swedish krona.
Fixes a bug that was introduced with #2288.

- [x] does not require update

### DIFF
--- a/src/main/java/module/playeroverview/TeamSummaryModel.java
+++ b/src/main/java/module/playeroverview/TeamSummaryModel.java
@@ -2,11 +2,20 @@ package module.playeroverview;
 
 import core.model.player.Player;
 import core.util.AmountOfMoney;
+import core.util.MathUtils;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
 
+import java.math.RoundingMode;
 import java.util.List;
 
+@Setter
 public class TeamSummaryModel {
 
+    @ToString
+    @EqualsAndHashCode
     static class TeamStatistics {
         int numPlayers;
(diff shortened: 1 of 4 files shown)


src/main/java/module/playeroverview/TeamSummaryModel.java
```diff
@@ -61,15 +58,19 @@ TeamStatistics getComparisonTeamStatistics() {
...
         stats.averageTsi = players.stream().mapToDouble(Player::getTsi).average().orElse(0.0);
         stats.averageAge = players.stream().mapToDouble(Player::getAlterWithAgeDays).average().orElse(0.0);
-        var average = players.stream().mapToDouble(p->p.getWage().toLocale().doubleValue()).average().orElse(0);
-        stats.averageSalary = new AmountOfMoney((long)average) ;
+        var average = MathUtils.average(players.stream()
+            .map(Player::getWage)
+            .map(AmountOfMoney::getSwedishKrona).toList(), RoundingMode.HALF_UP, 2);
+        stats.averageSalary = average.map(AmountOfMoney::new).orElse(new AmountOfMoney(0));
+        stats.averageSalary = average.map(AmountOfMoney::new).orElse(new AmountOfMoney(0));
+        stats.averageSalary = average.map(AmountOfMoney::new).orElse(new AmountOfMoney(0));
```
wsbrenk: One of these statements should be enough.

sgcr: Yes. I am wondering how this could happen.

PR #2398, opened 2026-01-20
Fix unit tests.

tychobrailleur:
- Restore the db folder in test/resources that was causing `BackupHelperTest` to fail; also ensure the backup folder is created without exception
- `HOCurrencyTest` was pulling the full application including `HOMainFrame`, to retrieve current exchange rate.  This change now allows the exchange rate to be overridden.  It also explicitly sets the locale to Germany, as the test was failing for my own Locale (en_IE), which formats currency as €10.

Cf. above.

 - [ ] has been updated
 - [x] does not require update

### DIFF
--- a/src/main/java/core/util/AmountOfMoney.kt
+++ b/src/main/java/core/util/AmountOfMoney.kt
@@ -6,7 +6,6 @@ import core.model.WorldDetailsManager
 import java.math.BigDecimal
 import java.math.RoundingMode
 import java.text.NumberFormat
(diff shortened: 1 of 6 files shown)


tychobrailleur:
> Thanks Sébastien!
> 
> Are the unit tests by now in a state where we could integrate them into the build pipeline?
> 
> Should we go ahead and do that?

At this point in time, they all pass, so yes, good idea!  Let me look into this.

PR #2399, opened 2026-01-20
Fix CVEs with scribejava and commons-text.

tychobrailleur:
This fixes the High severity CVEs identified in the HO dependencies:
- scribejava, by updating the jackson-core transitive dependency
- commons-text, by upgrading to a more recent compatible version.

This also updates junit-jupiter to a stable version.

As scribejava seems to be no longer maintained, I have been looking into alternatives such as Google Oauth client for
Java (https://developers.google.com/api-client-library/java/google-oauth-java-client/), but right now its own latest version pulls in CVEs in its dependencies.

Cf. above.

 - [x] has been updated
 - [ ] does not require update


wsbrenk (approved):
I've been looking for that for a long time - thank you!

sgcr (approved):

tychobrailleur:
> I've been looking for that for a long time - thank you!

No problem.  I don't think a migration to a different library is warranted at this time, but I'll keep an eye out.

PR #2400, opened 2026-01-21
Run unit tests as part of the build pipeline.

tychobrailleur:

This runs the unit tests as part of the build pipeline, as [discussed](https://github.com/ho-dev/HattrickOrganizer/pull/2398#pullrequestreview-3684610419).  This works fine locally, and fails the build when unit tests fail.  This only triggers on release and master branches, but we could create a GH action workflow that triggers and runs validations for all branches, if needs be.

 - [ ] has been updated
 - [x] does not require update


wsbrenk (approved):
Now we need to increase the code coverage.

PR #2408, opened 2026-01-27
refactor: Separate functions for encryption and decryption to HOEncryption

sgcr:
provided)
Separate functions for encryption and decryption to HOEncryption.

- [x] does not require update


tychobrailleur:
> Otherwise we’ll end up litigating “rewrite in Kotlin” on every small PR, which is not a good use of time given current priorities.

Note that this “litigating” is only caused by the fact that you systematically ignore any comment to do so in the first place.  You continuously dismiss any suggestion that does not meet your preference, ignoring the project context and other contributors' experience.  

Also note that the big kotlin rewrite [had](https://github.com/ho-dev/HattrickOrganizer/tree/feature/kotlin_support) [attempts](https://github.com/ho-dev/HattrickOrganizer/tree/feature/kotlin_support_with_hist), but the direction from @wsbrenk was to first ensure sufficient test coverage before the migration, and then to progressively migrate to preserve git history. 

I still fail to understand the value of this PR (and other ones), other than messing a bit more with git history, but at the very least it should be consistent with the rest of the codebase, and be `HOCrypto` or something like that (which, incidentally, would have fit in [this](https://github.com/ho-dev/HattrickOrganizer/commit/9850ead557e01162e93111937c1c3347f3c9d08a) at some point), because that's what the practice of the project is (please do not raise a PR renaming all the `HO...` classes to `Ho...`).

I do agree that given priorities, we have better things to do than this;  I'll leave you and the project in peace, and do my own thing.

wsbrenk (approved):

Issue #2411, opened 2026-01-30
[BUG]Startup error 10.0.833.0-portable-win-dev-JRE


cheilson:
**Describe the bug**
When I start HO, error appears. Close windows, and the program is very buggy.

**To Reproduce**
1. Start HO
2. See error

**Screenshots - copied text**
java.lang.NoSuchMethodError: 'com.github.weisj.jsvg.SVGDocument com.github.weisj.jsvg.parser.SVGLoader.load(java.net.URL, com.github.weisj.jsvg.parser.ParserProvider)'
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon$SVGDocumentHolder.ensureLoaded(DarkSVGIcon.java:336)
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon$SVGDocumentHolder.access$100(DarkSVGIcon.java:316)
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon.ensureLoaded(DarkSVGIcon.java:131)
	at com.github.weisj.darklaf.properties.icons.ThemedSVGIcon.ensureLoaded(ThemedSVGIcon.java:74)
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon.ensureImageLoaded(DarkSVGIcon.java:181)
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon.paintIcon(DarkSVGIcon.java:200)
	at com.github.weisj.darklaf.properties.icons.DarkSVGIcon.paintIcon(DarkSVGIcon.java:127)
	at java.desktop/javax.swing.plaf.synth.SynthGraphicsUtils.paintIcon(SynthGraphicsUtils.java:478)
	at java.desktop/javax.swing.plaf.synth.SynthGraphicsUtils.paintText(SynthGraphicsUtils.java:406)
	at java.desktop/javax.swing.plaf.synth.SynthLabelUI.paint(SynthLabelUI.java:220)
	at java.desktop/javax.swing.plaf.synth.SynthLabelUI.update(SynthLabelUI.java:180)
	at java.desktop/javax.swing.JComponent.paintComponent(JComponent.java:842)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1119)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:952)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1128)
[...]

PR #2415, opened 2026-02-01
#2411 Fix OptionenDialog.save

wsbrenk:

   - fixes issue #2411

 - [ ] has been updated
 - [x] does not require update

### DIFF
--- a/src/main/java/core/option/OptionenDialog.java
+++ b/src/main/java/core/option/OptionenDialog.java
@@ -10,7 +10,6 @@ import core.module.ModuleManager;
 import core.module.config.ModuleConfig;
 import core.util.AmountOfMoney;
 import core.util.Helper;
-import core.util.StringUtils;
 import core.util.Updater;
 
 import javax.swing.*;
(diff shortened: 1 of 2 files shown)


src/main/java/core/option/OptionenDialog.java
```diff
@@ -148,13 +147,15 @@ private void initComponents() {
 	}
 
 	private void save() {
-		// Store user colors before theme setting in user parameters might change
+		// Store user colours before theme setting in user parameters might change
```
sgcr: `colors` American English not British English.

wsbrenk: I agreed with Sébastien that we want to prefer British English.

src/main/java/core/option/OptionenDialog.java
```diff
@@ -148,13 +147,15 @@ private void initComponents() {
...
 		userColorsPanel.storeChangedColorSettings();
 
-		// Store currency setting
-		if ( AmountOfMoney.Companion.setCurrencyCountry(UserParameter.temp().currencyName) ) {
-			OptionManager.instance().setReInitNeeded();
-		}
+        if (!StringUtils.isEmpty(UserParameter.temp().currencyName)) {
```
sgcr: You should use `org.apache.commons.lang3.isEmpty` because we have it available.
`core.util.StringUtils.isEmpty` seem superfluous.

src/main/java/core/option/OptionenDialog.java
```diff
@@ -148,13 +147,15 @@ private void initComponents() {
...
+            // Store currency setting
+            if (AmountOfMoney.Companion.setCurrencyCountry(UserParameter.temp().currencyName)) {
```
sgcr: I think instead of testing above the _not-emptiness_, the `AmountOfMoney.setCurrencyCountry` function should be able to handle a `null` and should return `false` then.

src/main/java/core/option/OptionenDialog.java
sgcr: Okay. Is that somewhere written down?

Most of the time AE is used in source code. Especially in an API.
Mixing AE in the source code and BE in comments is no benefit and results in extra work.

The name of the product is also `Hattrick Organizer` and not `Hattrick Organiser`.
Also we only provide one English translation and I thought is also AE.

sgcr: @wsbrenk we should really think about BE vs AE.

The fix is very good - it like it! :-)

PR #2419, opened 2026-02-04
docs: Add documentation from use of available CHPP files

sgcr:
I was asking myself what the current state of the application is regarding the available CHPP files.
And here the current list of available files and the currently used version...

- [x] does not require update


sgcr:
@wsbrenk Do you agree with the documentation?
Is it possible to be merged from your side?

Issue #2363, opened 2025-12-20
[FEATURE] Calculate healing


wsbrenk:
From Schum: 

Treatment.

As we all know, a new staff system was introduced in 2014.  
And now we don't have 0 to 10 doctors,  
but one physician of level 1-5, or none at all.

The essence of the matter doesn't change, the principles of treatment remain the same.  
So I will try to repost my 2008 article, making it relevant:

1. Trauma Levels

There are 5 possible trauma levels, which are indicated on the player's page and expressed as a percentage of remaining health.*

Health Level, H
Plaster 90-99.99%
1 80-90%
2 70-80%
3 60-70%
4 50-60%

It is impossible to get an injury of more than +4 in a match, but for players over 40, injuries do not decrease; they increase and eventually reach +9.

*Starting from 17.01.2017, health with a bandage remains the same—from 90 to 99.99%. The evaluation of other degrees of injury in conditional weeks now depends both on the player's age and the level of the hired medic. But the essence does not change—the player's recovery speed has not changed, and it can still be measured and assessed based on the player's TSI.

2. Determining the injury sublevel.  
In the first update after sustaining an injury, a player's TSI decreases, and the more serious the injury, the greater the percentage decrease in TSI.  

The relationship here is direct:  
TSI_inj = TSI_healthy * H, where  
TSI_inj – TSI of the injured player  
TSI_healthy – TSI of the player at 100% health  
H – health coefficient.  

From this,  
H = TSI_inj / TSI_healthy.  
Injury sublevel  
S_inj = (1 – H) * 10

After the first update following an injury (usually on Monday or Thursday morning), we see TSIrec on the player's page. 

TSIact is almost always equal to the player's TSI during the match in which they got injured, but there is one important exception here: 

If an older player got injured in a league match on Sun, then for them at the moment after the first update on Mon

TSIact = TSI_Sun – X_Mon, where
TSI_Sun – the player's TSI at the time of the match on Sun
X_Mon – the change in this player's TSI on Mondays in healthy condition.
Age should correspond to the current one.

For example, on the last Monday before the injury, the player's TSI dropped by 80. At the time of the match in which he got injured, his TSI was TSIvs = 10000, and after the injury TSItr = 6150. For him, TSIzd = 10000 – 80 = 9920. H = 6150 / 9920 = 0.62 = 62%. Str = (1-0.62)*10 = 3.8

[...]

PR #2450, opened 2026-05-31
#2363 Calculate player's recovery

wsbrenk:

   - fixes issue #2363

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+        return (x1Factor * (double) doctorLevel + x0Factor) / (x0Factor + 5 * x1Factor);
+    }
+
+    private double calculateHealthIncrease(Player player, int doctorLevel, HODateTime dateTime) {
+        var regainerFactor = player.getSpecialty() == Specialty.Regainer.getValue() ? 10. / 9. : 1.0;
+        return regainerFactor * calcAgeFactor(player.getAgeAtDate(dateTime).toDouble()) * calcMedicianFactor(doctorLevel);
+    }
+
+    private void calculateRecovery(Player player) {
+
```
sgcr: Please remove superfluous empty line.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+                    break;
+                }
+            }
+            if (this.whenHealthy == null) {
+                nextDailyUpdates = nextDailyUpdates.stream().map(v -> v.plus(7, ChronoUnit.DAYS)).toList();
+            }
+        }
+    }
+
+    private double LoadFormBeforeInjured(Player player) {
```
sgcr: Start function names with lower case `loadFormBeforeInjured`

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public TypeOfEstimate getTypeOfEstimate() {
+        return typeOfEstimate;
+    }
+
+    /**
+     * Date when player gets healthy
+     * Date is null if player is healthy
+     */
+    HODateTime whenHealthy;
```
sgcr: please add private access

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Date when player gets slightly injured
+     * Date is null if player is healthy or slightly injured
+     */
+    HODateTime whenSlightlyInjured;
```
sgcr: please add private access

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * True if the player is an invalid (no recovery possible)
+     */
+    Boolean isInvalid = false;
```
sgcr: please add private access

sgcr: When there is no reason that the boolean can ever be ever `null`, then use `boolean` instead of `Boolean`

src/main/java/core/model/player/Player.java
```diff
@@ -875,6 +875,35 @@ public Player getLatestPlayerInfo() {
...
 
+    private Injury injury = null;
+
+    public Injury.TypeOfEstimate getInjuryTypeOfEstimate() {
+        if (injury == null) {
+            injury = new Injury(this);
+        }
+        return injury.getTypeOfEstimate();
+    }
+    public Boolean isInvalid() {
```
sgcr: Please add a blank line between the functions

src/main/java/core/model/XtraData.java
```diff
@@ -235,7 +236,24 @@ public HODateTime getLatestDailyUpdateDateBeforeTraining() {
...
      * @return List of HODateTime
      */
-    public ArrayList<HODateTime> getDailyUpdates() {return dailyUpdates;}
+    public List<HODateTime> getDailyUpdates() {return dailyUpdates;}
+
+    public List<HODateTime> getDailyUpdatesBetween(HODateTime from, HODateTime to ) {
+        var ret = new ArrayList<HODateTime>();
+        for (HODateTime dailyUpdate : getDailyUpdates()) {
+            while (dailyUpdate.isAfter(from)) {
+                dailyUpdate = dailyUpdate.minus(7, ChronoUnit.DAYS);
```
sgcr: Please add a constant for the `7` instead of using a "magic number".
And and use it in every place where this is meant.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+import core.model.HOVerwaltung;
+import core.util.HODateTime;
+import core.util.HOLogger;
+
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Health calculation based on Schum formula (<a href="https://www82.hattrick.org/Forum/Read.aspx?t=17404127&n=6&v=0&mr=0">...</a>)
+ */
+public class Injury {
```
sgcr: What about providing a test for that class?

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Calculate the age related factor of the healing increase.
+     * Fitting of the parameters released by Schum.
+     *
+     * @param age double: Age of the player [17..]
+     * @return double
+     */
+    private double calcAgeFactor(double age) {
```
sgcr: should be static

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+        return x2Factor * x * x + x1Factor * x + x0Factor;
+    }
+
+    /**
+     * Calculate the doctor related factor of the healing increase.
+     *
+     * @param doctorLevel int [0..5]
+     * @return double
+     */
+    private double calcMedicianFactor(int doctorLevel) {
```
sgcr: should be static

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+        double x1Factor = 0.2124;
+        double x0Factor = 1;
+        return (x1Factor * (double) doctorLevel + x0Factor) / (x0Factor + 5 * x1Factor);
+    }
+
+    private double calculateHealthIncrease(Player player, int doctorLevel, HODateTime dateTime) {
```
sgcr: should be static

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+    }
+
+    private double LoadFormBeforeInjured(Player player) {
+        var date = player.getHrfDate();
+        while (true) {
+            var playerBefore = DBManager.instance().getLatestPlayerDownloadBefore(player.getPlayerId(), date.toDbTimestamp());
+            if (playerBefore != null) {
+                if (playerBefore.getInjuryWeeks() == -1) {
```
sgcr: please provide a constant for `-1` and use it in every place where it is meant

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    Boolean isInvalid = false;
+
+    public enum TypeOfEstimate {
+        REALISTIC_ESTIMATE,
+        OPTIMISTIC_ESTIMATE,
+        PESSIMISTIC_ESTIMATE,
+    }
+
+    TypeOfEstimate typeOfEstimate = TypeOfEstimate.REALISTIC_ESTIMATE;
```
sgcr: please add private access

src/main/java/core/model/player/Player.java
```diff
@@ -875,6 +875,35 @@ public Player getLatestPlayerInfo() {
         return latestPlayerInformation;
     }
 
+    private Injury injury = null;
+
+    public Injury.TypeOfEstimate getInjuryTypeOfEstimate() {
+        if (injury == null) {
+            injury = new Injury(this);
```
sgcr, line 883: Instead of creating the instance of `Injury` in every function where it is used,
there should be a `getInjury()` function that does this job.

src/main/java/core/model/XtraData.java
```diff
@@ -235,7 +236,24 @@ public HODateTime getLatestDailyUpdateDateBeforeTraining() {
      * Get the list of daily updates of the next 7 days
      * @return List of HODateTime
      */
-    public ArrayList<HODateTime> getDailyUpdates() {return dailyUpdates;}
+    public List<HODateTime> getDailyUpdates() {return dailyUpdates;}
+
+    public List<HODateTime> getDailyUpdatesBetween(HODateTime from, HODateTime to ) {
```
sgcr: Please use the standard code-formatter of IntelliJ ;-)

src/main/java/core/model/XtraData.java
```diff
@@ -235,7 +236,24 @@ public HODateTime getLatestDailyUpdateDateBeforeTraining() {
      * Get the list of daily updates of the next 7 days
      * @return List of HODateTime
      */
-    public ArrayList<HODateTime> getDailyUpdates() {return dailyUpdates;}
+    public List<HODateTime> getDailyUpdates() {return dailyUpdates;}
```
sgcr: Please use the standard code-formatter of IntelliJ ;-)

sgcr (changes requested):
Please have a look at my comments.

src/main/java/core/model/XtraData.java
wsbrenk: Question: what is magic in writing Minus(7 days) ?

src/main/java/core/model/player/Injury.java
wsbrenk: Good idea - will be done soon...

wsbrenk:
Thanks - Almost all comments have been incorporated

src/main/java/core/model/XtraData.java
sgcr: I commented on the literal value `7` because it is a so-called “magic number”.

A magic number is a numeric value that appears directly in the source code without explaining its meaning. Even if the value itself is obvious mathematically, the intention behind it may not be immediately clear from the surrounding code.

For example, `7` could mean “days per week”, but it could also represent some business-specific limit, an offset, a retry count, or something else. By extracting it into a named constant, the code becomes more self-explanatory:

```java
private static final int DAYS_PER_WEEK = 7;
```

Using a constant improves readability, makes the intention explicit, and avoids repeating the same unexplained value in multiple places. It also makes future changes easier, because the value only has to be changed in one place if the business rule ever changes.

So the point is not that the number `7` is hard to understand, but that the code should communicate why this value is used in this specific context.

wsbrenk: For reasons i don't understand the ChronoUnit.Weeks is not supported. Would you program .minus(ONE weeks) and define

private static final int ONE = 1;

I don't think so....

sgcr: I think the reason is, that `Instant` is a point in time as defined seconds after or before `1970-01-01T00:00:00Z`.
`ChronoUnit` means the calendar system and that system has summer and winter time depending on the timezone you are living in.

It is not clear by definition when you say "7 days after an `Instant`" if you meant `604.800` seconds after the `Instant`.
Or, if the `Instant` is for the `2026-03-29T00:00:00` (Germany) then you have the UTC+1 (winter time) as the start point.
`604.800` after that `Instant` is very clear,
but 7 days are that Instant could mean `2026-04-05T00:00:00` (summer time UTC+2) or maybe `2026-04-05T01:00:00` (summer time UTC+2).

Anyway, I found it also very interesting that the API of Java here is very "restricted". :-)

I think for your calculations calendar days are meant, because Hattrick is calendar based for all match days and so on.

`HODateTime` uses internal an `Instant` and this might have been a `ZonedDateTime` with a specific `ZoneId`.

But I think that is a major refactoring...

sgcr: > Would you program .minus(ONE weeks) and define
> 
> private static final int ONE = 1;
> 
> I don't think so....

For a `1` I would write `1` for sure.

But sometimes it may be important to read it explicitly and then I would use:

`NumberUtils.INTEGER_ONE` or `NumberUtils.LONG_ONE` from package `org.apache.commons.lang3.math`.
`BigDecimal.ONE` also exists.

We have that constants available in our project.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public HODateTime getWhenHealthy() {
+        return whenHealthy;
+    }
+
+    public HODateTime getWhenSlightlyInjured() {
+        return whenSlightlyInjured;
+    }
+
+    public boolean getIsInvalid() {
```
sgcr: `isInvalid`

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+    public TypeOfEstimate getTypeOfEstimate() {
+        return typeOfEstimate;
+    }
+
+    /**
+     * Date when player gets healthy
+     * Date is null if player is healthy
+     */
+    private HODateTime whenHealthy;
+    /**
```
sgcr, line 213: Provide a blank line.

src/main/java/core/model/player/Player.java
```diff
@@ -875,6 +875,31 @@ public Player getLatestPlayerInfo() {
...
+        }
+        return injury;
+    }
+
+    public Injury.TypeOfEstimate getInjuryTypeOfEstimate() {
+        return getInjury().getTypeOfEstimate();
+    }
+
+    public Boolean isInvalid() {
```
sgcr: `boolean`

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+     * Calculate the doctor related factor of the healing increase.
+     *
+     * @param doctorLevel int [0..5]
+     * @return double
+     */
+    private static double calcMedicianFactor(int doctorLevel) {
+        double x1Factor = 0.2124;
+        double x0Factor = 1;
+        return (x1Factor * (double) doctorLevel + x0Factor) / (x0Factor + 5 * x1Factor);
```
sgcr: `(double)` cast not necessary

sgcr: `5` should be a constant instead of magic number.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+import core.model.HOVerwaltung;
+import core.util.HODateTime;
+import core.util.HOLogger;
+
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Health calculation based on Schum formula (<a href="https://www82.hattrick.org/Forum/Read.aspx?t=17404127&n=6&v=0&mr=0">...</a>)
+ */
+public class Injury {
```
sgcr: Class should be package private if the sole constructor is also package private.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+     * Date is null if player is healthy or slightly injured
+     */
+    private HODateTime whenSlightlyInjured;
+
+    /**
+     * True if the player is an invalid (no recovery possible)
+     */
+    private boolean isInvalid = false;
+
+    public enum TypeOfEstimate {
```
sgcr: When `Injury` is only the calculation and all values are delegated to Getters in `Player` then this enum should be a general public enum outside that implementation class. Maybe `InjuryEstimationType`

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+        REALISTIC_ESTIMATE,
+        OPTIMISTIC_ESTIMATE,
+        PESSIMISTIC_ESTIMATE,
```
sgcr: What about "no injury"?

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private TypeOfEstimate typeOfEstimate = TypeOfEstimate.REALISTIC_ESTIMATE;
```
sgcr: Why that default?

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+        double x0Factor = 1;
+        return (x1Factor * (double) doctorLevel + x0Factor) / (x0Factor + 5 * x1Factor);
+    }
+
+    private static double calculateHealthIncrease(Player player, int doctorLevel, HODateTime dateTime) {
+        var regainerFactor = player.getSpecialty() == Specialty.Regainer.getValue() ? 10. / 9. : 1.0;
+        return regainerFactor * calcAgeFactor(player.getAgeAtDate(dateTime).toDouble()) * calcMedicianFactor(doctorLevel);
+    }
+
+    private void calculateRecovery(Player player) {
```
sgcr, line 83: Method is very long: Could it be separated in parts of different sub functions?

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Date when player gets slightly injured
+     * Date is null if player is healthy or slightly injured
+     */
+    private HODateTime whenSlightlyInjured;
+
+    /**
+     * True if the player is an invalid (no recovery possible)
+     */
+    private boolean isInvalid = false;
```
sgcr: Suggestion: `chronicInvalid` as a name because `invalid` is very general and could be a state of the class instance itself.

src/main/java/core/model/player/Injury.java
wsbrenk: Started now the unit tests for the injury class. A major refactoring of the persistance management is required, 
Please have a look on it. I will add further injury tests.

src/main/java/core/db/FaktorenTable.java
```diff
@@ -40,15 +40,15 @@ void pushFactorsIntoDB(FactorObject fo) {
...
-			for (var factor : factors) {
-				FormulaFactors.instance().setPositionFactor(factor.getPosition(), factor);
-			}
-		} else {
-			// use hardcoded values
-			FormulaFactors.instance().importDefaults();
-		}
+	public List<FactorObject> getFaktorenFromDB() {
+		return load(FactorObject.class);
+//		if (!factors.isEmpty()) {
```
sgcr: Please delete code that is not used anymore.

src/test/java/core/model/player/InjuryTests.kt
```diff
@@ -0,0 +1,27 @@
+package core.model.player
+
+import core.HOModelBuilder
+import core.model.HOVerwaltung
+import core.util.HODateTime
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIsNot
+
+class InjuryTests {
```
sgcr: All tests should be named with the name of the class with the postfix "Test" in singular.

src/test/java/core/model/player/InjuryTests.kt
```diff
@@ -0,0 +1,27 @@
...
+    @Test
+    fun testInjury() {
```
sgcr: That is a test but it should cover the cases that could happen (each at least for one possible result).
So multiple tests or if possible better a parameterized test.

I can offer to write tests for that class if you wish. Then you have a template.

src/test/java/core/model/player/InjuryTests.kt
```diff
@@ -0,0 +1,27 @@
...
+
+        val hoAdmin = HOVerwaltung.instance()
```
sgcr: Is the instance of HOVerwaltung really necessary?

src/test/java/core/model/player/InjuryTests.kt
wsbrenk: this is in progress - I will push the draft, which doesn't run successfully until now.

It is still a lot to do because the injury calculation requires a lot of data from the ho model, which will be provided by the new implemented test persistance manager.

src/test/java/core/model/player/InjuryTests.kt
wsbrenk: i will see...

src/test/java/core/model/player/InjuryTests.kt
sgcr: Oh, sorry. Didn't see that.

src/main/java/core/model/player/Injury.java
wsbrenk: Not needed I think - if not injured, the whenHealthy date is null, meaning that it is realistic that we don't need any recovery.

src/main/java/core/model/player/Injury.java
wsbrenk: If the player isn’t injured, which is hopefully the most likely situation for most players, no calculations are done, which means it’s realistic that no recovery is needed.

src/test/java/core/model/player/InjuryTests.kt
wsbrenk: The injury calculation uses HOVerwaltung.instance.

sgcr: Yes, but this could be injected from Player:

```
    private Injury getInjury() {
        if (injury == null) {
            injury = new Injury(this, getVerein(), nextDailyUpdates());
        }
        return injury;
    }

    private Verein getVerein() {
        var hoModel = HOVerwaltung.instance().getModel();
        var persistenceManager = hoModel.getPersistenceManager();
        return persistenceManager.getVerein(this.getHrfId());
    }

    private List<HODateTime> nextDailyUpdates() {
        var hoModel = HOVerwaltung.instance().getModel();
        return hoModel.getXtraDaten().getDailyUpdates().stream().sorted().toList();
    }
    ```
    
    The benefit is you can test the Injury class without the need of `HOVerwaltung`.

Comment (sgcr) [reply src/test/java/core/model/player/InjuryTests.kt]: loadFormBeforeInjured may be mocked.

File: src/main/java/core/model/player/Injury.java (hunk 25 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java:83]: Done

File: src/main/java/core/model/player/Injury.java (hunk 18 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java:213]: Done

File: src/main/java/core/model/player/Injury.java (hunk 22 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java]: What will be the advantage?

File: src/main/java/core/model/player/Player.java (hunk 19 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Player.java]: done

File: src/main/java/core/model/player/Injury.java (hunk 20 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java]: Done

Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java]: MAX_DOCTOR_LEVEL

File: src/main/java/core/model/player/Injury.java (hunk 21 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java]: Done

File: src/main/java/core/model/player/Injury.java (hunk 26 above)
Comment (wsbrenk) [reply src/main/java/core/model/player/Injury.java]: isSportsInvalid

File: src/main/java/core/db/FaktorenTable.java (hunk 27 above)
Comment (wsbrenk) [reply src/main/java/core/db/FaktorenTable.java]: Done

File: src/test/java/core/model/player/InjuryTests.kt (hunk 28 above)
Comment (wsbrenk) [reply src/test/java/core/model/player/InjuryTests.kt]: Done

File: src/test/java/core/model/player/InjuryTests.kt (hunk 29 above)
Comment (wsbrenk) [reply src/test/java/core/model/player/InjuryTests.kt]: Done

File: src/test/java/core/model/player/InjuryTests.kt (hunk 30 above)
Comment (wsbrenk) [reply src/test/java/core/model/player/InjuryTests.kt]: Since HOVerwaltung is used in many places, this could be the preparation for a lot of further unit tests.
I think mocking the persistance is a good approach.

File: src/main/java/core/model/player/Injury.java [hunk 31]
```diff
@@ -0,0 +1,240 @@
...
+     * @return boolean
+     */
+    public boolean isSportsInvalid() {
+        return isSportsInvalid;
+    }
+
+    /**
+     * Get type of recovery estimate.
+     * @return
+     *  REALISTIC - if date should match the real recovery
```
sgcr: Documentation of `REALISTIC`, `OPTIMISTIC` and `PESSIMISTIC` is here a duplicate.
There should be only one place for that: the enum and its values

src/main/java/core/model/player/Injury.java
sgcr: Decoupling and better design.

Only the date and the enum itself are necessary to make it working in the whole application anywhere. Calculation is done within the `Player` with `Injury`.
Beside that you cannot use the `Injury` class because it is package private.

It makes no sense to import the `Injury` class anywhere currently.
The only reason for that is that someone needs that enum for a reason.

Beside that other enums in HO are in a separate files. That is normal.

The normal case for an inner class enum is an enum in private scope when only used for the implementation of that class.

src/main/java/core/model/player/Injury.java
```diff
@@ -0,0 +1,240 @@
...
+    private HODateTime whenSlightlyInjured;
+
+    /**
+     * True if the player cannot recover from injury anymore
+     */
+    private boolean isSportsInvalid = false;
+
+    /**
+     * Type of recovery estimate.
+     *  REALISTIC - if date should match the real recovery
```
sgcr: The comment should be moved directly to the enum value.
Also for `OPTIMISTIC` and `PESSIMISTIC`.

src/main/java/core/model/player/Injury.java
wsbrenk: Changed it

src/main/java/core/model/player/Injury.java
wsbrenk: removed

src/main/java/core/model/player/Injury.java
wsbrenk: removed

src/main/java/core/model/player/TypeOfRecoveryEstimation.java
```diff
@@ -0,0 +1,13 @@
+package core.model.player;
+
+/**
+ * Type of recovery estimate.
+ * REALISTIC - if date should match the real recovery
```
sgcr: Please move the comment to the enum value itself, so that the java doc is working. Beside that the enum value differs here with name below:
```
/**
  * The estimated recovery date should match the expected actual recovery date.
  */
 REALISTIC_ESTIMATE
```

src/main/java/core/model/player/TypeOfRecoveryEstimation.java
```diff
@@ -0,0 +1,13 @@
...
+ * OPTIMISTIC - if date of real recovery is most likely after the estimated date
```
sgcr: What about?
```
/**
  * The actual recovery is most likely to occur after the estimated recovery date.
  */
 OPTIMISTIC_ESTIMATE,
```

src/main/java/core/model/player/TypeOfRecoveryEstimation.java
```diff
@@ -0,0 +1,13 @@
...
+ * PESSIMISTIC  - if date of real recovery is most likely before the estimated date
```
sgcr:
```
 /**
  * The actual recovery is most likely to occur before the estimated recovery date.
  */
 PESSIMISTIC_ESTIMATE,
```

wsbrenk:
@sgcr two weeks without a response means agreement?

sgcr:
> @sgcr two weeks without a response means agreement?

Nope. I will review that first. Time is limited currently.

src/test/java/core/model/player/InjuryTests.kt
sgcr: Well, the test is now better, but:

**A:** I don't like the dependency on `HOVerwaltung`. This may be abstracted from the `Injury` class what is needed here:
Needed is a function that loads the **form** of the player before injured.
Additionally the current level of doctors is needed from the club and the next daily updates...

**B:** Instead of `@Test` there is the possibility for parameterized tests with `@ParameterizedTest`. That what you want here to make it clear what is tested.
No map needed then to store the expected values and you would have a clear test code.

**C:** Test cases should have a result in `OPTIMISTIC_ESTIMATE` and `PESSIMISTIC_ESTIMATE` and not all the time in `REALISTIC_ESTIMATE`.
So maybe more test cases to demonstrate that it works in general.

I am not sure if **B** and **C** is possible without also changing ''A**.
I think the current implementation should work as expected.

Proposal is that we may refactor that with two additional PR's:
Frist for solving **A**.
Afterwards **B** and then **C**.

What do you think?

sgcr (changes requested):
Please see my comment here:
https://github.com/ho-dev/HattrickOrganizer/pull/2450#discussion_r3875151862

Let me know how to proceed. :-)

wsbrenk:
> Please see my comment here: [#2450 (comment)](https://github.com/ho-dev/HattrickOrganizer/pull/2450#discussion_r3875151862)
> 
> Let me know how to proceed. :-)

My approach is to mock the persistence for the tests. The mock model is set into the HOVerwaltung class.
Actually, I found another spot in AmountOfMoney that was accessing the database instance even before this setting, which opened the connection to the DB. I've now prevented that, so now only the test persistence is being used.

Up to now there are a lot of direct calls do DBManager.instance in our model classes which are candidates for further unit tests. If we replaced those calls to HOVerwaltung.instance().getModel().getPersistenceManager(), they can be used in future unit tests without any change and without a complete test database.

So far, I’m not really familiar with either the Java or Kotlin test frameworks (In real world I'm c++/c# develover). If the tests can be built more elegantly, feel free to take over - I’ll learn it then.  I think it’d be good if we handled this in new tickets.

For now, I've adjusted the test so that all estimate types are covered.

sgcr:
> > Please see my comment here: [#2450 (comment)](https://github.com/ho-dev/HattrickOrganizer/pull/2450#discussion_r3875151862)
> > Let me know how to proceed. :-)
> 
> My approach is to mock the persistence for the tests. The mock model is set into the HOVerwaltung class. Actually, I found another spot in AmountOfMoney that was accessing the database instance even before this setting, which opened the connection to the DB. I've now prevented that, so now only the test persistence is being used.
> 
> Up to now there are a lot of direct calls do DBManager.instance in our model classes which are candidates for further unit tests. If we replaced those calls to HOVerwaltung.instance().getModel().getPersistenceManager(), they can be used in future unit tests without any change and without a complete test database.
> 
> So far, I’m not really familiar with either the Java or Kotlin test frameworks (In real world I'm c++/c# develover). If the tests can be built more elegantly, feel free to take over - I’ll learn it then. I think it’d be good if we handled this in new tickets.
> 
> For now, I've adjusted the test so that all estimate types are covered.

Good to know. :-)

Well, you're in good company - I was a C++ developer for 14 years myself before switching to Java in 2019.

Here is my version of **B** with the parameterized tests with (@ParameterizedTest):
[InjuryTest.zip](https://github.com/user-attachments/files/31609434/InjuryTest.zip)
Uploaded as ZIP and you can replace the current file in that branch and have a look at it how this works.

Since C is covered now as well, my proposal would be to handle A separately in a new ticket at a later point.

wsbrenk:
Changed it!
Your code looks nicely tidy, but as a C++ developer, it's hard for me to see that the persistence mock not only has to be created once, but for every test step (4), and that, despite the tidiness, the code has become 20 lines longer;-)

sgcr:
> Changed it! Your code looks nicely tidy, but as a C++ developer, it's hard for me to see that the persistence mock not only has to be created once, but for every test step (4), and that, despite the tidiness, the code has become 20 lines longer;-)



> Changed it! Your code looks nicely tidy, but as a C++ developer, it's hard for me to see that the persistence mock not only has to be created once, but for every test step (4), and that, despite the tidiness, the code has become 20 lines longer;-)

Yes, that's the price of having four real test executions instead of one test containing a loop. ;-)

`@BeforeEach` is executed once for every parameterized test invocation, so in this case the mock model is indeed created four times.

But that also means every test case starts with a clean model state, so I actually think creating the model four times is preferable here. It keeps the test cases isolated from each other.

The code also got a little longer, but I think the individual test cases are much easier to understand now, and if one of them fails, JUnit shows exactly which case failed instead of just failing somewhere inside the loop.

So: more lines, but less hidden logic. :-)

sgcr (approved):

Issue #2455, opened 2026-06-18
[BUG] OwnerNotes exceed VARCHAR(512) limit due to missing deserialization of <br> newline encoding


cpf-fin:
**Describe the bug**

Downloading team data fails with a `SQLDataException` when a player has owner notes containing line breaks. The error makes HO abort the entire download.

**Root cause**

`HRFStringBuilder.serializeMultiLine()` (line 559) converts each `\n` in `OwnerNotes` to `<br>` when building the intermediate HRF string — necessary because HRF uses newlines as record separators. However, there is **no corresponding deserialization** when the HRF is parsed back: `Player.java` line 460 reads the value as-is, so the DB stores `<br>` tags instead of actual newlines.

This causes two problems:
1. **Length overflow**: each newline expands from 1 char to 4 chars (`<br>`). Notes near Hattrick's limit with many line breaks easily exceed the 512-char DB column.
2. **UI defect**: the OwnerNotes column in the player list shows raw `<br>` tags instead of line breaks.

Additionally, `DBUpdater.java` line 214 runs `tryChangeColumn("OwnerNotes", "VARCHAR(512)")` on every startup, reverting any manual DB fix.

**To Reproduce**
1. Write owner notes with many line breaks for any player on hattrick.org (fill the notes field, use Enter liberally)
2. Open HO and download team data
3. `SQLDataException: data exception: string data, right truncation; size limit: 512 table: SPIELER column: OWNERNOTES`

**Suggested fix**

In `Player.java` line 460, deserialize after reading:
```java
ownerNotes = properties.getProperty("ownernotes", "").replace("<br>", "\n");

And increase the column limit in both SpielerTable.java (.setLength(512)) and DBUpdater.java (tryChangeColumn("OwnerNotes", "VARCHAR(512)")) to at least 2048 as a safety margin.

Platform information:
- OS: Windows 11 amd64 (10.0)
- Version: HO! 9.3

wsbrenk:
@cpf-fin great finding - thanks

PR #2456, opened 2026-06-19
#2455 Fix line feed handling in player's owner notes

wsbrenk:

   - fixes issue #2455 

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/core/gui/model/UserColumnFactory.java
+++ b/src/main/java/core/gui/model/UserColumnFactory.java
@@ -676,6 +676,7 @@ public final class UserColumnFactory {
                         new ColorLabelEntry(diff, ColorLabelEntry.BG_STANDARD));
             }
         };
+
         playerAdditionalArray[8] = new PlayerColumn(430, "ls.player.tsi", 0) {
             @Override
             public IHOTableCellEntry getTableEntry(Player player, Player playerCompare) {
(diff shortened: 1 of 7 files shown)


src/main/java/core/gui/model/UserColumnFactory.java
```diff
@@ -804,33 +805,32 @@ public IHOTableCellEntry getTableEntry(Player player, Player playerCompare) {
         playerAdditionalArray[14] = new PlayerColumn(439, "ls.player.statement", "ls.player.statement", 25) {
             @Override
             public IHOTableCellEntry getTableEntry(Player player, Player playerCompare) {
-                return new ColorLabelEntry(player.getPlayerStatement(), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
+                var ret = new ColorLabelEntry(player.getPlayerStatement(), ColorLabelEntry.FG_STANDARD, ColorLabelEntry.BG_STANDARD, SwingConstants.LEFT);
+                ret.setToolTipText("<html>"+player.getPlayerStatement().replace("\n", "<br>")+"</html>");
```
sgcr:
```
public final class SwingToolTips {

    private SwingToolTips() {
    }

    public static String htmlToolTip(String text) {
        if (text == null || text.isBlank()) {
            return null;
        }

        return "<html>"
                + escapeHtml(text).replaceAll("\\R", "<br>")
                + "</html>";
    }

    private static String escapeHtml(String text) {
        return text
                .replace("&", "&amp;")
                .replace("<", "&lt;")
                .replace(">", "&gt;")
                .replace("\"", "&quot;")
                .replace("'", "&#39;");
    }
}
```

Better create a small utility class like this here and use it in every place where needed.

sgcr (changes requested):

src/main/java/core/gui/model/UserColumnFactory.java
wsbrenk: Good idea - i added it to StringUtils

src/main/java/core/util/StringUtils.java
```diff
@@ -87,4 +84,9 @@ public static String capitalizeWord(String str){
 		return capitalizeWord.toString().trim();
 	}
 
+    public static String stringToHtml(String plainText) {
+        if ( plainText == null || plainText.isEmpty() ) {return null;}
```
sgcr: Please use the standard code formatter. ;-)

sgcr: `org.apache.commons.lang3.StringUtils.isEmpty` is your friend instead of `== null || isEmpty`

src/main/java/core/util/StringUtils.java
```diff
@@ -87,4 +84,9 @@ public static String capitalizeWord(String str){
...
+        String htmlEscaped = StringEscapeUtils.escapeHtml4(plainText);
```
sgcr, line 91: `StringEscapeUtils.escapeHtml4` is very good! :-)

Much better than my solution.
I didnt know if `org.apache.commons.text` is available :-)

src/main/java/module/youth/YouthPlayer.java
```diff
@@ -264,7 +264,7 @@ public String getOwnerNotes() {
     }
 
     public void setOwnerNotes(String ownerNotes) {
-        this.ownerNotes = ownerNotes;
+        this.ownerNotes = ownerNotes.replace("<br>", "\n");
```
sgcr: When `.replace("<br>", "\n")` is the reverse function of `HRFStringBuilder.serializeMultiLine` then the two function should be in the same place as public static and should be used when needed

sgcr (approved):

PR #2457, opened 2026-06-22
[FEATURE] General cache location for stadium image in Area Sizer #2440

sgcr:
- fixes issue #2440

- [x] has been updated

### DIFF
--- a/src/main/java/core/gui/image/CachedUrlImageProvider.java
+++ b/src/main/java/core/gui/image/CachedUrlImageProvider.java
@@ -1,0 +1,143 @@
+package core.gui.image;
+
+import core.util.FileNameUtil;
+import core.util.HOLogger;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.Strings;
+
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+public class CachedUrlImageProvider implements ImageProvider {
+
+    private static final String IMAGE_FORMAT_NAME = "PNG";
+
+    private final Path cacheDirectory;
+    private final String primaryUrl;
+    private final String fallbackUrl;
+
+    public CachedUrlImageProvider(Path cacheDirectory, String primaryUrl, String fallbackUrl) {
+        this.cacheDirectory = cacheDirectory;
+        this.primaryUrl = toUrlWithHttps(primaryUrl);
+        this.fallbackUrl = toUrlWithHttps(fallbackUrl);
+    }
+
+    @Override
+    public Optional<BufferedImage> load() {
+        Optional<BufferedImage> primary = tryLoad(primaryUrl);
+
+        if (primary.isPresent()) {
+            return primary;
+        }
+
+        if (StringUtils.isNotBlank(fallbackUrl)) {
+            return tryLoad(fallbackUrl);
+        }
+
+        return Optional.empty();
+    }
+
+    private Optional<BufferedImage> tryLoad(String urlString) {
+        if (StringUtils.isBlank(urlString)) {
+            return Optional.empty();
+        }
+
+        Path fullPath = cacheDirectory
+            .toAbsolutePath()
+            .resolve(toCacheFilename(urlString));
+
+        return loadFromFile(fullPath)
+            .or(() -> loadFromUrlAndSave(urlString, fullPath));
+    }
+
+    private static Optional<BufferedImage> loadFromUrlAndSave(String urlString, Path fullPath) {
+        Optional<BufferedImage> fromUrl = loadFromUrl(urlString);
+        fromUrl.ifPresent(bufferedImage -> saveToFile(bufferedImage, fullPath));
+        return fromUrl;
+    }
+
+    private static Optional<BufferedImage> loadFromUrl(String urlString) {
+        try {
+            URL url = new URL(urlString);
+            URLConnection con = url.openConnection();
+
+            con.setRequestProperty("User-Agent", "Mozilla/5.0");
+            con.setConnectTimeout(4000);
+            con.setReadTimeout(8000);
+
+            try (InputStream in = con.getInputStream()) {
+                return Optional.ofNullable(ImageIO.read(in));
+            }
+        } catch (IOException e) {
+            HOLogger.instance().debug(CachedUrlImageProvider.class,
+                "Image could not be loaded from URL '%s': %s".formatted(urlString, e.getMessage()));
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<BufferedImage> loadFromFile(Path path) {
+        if (!Files.isRegularFile(path)) {
+            return Optional.empty();
+        }
+
+        try {
+            Optional<BufferedImage> image = Optional.ofNullable(ImageIO.read(path.toFile()));
+
+            if (image.isPresent()) {
+                HOLogger.instance().log(CachedUrlImageProvider.class,
+                    "Image loaded from file '%s'.".formatted(path));
+            }
+
+            return image;
+        } catch (IOException e) {
+            HOLogger.instance().error(CachedUrlImageProvider.class,
+                "Failed to load image from file '%s': %s".formatted(path, e.getMessage()));
+            return Optional.empty();
+        }
+    }
+
+    private static void saveToFile(BufferedImage image, Path path) {
+        try {
+            Path parent = path.getParent();
+            if (parent != null) {
+                Files.createDirectories(parent);
+            }
+
+            boolean written = ImageIO.write(image, IMAGE_FORMAT_NAME, path.toFile());
+
+            if (written) {
+                HOLogger.instance().log(CachedUrlImageProvider.class,
+                    "Saved image (format '%s') to file '%s'.".formatted(IMAGE_FORMAT_NAME, path));
+            } else {
+                HOLogger.instance().warning(CachedUrlImageProvider.class,
+                    "Could not save image (format '%s') to file '%s'.".formatted(IMAGE_FORMAT_NAME, path));
+            }
+        } catch (IOException e) {
+            HOLogger.instance().error(CachedUrlImageProvider.class,
+                "Failed to save image (format '%s') to file '%s': %s"
+                    .formatted(IMAGE_FORMAT_NAME, path, e.getMessage()));
+        }
+    }
+
+    private static String toCacheFilename(String fixedUrl) {
+        return "%s.%s".formatted(
+            FileNameUtil.fileNameFromUrl(fixedUrl),
+            IMAGE_FORMAT_NAME.toLowerCase()
+        );
+    }
+
+    private static String toUrlWithHttps(String url) {
+        if (StringUtils.isBlank(url)) {
+            return url;
+        }
+
+        return Strings.CS.startsWith(url, "//") ? "https:" + url : url;
+    }
+}
--- a/src/main/java/core/gui/image/CachedUrlImageProvider.java
+++ b/src/main/java/core/gui/image/CachedUrlImageProvider.java
@@ -7,6 +7,7 @@ import org.apache.commons.lang3.Strings;
 
 import javax.imageio.ImageIO;
 import java.awt.image.BufferedImage;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
@@ -75,9 +76,12 @@ public class CachedUrlImageProvider implements ImageProvider {
             try (InputStream in = con.getInputStream()) {
                 return Optional.ofNullable(ImageIO.read(in));
             }
+        } catch (FileNotFoundException e) {
+            // Expected if the club has no custom stadium image.
+            return Optional.empty();
         } catch (IOException e) {
             HOLogger.instance().debug(CachedUrlImageProvider.class,
-                "Image could not be loaded from URL '%s': %s".formatted(urlString, e.getMessage()));
+                "Image could not be loaded from URL '%s': %s: %s".formatted(urlString, e.getClass().getSimpleName(), e.getMessage()));
             return Optional.empty();
         }
     }
--- a/src/main/java/core/gui/image/CachedUrlImageProvider.java
+++ b/src/main/java/core/gui/image/CachedUrlImageProvider.java
@@ -1,0 +1,273 @@
+package core.gui.image;
+
+import core.util.FileNameUtil;
+import core.util.HOLogger;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.Strings;
+
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Duration;
+import java.time.Instant;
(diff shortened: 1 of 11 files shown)


wsbrenk:
` [Debug]   CachedUrlImageProvider: Image could not be loaded from URL 'https://res.hattrick.org/arenas/13/124/1239/1238716/custom-220-100.jpg': https://res.hattrick.org/arenas/13/124/1239/1238716/custom-220-100.jpg

 [Debug]   CachedUrlImageProvider: Saved image (format 'PNG') to file 'C:\Users\wolfg\Documents\HO\img\stadiumImages\2f0598f59db03c179b1fa13a88b94ca4bc08ceb2119f20e0914ebf532a3d3d33.png'.
`

2nd start of arena sizer gives this logs: 

` [Debug]   CachedUrlImageProvider: Image could not be loaded from URL 'https://res.hattrick.org/arenas/13/124/1239/1238716/custom-220-100.jpg': https://res.hattrick.org/arenas/13/124/1239/1238716/custom-220-100.jpg

 [Debug]   CachedUrlImageProvider: Image loaded from file 'C:\Users\wolfg\Documents\HO\img\stadiumImages\2f0598f59db03c179b1fa13a88b94ca4bc08ceb2119f20e0914ebf532a3d3d33.png'.
`

These logs confuse me a bit: Why is it being reported that a load isn't working, even though an image is being saved? Why is the download being done again on the second try, even though there's an image in the cache?

sgcr:
> age could not be loaded from UR

Well, there is a primary image URL and a fallback image URL:
Your club has no custom stadium image.
In that case, the primary image URL is therefore pointing to hyperspace (HTTP code 404 for file not found is returned and that is the reason for the "Image could not be loaded from URL" message).

The fallback image URL can be loaded and can be saved to a file and is therefore cached the next time.

When the dialog is opened the second time, the primary image URL is looked up in the cache first and cannot be found.
That's true because the image was not saved due to "404 - file not found".
The message in that case appears again that the image could not be loaded from the primary URL.

Afterwards the fallback URL should be read from the file cache.

Proposal:
In case of `FileNotFoundException` there will be no log call anymore (this should be the fix for that).

But in that case the URL is still read and 404 is returned until you design your custom stadium image.

wsbrenk:
> But in that case the URL is still read and 404 is returned until you design your custom stadium image.

I think, most of the HO users don't have custom stadium images. I would prefer if the cached image file is available, it should be used.

The players' avatars are presented as buttons, which can be used to download the avatars on user request. What do you think about the same solution for the stadium images?

sgcr:
> > But in that case the URL is still read and 404 is returned until you design your custom stadium image.
> 
> I think, most of the HO users don't have custom stadium images. I would prefer if the cached image file is available, it should be used.
> 
> The players' avatars are presented as buttons, which can be used to download the avatars on user request. What do you think about the same solution for the stadium images?

I agree that if a cached image is available, it should be used.
I also agree that repeatedly requesting the primary URL and receiving 404 is not ideal, even though the actual traffic caused by a 404 response should be very small.

If we want to avoid these repeated 404 requests, I think the clean solution would be to cache the missing primary image state as well.
For example, if the primary stadium image URL returns 404, we could store a small marker for that URL and skip the primary request while this marker is valid.

However, I would not cache this state forever.
Otherwise, if a user creates a custom stadium image later, HO would never pick it up unless the user manually refreshes it.

So I would suggest combining both ideas:

* use cached images if available
* cache a 404 result for the primary image as a small marker
* use a fixed TTL for cached images and not-found markers
* after the TTL expires, try to refresh automatically
* if the refresh fails, continue using the existing cached image
* allow manual refresh by clicking the stadium image

This avoids repeated 404 requests, avoids permanently stale cache entries, and should still keep the traffic very low because the image is only requested when the Arena Sizer is opened.

src/main/java/core/gui/theme/ThemeManager.java
```diff
@@ -22,82 +24,84 @@
...
-	private static final Path playerAvatarPath = tempImgPath.resolve("playersAvatar");
-	private static final File playerAvatarDir = new File(String.valueOf(playerAvatarPath));
-	private static final Map<String, Theme> themes = new LinkedHashMap<>();
-	private static final ThemeManager MANAGER = new ThemeManager();
+    private static final Path tempImgPath = Paths.get(UserManager.instance().getDbParentFolder(), "img");
+    private static final Path stadiumImagesPath = tempImgPath.resolve("stadiumImages");
+    private static final Path teamLogoPath = tempImgPath.resolve("clubLogos");
+    private static final Path playerAvatarPath = tempImgPath.resolve("playersAvatar");
+    private static final Map<String, Theme> themes = new LinkedHashMap<>();
+    private static final ThemeManager MANAGER = new ThemeManager();
```
wsbrenk, line 57: why are these lines changed?

wsbrenk (commented):
These changes show a lot of changes which are no changes. 
It makes it difficult to find the relevant changes.

src/main/java/core/gui/theme/ThemeManager.java
sgcr: In general:
Due to tabs that are replaces by spaces when reformatting the code with the standard code formatter of IDEA.

**I've reverted some of these unnecessary changes.**

Beside that you can change your view options with the gear at the top when viewing the changes and there you can activate: "hide white-space".
This solves this problem most of the time.

Let me know in which remaining changes you are interested in and then I will explain them.

wsbrenk: we should try to avoid such unnecessary changes, as they not only make reviewing harder but also make the git history (and git blame) useless, which can sometimes be quite helpful.

src/main/java/core/gui/image/CachedUrlImageProvider.java
```diff
@@ -0,0 +1,247 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Optional;
+
+public class CachedUrlImageProvider implements ImageProvider {
+
+    private static final String IMAGE_FORMAT_NAME = "PNG";
+    private static final String NOT_FOUND_EXTENSION = "notfound";
+
+    private static final Duration CACHE_TTL = Duration.ofHours(24);
```
wsbrenk: Is the intention, that downloaded files are only valid for one single day? In case of the stadium pictures i would expect that those won't be changed for months (in my case: never). 

Since I rarely download multiple times a day, in my case that would mean that the image would be updated every time the page is loaded? Then I don't think it needs the cache.

I would increase the value at least with a factor of 30 or better 100.

src/main/java/core/gui/image/CachedUrlImageProvider.java
```diff
@@ -0,0 +1,247 @@
...
+            Optional<BufferedImage> downloadedImage = loadFromUrl(urlString);
+
+            if (downloadedImage.isPresent()) {
+                saveToFile(downloadedImage.get(), imagePath);
+                deleteIfExists(notFoundPath);
+                return downloadedImage;
+            }
+        } catch (FileNotFoundException e) {
+            // Cache the expected HTTP 404.
+            deleteIfExists(imagePath);
```
wsbrenk, line 128: Why delete the cached file in case of download exceptions?
In case of hattrick errors (or changes of the image file access) the existing cached image could remain in place?

wsbrenk (changes requested):
In my opinion, the TTL feature is over the top.

If you insist, the time span should be increased to the order of months, otherwise the images would have to be updated way too often. I definitely don't want that for the player avatars and the team logos.

src/main/java/core/gui/theme/ThemeManager.java
sgcr: **Yes, you are absolutely right about the impact on reviews, Git history and** `git blame`.

I think the best long-term solution would be to introduce an automatic code formatter for the project that is independent of individual IDE settings.

`.editorconfig` already helps with some basic formatting rules such as whitespace and indentation, but it is not a full source-code formatter.

With a common formatter, formatting would be deterministic for everyone and code reviews would mainly show the actual changes instead of formatting differences. It would also ensure consistent formatting throughout the project.

The particular formatter is less important to me.
For example, Spotless with google-java-format would be one option, but there are certainly others.

The main challenge when introducing a formatter into an existing project is the initial formatting.
Formatting the whole codebase at once creates a large commit and affects `git blame`, while formatting only touched files avoids that big initial change but results in a gradual transition.

If you think introducing a formatter is worth considering, I can make a proposal for a few options and possible migration strategies.

src/main/java/core/gui/image/CachedUrlImageProvider.java
sgcr: Fair point. I agree that 24 hours might be a bit short.

Personally, I would prefer something around 7 days.
The TTL does not cause any periodic background requests; a refresh is only attempted when the Arena Sizer is actually opened and the cached entry is older than the TTL.

Since the Arena Sizer is probably not opened very frequently by most users, even a 7-day TTL should result in very little traffic. If someone opens it only once a month, there would still only be one request per month.

That said, I can use 30 days as a compromise. I would consider that the upper end, though, because otherwise automatic invalidation becomes rather infrequent.
I will change the implementation to 30 days accordingly.

With the manual reload by clicking the image, an immediate update is still possible if needed.

Regarding player avatars and team logos, I would leave their current behavior unchanged.
They are outside the scope of this feature, and I haven't looked closely enough at their caching and update mechanisms to suggest changing them as part of this PR.

src/main/java/core/gui/image/CachedUrlImageProvider.java
sgcr: The cached image is only deleted in case of a `FileNotFoundException`, which represents the HTTP 404 case here.

Also, we can only reach this 404 case when an actual download is attempted. This happens when there is no cached image yet, when the cached image has expired, or when a forced reload was requested. A valid cached image is returned without making a request to Hattrick.

In the 404 case, a `.notfound` marker is created and the fallback image is used. As long as that marker is valid, the stale primary image would not be used anyway. Keeping the old image file therefore would not provide any real benefit.

Other `IOException`s are handled differently. In those cases, the cache is not modified at all. The exception is only logged, and if an existing cached image could be loaded, that stale image is returned and remains available.

So the intended behavior is:

* `404 / FileNotFoundException`: remove the old cached image, create a negative cache marker, and use the fallback image
* other `IOException`s: keep the existing cache unchanged and use the stale cached image if available

The distinction is that a 404 explicitly tells us that the resource does not exist, while for other I/O errors we simply cannot determine the current state of the resource.

sgcr:
@wsbrenk Added tests for `FileNameUtil` and `CachedUrlImageProvider`.

src/main/java/core/gui/theme/ThemeManager.java
wsbrenk: A uniform project-specific formatter is fine for me - but it doesn't change the problem if everything gets reformatted across the source, instead of just formatting the lines that are actually changed.

src/main/java/core/gui/image/CachedUrlImageProvider.java
wsbrenk: Wouldn't a 404 response also occur if Hattrick changed the URL or something else regarding access to the image file? Since this isn't a guaranteed CHPP API feature, I wouldn't rely too much on its stability.

sgcr: Here is a summary of how the cache currently works and why the 404 case is handled separately.

The provider uses the complete URL to derive the cache filename.
Therefore, if Hattrick changes the URL, this automatically results in a different cache entry.

The current behavior is:

* If a cached image exists and is still within the TTL, it is returned without accessing Hattrick.
* If the cached image has expired, the URL is accessed again.
* If the download succeeds, the cached image is replaced.
* If the URL returns 404, the cached image is removed and a `.notfound` marker is stored for the same TTL. The reason for caching the 404 is to avoid repeatedly requesting an image that currently does not exist, which is especially relevant for users without a custom stadium image.
* While the `.notfound` marker is valid, no further automatic request is made for that URL and the fallback image is used instead.
* After the `.notfound` marker expires, the URL is tried again automatically.
* A manual reload bypasses both a valid image cache entry and a valid `.notfound` marker, so the current state can always be checked immediately.
* Other `IOException`s are handled differently from a 404. In those cases the existing cached image is not removed. If one is available, it is returned as a stale fallback.

The same mechanism is applied to the fallback URL as well.

The intention is to distinguish between a negative cache result (`404`) and a temporary download/access problem (`IOException`). The latter already preserves the previously cached image.

src/main/java/core/gui/theme/ThemeManager.java
sgcr: I understand your point, especially the intention to keep changes and their history as traceable as possible.
That is definitely a legitimate goal.

Still, if we introduce a common formatter, I would seriously consider doing a one-time formatting commit for the whole codebase.

The main advantage would be that we would have a completely consistent baseline afterwards.
Normal PRs would then mainly contain the actual changes, because the formatter would enforce the same rules for all developers and IDEs.

The obvious downside of such a one-time commit is that it would initially take over a lot of lines in `git blame`.
However, Git provides [`.git-blame-ignore-revs` ](https://git-scm.com/docs/git-blame)specifically for this kind of case.

The pure formatting commit can be added there. `git blame` then tries to skip that commit and attribute the lines to the previous meaningful commit instead. This way, a large part of the usefulness of the existing history can still be preserved.

Of course, this works best if that commit is strictly mechanical: formatting only, no renames, no refactorings and no functional changes.

Therefore, I don't think the trade-off between a consistently formatted codebase and useful `git blame` information is quite as strong as it may initially seem. A possible approach could be:

* agree on a formatter,
* format the complete codebase once,
* put that formatting into a dedicated commit,
* add that commit to `.git-blame-ignore-revs`,
* and then enforce the formatter for all future changes.

The alternative of formatting only the exact lines that are changed certainly preserves the existing history as much as possible. But it also means that the existing inconsistencies may remain in the codebase for a very long time.

In the end, it is a trade-off.
Sometimes you have to consciously accept a disadvantage or give something up in one area in order to gain a larger and more lasting benefit in another.
In this case, for me, that would mean accepting some loss of perfect historical traceability in exchange for a clearly defined and consistent codebase going forward.

In general, I prefer establishing clear concepts and then applying them consistently.
That is part of what I mean by "merciless refactoring": improving things consistently, but in a controlled and traceable way - merciless, **not reckless**.

So I would at least seriously consider the one-time full formatting approach, especially because its biggest downside for `git blame` can be reduced significantly with `.git-blame-ignore-revs`.

wsbrenk: Somehow, I'm hesitant to do such a formatting commit for all the sources. I don't think the current formatting is that bad. I don't see any issue with using the project-specific formatter, which we can happily introduce.

Even with the topic of translating the former German names into English, we dealt with it using the scout strategy, so to do it step by step in the places where changes need to be made anyway.

wsbrenk (approved):

src/main/java/core/gui/theme/ThemeManager.java
sgcr: That sounds like a reasonable compromise to me.

The most important part for me is having a project-specific formatter, so formatting no longer depends on individual IDE settings.

Using the scout strategy for the existing codebase also makes sense.
I would just interpret it a little more broadly than formatting only the exact lines that were changed:
if I work on a method or another reasonably small unit of code, I think it should be fine to leave that unit consistently formatted afterwards.

That way we avoid large unrelated formatting changes, while the codebase can still gradually improve instead of preserving all existing inconsistencies indefinitely.

So basically:
introduce a common formatter for new and changed code, and apply the scout rule to existing code - leave the area you touch a little better than you found it.

I think we should also document this policy in the Coding Guidelines or Codex, together with the formatter to use.
That way it is not just an agreement between us, but a clear project rule that every contributor can follow.

Would you like me to prepare a concrete proposal for the formatter, its configuration, and the corresponding guideline?

Issue #2433, opened 2026-02-17
[FEATURE] Matches: Show the ratings for `Indirect Set Pieces`


sgcr:
**Is your feature request related to a problem? Please describe.**
When you go to the `Matches` to view the details from one of you games you cannot find 
the ratings for `Indirect Set Pieces`.
It cannot be found in `General`, `Ratings`, `Highlights` or `Match report`

**Describe the solution you'd like**
Because it is part of the ratings I would expect to see it there.

**Additional context**
HO web page of match [matchId=758154320](https://www87.hattrick.org/Club/Matches/Match.aspx?matchId=758154320&sourceSystem=Hattrick):
<img width="664" height="345" alt="Image" src="https://github.com/user-attachments/assets/0eb81be3-0024-4ec3-80b3-5999806d662e" />

wsbrenk:
Database change required: 

Table: MatchDetailsTable
RatingIndirectSetPieces* ->RatingIndirectSetPiecesHome*
new RatingIndirectSetPiecesGuest*

Fix XmlMatchDetailsParser::readGuestTeam

PR #2469, opened 2026-08-25
#2433 Add indirect set pieces ratings to match rating panel

wsbrenk:

   - fixes issue #2433

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/file/xml/XMLMatchdetailsParser.java
```diff
@@ -472,9 +472,9 @@ private static void readHomeTeam(Document doc, Matchdetails md) {
 			if ( ele != null ) md.setHomeLeftAtt(Integer.parseInt(ele.getFirstChild().getNodeValue()));
 
 			ele = (Element) root.getElementsByTagName("RatingIndirectSetPiecesAtt").item(0);
-			if ( ele != null ) md.setRatingIndirectSetPiecesAtt(Integer.parseInt(ele.getFirstChild().getNodeValue()));
+			if ( ele != null ) md.setHomeRatingIndirectSetPiecesAtt(Integer.parseInt(ele.getFirstChild().getNodeValue()));
```
sgcr: Please add curly braces.

src/main/java/core/file/xml/XMLMatchdetailsParser.java
```diff
@@ -472,9 +472,9 @@ private static void readHomeTeam(Document doc, Matchdetails md) {
...
 			ele = (Element) root.getElementsByTagName("RatingIndirectSetPiecesDef").item(0);
-			if ( ele != null ) md.setRatingIndirectSetPiecesDef(Integer.parseInt(ele.getFirstChild().getNodeValue()));
+			if ( ele != null ) md.setHomeRatingIndirectSetPiecesDef(Integer.parseInt(ele.getFirstChild().getNodeValue()));
```
sgcr: Please add curly braces.

src/main/java/core/model/match/Matchdetails.java
```diff
@@ -238,22 +240,40 @@ public static Matchdetails getMatchdetails(int matchId, MatchType type) {
...
     }
 
-    public int getRatingIndirectSetPiecesAtt() {
-        return ratingIndirectSetPiecesAtt;
+    public int getHomeRatingIndirectSetPiecesAtt() {
+        return homeRatingIndirectSetPiecesAtt;
+    }
+
+    public void setGuestRatingIndirectSetPiecesAtt(Integer ratingIndirectSetPiecesAtt) {
+        if ( ratingIndirectSetPiecesAtt != null)
```
sgcr: Code-Formatter ;-)

sgcr: Please add curly braces.

src/main/java/core/model/match/Matchdetails.java
```diff
@@ -238,22 +240,40 @@ public static Matchdetails getMatchdetails(int matchId, MatchType type) {
...
+            this.guestRatingIndirectSetPiecesAtt = ratingIndirectSetPiecesAtt;
```
sgcr, line 265: What about the case when we have a `null`.
May be reset to `-1`?

src/main/java/core/model/match/Matchdetails.java
```diff
@@ -238,22 +240,40 @@ public static Matchdetails getMatchdetails(int matchId, MatchType type) {
...
+    }
+
+    public void setGuestRatingIndirectSetPiecesDef(Integer ratingIndirectSetPiecesDef) {
+        if ( ratingIndirectSetPiecesDef != null)
+            this.guestRatingIndirectSetPiecesDef = ratingIndirectSetPiecesDef;
```
sgcr, line 271: What about the case when we have a `null`.
May be reset to `-1`?

src/main/java/core/model/match/Matchdetails.java
```diff
@@ -238,22 +240,40 @@ public static Matchdetails getMatchdetails(int matchId, MatchType type) {
...
+        return homeRatingIndirectSetPiecesAtt;
+    }
+
+    public void setGuestRatingIndirectSetPiecesAtt(Integer ratingIndirectSetPiecesAtt) {
+        if ( ratingIndirectSetPiecesAtt != null)
+            this.guestRatingIndirectSetPiecesAtt = ratingIndirectSetPiecesAtt;
+    }
+
+    public void setGuestRatingIndirectSetPiecesDef(Integer ratingIndirectSetPiecesDef) {
+        if ( ratingIndirectSetPiecesDef != null)
```
sgcr: Please add curly braces and use the code-formatter.

src/main/java/module/matches/TeamsRatingPanel.java
```diff
@@ -65,7 +69,11 @@ private enum RatingSector {
         private final Function<Matchdetails, Integer> versusGuestTeamFunction;
 
         public int getRowNumber() {
-            return getRowIndex() + 1;
+            var ret = getRowIndex() + 1;
+            if (ret > ROW_INDEX_INDIRECT_SET_PIECES){
```
sgcr: code-formatter ;-)

src/main/resources/language/English.properties
```diff
@@ -534,6 +534,9 @@ ls.match.ratingsector.leftdefence=Left defence
 ls.match.ratingsector.rightattack=Right attack
 ls.match.ratingsector.centralattack=Central attack
 ls.match.ratingsector.leftattack=Left attack
+ls.match.ratingsector.indirect.setpieces=Indirect Set pieces
```
sgcr, line 537: Would be nice if you provide for some reason also `German`.

src/main/java/module/teamanalyzer/manager/MatchPopulator.java
```diff
@@ -81,6 +80,8 @@ private MatchRating buildMatchRating(Matchdetails aMatchDetail) {
             mr.setCentralDefense(aMatchDetail.getHomeMidDef());
             mr.setRightAttack(aMatchDetail.getHomeRightAtt());
             mr.setRightDefense(aMatchDetail.getHomeRightDef());
+            mr.setIndirectSetPiecesAtt(aMatchDetail.getHomeRatingIndirectSetPiecesAtt());
```
sgcr, line 83: Is here the possibility of `-1` from the get function?

sgcr (changes requested):

src/main/java/core/model/match/Matchdetails.java
wsbrenk: I had to add this "if" to handle the transition period when the database still has missing entries. Once that's over, a valid entry should never be overwritten by an invalid one.

src/main/java/core/model/match/Matchdetails.java
wsbrenk: see above

src/main/java/module/teamanalyzer/manager/MatchPopulator.java
wsbrenk: Yes, in old database entries of the match details table where those values wasn't loaded from the xml files.

src/main/resources/language/English.properties
wsbrenk: Done

src/main/java/module/matches/TeamsRatingPanel.java
```diff
@@ -56,7 +57,10 @@ private enum RatingSector {
         LEFT_DEFENSE(3, "ls.match.ratingsector.leftdefence", Matchdetails::getHomeLeftDef, Matchdetails::getGuestLeftDef, Matchdetails::getGuestRightAtt),
         RIGHT_ATTACK(4, "ls.match.ratingsector.rightattack", Matchdetails::getHomeRightAtt, Matchdetails::getGuestRightAtt, Matchdetails::getGuestLeftDef),
         CENTRAL_ATTACK(5, "ls.match.ratingsector.centralattack", Matchdetails::getHomeMidAtt, Matchdetails::getGuestMidAtt, Matchdetails::getGuestMidDef),
-        LEFT_ATTACK(6, "ls.match.ratingsector.leftattack", Matchdetails::getHomeLeftAtt, Matchdetails::getGuestLeftAtt, Matchdetails::getGuestRightDef);
+        LEFT_ATTACK(6, "ls.match.ratingsector.leftattack", Matchdetails::getHomeLeftAtt, Matchdetails::getGuestLeftAtt, Matchdetails::getGuestRightDef),
+
+        INDIRECT_SET_PIECES_ATTACK(ROW_INDEX_INDIRECT_SET_PIECES, "ls.match.ratingsector.indirect.setpieces.att", Matchdetails::getHomeRatingIndirectSetPiecesAtt, Matchdetails::getGuestRatingIndirectSetPiecesAtt, Matchdetails::getGuestRatingIndirectSetPiecesDef),
```
sgcr, line 62: Please define `7` here instead of using `ROW_INDEX_INDIRECT_SET_PIECES`.
If `ROW_INDEX_INDIRECT_SET_PIECES` is needed, the initialize it from `RatingSector.INDIRECT_SET_PIECES_ATTACK.getRowIndex()`

src/main/java/module/matches/TeamsRatingPanel.java
```diff
@@ -56,7 +57,10 @@ private enum RatingSector {
...
+        INDIRECT_SET_PIECES_DEFENSE(ROW_INDEX_INDIRECT_SET_PIECES + 1, "ls.match.ratingsector.indirect.setpieces.def", Matchdetails::getHomeRatingIndirectSetPiecesDef, Matchdetails::getGuestRatingIndirectSetPiecesDef, Matchdetails::getGuestRatingIndirectSetPiecesAtt);
```
sgcr, line 63: Define `8` instead of `ROW_INDEX_INDIRECT_SET_PIECES + 1`.

src/main/java/module/matches/TeamsRatingPanel.java
sgcr, line 62: Switch `ROW_INDEX_INDIRECT_SET_PIECES` and `ROW_INDEX_INDIRECT_SET_PIECES` in the enum:
The row numbers and the position in the enum shall be switched because `Defense` comes first in Hattrick.

src/main/java/core/model/match/Matchdetails.java
sgcr: I agree but the default value is `-1` and I think the meaning is that it is not set.
Therefore I think that a `null` should set `-1` instead ignoring a `null`.

src/main/java/core/model/match/Matchdetails.java
sgcr: see above. 😉

sgcr (commented):
Please see my comments.

src/main/java/core/model/match/Matchdetails.java
wsbrenk: The default value in the database is not -1 but null. This is used in the old database entries stored before this column was added to the table.
I didn't want to give them a chance to reset any valid (or still invalid -1).
At the moment, though, I don't really believe that can happen anymore.

Issue #2249, opened 2025-03-21
[BUG] Friendly flags of teams in Hattrick International are handled differently in hattrick and HO


wsbrenk:
Feedback from Morkvomork1 in hattrick forum:

I've had a game against a team from St.Vincent and the Grenadines.
The Team is playing in a Hattrick International League.

In Hattrick i collected the Flag of St.Vincent and the Grenadines,
in HO (Module International Friendly) the "Flag" of Hattrick International.

Bug or Feature?

PR #2473, opened 2026-08-30
#2249 International Friendly Analyzer  statistics regarding team's country ID instead of league ID

wsbrenk:

   - fixes issue #2249

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -153,6 +153,11 @@ final class DBUpdater {
         var hofTable = dbManager.getTable(HallOfFamePlayersTable.TABLENAME);
         hofTable.createTable();
 
+        var ifaMatchTable = dbManager.getTable(IfaMatchTable.TABLENAME);
+        if (ifaMatchTable.tryAddColumn("HOME_COUNTRYID", "INTEGER")){
+            ifaMatchTable.tryAddColumn("AWAY_COUNTRYID", "INTEGER");
+        }
+
 		updateDBVersion(dbVersion, 1000);
 	}
 
--- a/src/main/java/core/db/DBUpdater.java
+++ b/src/main/java/core/db/DBUpdater.java
@@ -99,54 +99,54 @@ final class DBUpdater {
 		}
 	}
 
-	private void updateDBv1000(int dbVersion) throws  SQLException{
+	private void updateDBv1000(int dbVersion) throws  SQLException {
 
         var economyTable = dbManager.getTable(EconomyTable.TABLENAME);
-        for (var column : economyTable.getColumns()){
-            if ( column.getType() == Types.DECIMAL){
+        for (var column : economyTable.getColumns()) {
+            if (column.getType() == Types.DECIMAL) {
                 if (economyTable.columnHasDataType(column.getColumnName(), "DECIMAL")) break; // already done
                 economyTable.tryChangeColumnDataType(column.getColumnName(), "INTEGER", "DECIMAL");
             }
         }
 
-		// Add all available file infos to the world details table, especially the currency information
-		var worldDetailsTable  = dbManager.getTable(WorldDetailsTable.TABLENAME);
-		if ( worldDetailsTable.tryAddColumn("COUNTRY_CODE", "VARCHAR(128)") ) {
-			worldDetailsTable.tryAddColumn("CURRENCY_NAME", "VARCHAR(128)");
-			worldDetailsTable.tryAddColumn("CURRENCY_RATE", "VARCHAR(128)");
-			worldDetailsTable.tryAddColumn("DATE_FORMAT", "VARCHAR(128)");
-			worldDetailsTable.tryAddColumn("TIME_FORMAT", "VARCHAR(128)");
-		}
+        // Add all available file infos to the world details table, especially the currency information
+        var worldDetailsTable = dbManager.getTable(WorldDetailsTable.TABLENAME);
+        if (worldDetailsTable.tryAddColumn("COUNTRY_CODE", "VARCHAR(128)")) {
+            worldDetailsTable.tryAddColumn("CURRENCY_NAME", "VARCHAR(128)");
+            worldDetailsTable.tryAddColumn("CURRENCY_RATE", "VARCHAR(128)");
+            worldDetailsTable.tryAddColumn("DATE_FORMAT", "VARCHAR(128)");
+            worldDetailsTable.tryAddColumn("TIME_FORMAT", "VARCHAR(128)");
+        }
 
-		// Change all amount of money columns to big decimal type
-		var playerTable = dbManager.getTable(SpielerTable.TABLENAME);
-		if ( playerTable.tryChangeColumnDataType("GEHALT", "INTEGER", "DECIMAL") ) {
+        // Change all amount of money columns to big decimal type
+        var playerTable = dbManager.getTable(SpielerTable.TABLENAME);
+        if (playerTable.tryChangeColumnDataType("GEHALT", "INTEGER", "DECIMAL")) {
 
-			var scoutTable = dbManager.getTable(ScoutTable.TABLENAME);
-			scoutTable.tryChangeColumnDataType("PRICE", "INTEGER", "DECIMAL");
-			scoutTable.tryChangeColumnDataType("baseWage", "INTEGER", "DECIMAL");
+            var scoutTable = dbManager.getTable(ScoutTable.TABLENAME);
+            scoutTable.tryChangeColumnDataType("PRICE", "INTEGER", "DECIMAL");
+            scoutTable.tryChangeColumnDataType("baseWage", "INTEGER", "DECIMAL");
 
-			var transferTable = dbManager.getTable(TransferTable.TABLENAME);
-			transferTable.tryChangeColumnDataType("PRICE", "INTEGER", "DECIMAL");
-			transferTable.tryChangeColumnDataType("MOTHERCLUBFEE", "INTEGER", "DECIMAL");
-			transferTable.tryChangeColumnDataType("previousclubcommission", "INTEGER", "DECIMAL");
+            var transferTable = dbManager.getTable(TransferTable.TABLENAME);
+            transferTable.tryChangeColumnDataType("PRICE", "INTEGER", "DECIMAL");
+            transferTable.tryChangeColumnDataType("MOTHERCLUBFEE", "INTEGER", "DECIMAL");
+            transferTable.tryChangeColumnDataType("previousclubcommission", "INTEGER", "DECIMAL");
 
-			var squadInfoTable = dbManager.getTable(SquadInfoTable.TABLENAME);
-			squadInfoTable.tryChangeColumnDataType("SALARY", "INTEGER", "DECIMAL");
+            var squadInfoTable = dbManager.getTable(SquadInfoTable.TABLENAME);
+            squadInfoTable.tryChangeColumnDataType("SALARY", "INTEGER", "DECIMAL");
 
-			var taPlayerTable = dbManager.getTable(TAPlayerTable.TABLENAME);
-			taPlayerTable.tryChangeColumnDataType("SALARY", "INTEGER", "DECIMAL");
+            var taPlayerTable = dbManager.getTable(TAPlayerTable.TABLENAME);
+            taPlayerTable.tryChangeColumnDataType("SALARY", "INTEGER", "DECIMAL");
 
-			var stadiumTable = dbManager.getTable(StadionTable.TABLENAME);
-			stadiumTable.tryChangeColumnDataType("AusbauKosten", "INTEGER", "DECIMAL");
-		}
+            var stadiumTable = dbManager.getTable(StadionTable.TABLENAME);
+            stadiumTable.tryChangeColumnDataType("AusbauKosten", "INTEGER", "DECIMAL");
+        }
 
         // #2254: Career Assists and Assists for the Team
         playerTable.tryAddColumn("CareerAssists", "INTEGER");
         playerTable.tryAddColumn("AssistsCurrentTeam", "INTEGER");
 
-		var stadiumTable = dbManager.getTable(StadionTable.TABLENAME);
-		stadiumTable.tryChangeColumn("AusbauKosten", "NULL");
+        var stadiumTable = dbManager.getTable(StadionTable.TABLENAME);
+        stadiumTable.tryChangeColumn("AusbauKosten", "NULL");
         stadiumTable.tryAddColumn("ARENA_IMAGE", "VARCHAR(256)");
         stadiumTable.tryAddColumn("ARENA_FALLBACK_IMAGE", "VARCHAR(256)");
 
@@ -154,12 +154,12 @@ final class DBUpdater {
         hofTable.createTable();
 
         var ifaMatchTable = dbManager.getTable(IfaMatchTable.TABLENAME);
-        if (ifaMatchTable.tryAddColumn("HOME_COUNTRYID", "INTEGER")){
+        if (ifaMatchTable.tryAddColumn("HOME_COUNTRYID", "INTEGER")) {
             ifaMatchTable.tryAddColumn("AWAY_COUNTRYID", "INTEGER");
         }
 
-		updateDBVersion(dbVersion, 1000);
-	}
+        updateDBVersion(dbVersion, 1000);
+    }
 
     private void updateDBv901(int dbVersion) throws SQLException {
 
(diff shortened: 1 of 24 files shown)


src/main/java/core/db/DBUpdater.java
```diff
@@ -153,6 +153,11 @@ private void updateDBv1000(int dbVersion) throws  SQLException{
         var hofTable = dbManager.getTable(HallOfFamePlayersTable.TABLENAME);
         hofTable.createTable();
 
+        var ifaMatchTable = dbManager.getTable(IfaMatchTable.TABLENAME);
+        if (ifaMatchTable.tryAddColumn("HOME_COUNTRYID", "INTEGER")){
```
sgcr: When adding `HOME_COUNTRYID` works then add `AWAY_COUNTRYID`?
Why that way?

sgcr: Please use the Code-Formatter.

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -125,7 +125,7 @@ private ConvertXml2Hrf() {
 			// We need to get hold of the currency info for the primary team, no matter which team we download.
 			usersPremierTeamInfo = XMLWorldDetailsParser.updateTeamInfoWithCurrency(usersPremierTeamInfo, mc.getWorldDetails(usersPremierTeamInfo.getLeagueId()));
 			ModuleConfig.instance().setString("CurrencyRate", usersPremierTeamInfo.getCurrencyRate().trim());
-			ModuleConfig.instance().setString("CountryId", usersPremierTeamInfo.getCountryId());
+			ModuleConfig.instance().setString("CountryId", String.valueOf(usersPremierTeamInfo.getCountryId()));
```
sgcr, line 128: Why String instead of Int?

src/main/java/core/file/xml/TeamInfo.java
```diff
@@ -2,86 +2,94 @@
 
 public class TeamInfo {
 
-	private int teamId;
```
sgcr, line 5: What about?

```
package core.file.xml;

import lombok.Getter;
import lombok.Setter;

@Setter
@Getter
public class TeamInfo {

    private int teamId;
    private Integer youthTeamId;
    private String name;
    private String country;
    private String league;
    private String leagueLevel;
    private int leagueId;
    private String currencyRate;
    private boolean primaryTeam;
    private int countryId;
}
```

src/main/java/core/file/xml/XMLTeamDetailsParser.java
```diff
@@ -18,43 +17,7 @@ public class XMLTeamDetailsParser {
...
-
-		} catch (Exception ex) {
-			HOLogger.instance().log(XMLTeamDetailsParser.class, ex);
-		}
-
-		return "-1";
-	}
-
-	public static Map<String, String> parseTeamdetailsFromString(String inputStream, int teamId) {
+	public static Map<String, String> parseTeamDetailsFromString(String inputStream, int teamId) {
```
sgcr, line 20: I am missing a test for that parse function to verify the changes.

src/main/java/module/ifa/model/IfaModel.java
```diff
@@ -1,198 +1,208 @@
...
+
+    public void reload() {
+        this.visitedStatistic = null;
+        this.hostedStatistic = null;
+        this.visitedSummary = null;
+        this.hostedSummary = null;
+        init();
+    }
+
+    public int getVisitedCountriesCount() {
```
sgcr, line 117: Good catch :D

src/main/java/module/ifa/model/IfaModel.java
```diff
@@ -1,198 +1,208 @@
...
+
+    /**
+     * Alternative country for ifa statistics
+     * Replaced by <a href="https://github.com/ho-dev/HattrickOrganizer/issues/2249">...</a> in September 2026
+     * Can be reset if HO users complain and prefer to use league stats instead of countries again
+     *
+     * @param leagueId int League id
+     * @return Integer of league's country, null if not found
+     */
+    @Deprecated
```
sgcr, line 196: Unused function.
Provide the version of HO since it is deprecated. Its introduced here as unused function.

I did not get the case when someone complains and what should we do then?

src/main/java/module/ifa/IfaMatch.java
```diff
@@ -1,98 +1,151 @@
...
+    public final void setAwayLeagueId(int awayLeagueId) {
+        this.awayLeagueId = awayLeagueId;
+    }
+
+    public final Integer getHomeCountryId() {
+        return getHomeCountryIdWithReload(false);
+    }
+
+    public final Integer getAwayCountryIdWithReload(boolean isReload) {
+        if (awayCountryId == null) {
```
sgcr, line 103: Reload should also force to call `getCountryIdFromLeagueOrReload`?

src/main/java/module/ifa/IfaMatch.java
```diff
@@ -1,98 +1,151 @@
...
+            awayCountryId = getCountryIdFromLeagueOrReload(awayLeagueId, isReload);
+        }
+        return awayCountryId;
+    }
+
+    public final Integer getHomeCountryIdWithReload(boolean isReload) {
+        if (homeCountryId == null) {
```
sgcr, line 121: Reload should also force to call `getCountryIdFromLeagueOrReload`?

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        WorldDetailLeague league = WorldDetailsManager.instance().getWorldDetailLeagueByCountryId(countryId);
+        if (league == null || league.getActiveUsers() == 0) return 0;
+        return (double) WorldDetailsManager.instance().getTotalUsers()
+            / (double) league.getActiveUsers();
+    }
+
+    public static void downloadMatch(IfaMatch match) {
+        var matchesArchive = Connector.instance().getMatchesArchive(HOVerwaltung.instance().getModel().getBasics().getTeamId(), match.getPlayedDate(), match.getPlayedDate());
+        var xmlDoc = XMLManager.parseString(matchesArchive);
+        if (xmlDoc == null) return;
```
sgcr: Please provide curly braces for the if-statement.

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        readMatch(0, xmlDoc, match);
+    }
+
+    private static boolean readMatch(int index, Document xmlDoc, IfaMatch match) {
+        var basics = HOVerwaltung.instance().getModel().getBasics();
+        int ownLeague = basics.getLiga();
+        int ownId = basics.getTeamId();
+        int ownCountryId = basics.getCountryId();
+        int opponentId;
```
sgcr: you can move that declaration where it is initialized, see `if (homeTeamID == ownId) {...`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+
+    public static void downloadMatch(IfaMatch match) {
+        var matchesArchive = Connector.instance().getMatchesArchive(HOVerwaltung.instance().getModel().getBasics().getTeamId(), match.getPlayedDate(), match.getPlayedDate());
+        var xmlDoc = XMLManager.parseString(matchesArchive);
+        if (xmlDoc == null) return;
+        readMatch(0, xmlDoc, match);
+    }
+
+    private static boolean readMatch(int index, Document xmlDoc, IfaMatch match) {
+        var basics = HOVerwaltung.instance().getModel().getBasics();
```
sgcr: `final`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        int ownLeague = basics.getLiga();
```
sgcr: `final`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        int ownId = basics.getTeamId();
```
sgcr: `final`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        int ownCountryId = basics.getCountryId();
```
sgcr: `final`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        int opponentId;
+        int opponentLeagueId = 0;
```
sgcr: `opponentLeagueId` and `opponentCountryId` should be moved to toe loop `for (TeamInfo o : opp) {`

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+        }
+        return true;
+    }
+
+    private static void insertMatches(HODateTime from, HODateTime to) {
+        HODateTime matchDate = null;
+        String matchesArchive = Connector.instance().getMatchesArchive(HOVerwaltung.instance().getModel().getBasics().getTeamId(), from, to);
+        Document doc = XMLManager.parseString(matchesArchive);
+
+        assert doc != null;
```
sgcr: I am not sure if an assert is useful in productive code.

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+
+        for (int i = 0; i < matchesCount; i++) {
+            int matchTypeId = Integer.parseInt(parseXmlElement(doc, "MatchType", i, "Match"));
+            IfaMatch match = new IfaMatch(matchTypeId);
+            if (!readMatch(i, doc, match)) {
+                continue;
+            }
+
+            var matchType = MatchType.getById(match.getMatchTyp());
+            if (matchType == MatchType.FRIENDLYCUPRULES || matchType == MatchType.FRIENDLYNORMAL
```
sgcr: Should be moved in a small private static function that explains what is meant by the types.

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+                || matchType == MatchType.INTFRIENDLYCUPRULES
+                || matchType == MatchType.INTFRIENDLYNORMAL
+                || matchType == MatchType.NATIONALFRIENDLY) {
+                if (!DBManager.instance().isIFAMatchinDB(match.getMatchId(), matchTypeId)) {
+                    DBManager.instance().insertIFAMatch(match);
+                }
+            }
+        }
+        if (matchesCount == 50) {
```
sgcr: Magic number should be a private static final constant.

src/main/java/module/ifa/PluginIfaUtils.java
```diff
@@ -83,142 +83,165 @@ public static void updateMatchesTable() {
...
+    }
+
+    public static void downloadMatch(IfaMatch match) {
+        var matchesArchive = Connector.instance().getMatchesArchive(HOVerwaltung.instance().getModel().getBasics().getTeamId(), match.getPlayedDate(), match.getPlayedDate());
+        var xmlDoc = XMLManager.parseString(matchesArchive);
+        if (xmlDoc == null) return;
+        readMatch(0, xmlDoc, match);
+    }
+
+    private static boolean readMatch(int index, Document xmlDoc, IfaMatch match) {
```
sgcr: Instead of returning a boolean the function should return a created `IfaMatch` object with filled values.
On errors an exception should be thrown that should be caught by the caller with the normal logging as it is now.

sgcr (changes requested):

src/main/java/core/db/DBUpdater.java
wsbrenk: The idea is to make sure this block only runs once.
If HOME_COUNTRY does already exist, the away country exists too.

src/main/java/core/file/xml/ConvertXml2Hrf.java
wsbrenk: I don't want to change the database entry of this value.

src/main/java/core/file/xml/TeamInfo.java
wsbrenk: This is up to you (and another ticket). I'm still not a friend of this;-)

src/main/java/core/file/xml/XMLTeamDetailsParser.java
wsbrenk: Hmm, the only changes were the size of the character D in TeamDetails and an unnecessary  if (list!=null)  in a for loop condition, because the list is a NotNull result of getElementsByTagName.

src/main/java/module/ifa/model/IfaModel.java
wsbrenk: Before this change, the IFA counts the leagues of the teams and not the countries. HATTRICK_INTERNATIONAL and maybe other international leagues were in the results.
Now this leagues are replaced by the country where the teams comes from.

The user himselfs asked the question if it was a bug or a feature.

src/main/java/module/ifa/PluginIfaUtils.java
wsbrenk: Values are now being passed as arguments into the method

src/main/java/module/ifa/PluginIfaUtils.java
wsbrenk: I thought of that too, but it doesn't work in the case of IFAMatch.downloadMatch, which I need for the new reloads

src/main/java/core/file/xml/TeamInfo.java
sgcr: As a former C++ developer, I can understand your discomfort with Lombok.

In the end, the main benefit is that there is less boilerplate code to maintain and less repetitive code to read. Things that always follow the same pattern can be generated automatically.

What remains is the code that differs from the defaults, which makes the class more expressive because the important parts stand out more clearly.

src/main/java/core/db/DBUpdater.java
sgcr: Ok.

There are several libraries available for managing incremental database schema changes across different DBMSs.

Typically, these libraries keep track of the current schema version. This is quite similar to the `DBVersion` that we update after applying our database changes.

The versions are usually organized sequentially. If you have version `1` and want to upgrade to version `3`, you first apply the changes for version `2` and then those for version `3`.

So the upgrade path is:

`1` ==> `2` ==> `3`

This is normally handled automatically by those libraries, quite similar to what our `DBUpdater` does.

One weakness in our current approach is that the development code (`901` to `1000`) is executed without increasing the current database version when the changes are applied.

As a result, every run of the application tries to perform changes again, for example creating columns that already exist. This currently does not fail because these operations are only attempted and an already existing column is not treated as an error.

I think it might be better to have clearly defined incremental steps for database schema changes and less dynamic code that tries to determine whether the database is already in the expected state.

We can leave it as it is for now, but I am thinking about a better modus operandi for future database changes.

src/main/java/core/file/xml/ConvertXml2Hrf.java
sgcr: Ok.

src/main/java/core/db/DBUpdater.java
wsbrenk: Please take a closer look before you want to change anything. That a development version doesn't adjust the database version is intentional and not an error!

And that happens on a different level than the above-mentioned block changes, which are not supposed to be repeated.

sgcr: Yes, I know that and got it before.

The behavior is clear for me. I am only thinking about that topic for a while.
Does not matter for this PR.

src/main/java/module/ifa/IfaMatch.java
wsbrenk: Changed it.

src/main/java/module/ifa/IfaMatch.java
wsbrenk: Changed it.

Issue #2449, opened 2026-05-29
[FEATURE] Total player salary on player information page.


Esch1234:
I am requesting something that should be easy to implement because the data is already being collected. Right now, the only way to see if I made any money on a player through HO is to sell the player. Then I would go in the TRANSFERS tab and select the player and see the sum of the total wages and "Total cost of ownership" to see if I made money or not. That's great, but not for players you are considering selling but aren't sure.

My request is in the SQUAD tab when you select a player, it will show their current wage, and next to that, total wages paid up to that point since they joined the club. Something like this:

Wage $670 / Total: $38860

Then at the top of that section where it shows how long they've been in the club, add the total number of weeks they've been with the club. Example:

In the team since Apr 14, 2025, 11:52:00 AM (58 weeks)

Then I can use that data along with Transfer Compare to see how much I would make per week owning the player if I sold them right now. (Especially useful for trainees)

Thank you for considering!
E

PR #2474, opened 2026-09-01
refactor: Integrate HumanDuration into HODuration

sgcr:
Extend `HumanDuration` to support weeks and seasons.

 - [x] does not require update


sgcr:
@wsbrenk Maybe its now a `HattrickDuration` instead of `HumanDuration`.

wsbrenk:
We should combine it with HODateTime.HODuration. I think we don't need two implementations of duration.

In the current context of #2449  a method to get the number of weeks without any fraction is required.

H*Duration.get(ChronoUnit::Weeks).

src/main/resources/language/German.properties
```diff
@@ -2308,9 +2308,23 @@ ArenaInfoPanel.future=Zukunft
 ArenaInfoPanel.finished_in_n_days_format=Fertigstellung in %s Tagen.
 ArenaInfoPanel.built=Ausgebaut!
 ArenaInfoPanel.stadium_image.tooltip=Klicken, um das Stadionbild neu zu laden
+Duration.seasons_abbreviation=Saisons
+Duration.weeks_abbreviation=Wochen
```
wsbrenk: In english properties the abbreviation only has one single character

src/test/java/core/model/TranslationFacilityTest.java
```diff
@@ -61,6 +61,35 @@ void tr_with_initial_translator_results_in_no_translation() {
...
+        // then
+        assertThat(translation).isEqualTo("Offensive");
+    }
+
+    @Test
+    void trSingularOrPlural() {
+        // given
+        TranslationFacility.setLanguage("English");
+        final var keySingular = "ls.module.lineup.coachtype.offensive";
+        final var keyPlural = "ls.module.lineup.coachtype.defensive";
```
wsbrenk: This examples are misleading. Please use other ones.

src/test/java/core/util/HumanDurationTest.java
```diff
@@ -14,25 +14,44 @@
...
+    private static final long MINUTES_PER_HOUR = 60L;
+    private static final long HOURS_PER_DAY = 24L;
+    private static final long DAYS_PER_WEEK = 7L;
+    private static final long WEEKS_PER_SEASON = 16L;
+
+    private static final long SECONDS_IN_MINUTE = 60L;
+    private static final long SECONDS_IN_HOUR = SECONDS_IN_MINUTE * MINUTES_PER_HOUR;
+    private static final long SECONDS_IN_DAY = SECONDS_IN_HOUR * HOURS_PER_DAY;
+    private static final long SECONDS_IN_WEEK = SECONDS_IN_DAY * DAYS_PER_WEEK;
+    private static final long SECONDS_IN_SEASON = SECONDS_IN_WEEK * WEEKS_PER_SEASON;
```
wsbrenk: why are constants being redefined here that already exist in the code?

wsbrenk (changes requested):

src/test/java/core/model/TranslationFacilityTest.java
sgcr: Yes, but it is only important that the translation for singular = `true` is returned and on `false` the other one.
I might use random no existing keys.

src/main/resources/language/German.properties
sgcr: Yes, the single character translation is not good if we add weeks and seasons because there is no useful abbreviation for that.

I added there ones above during development and then recognized that singular and plural expressive texts are better.

Will be removed.

src/test/java/core/model/TranslationFacilityTest.java
sgcr: Fixed.

src/test/java/core/util/HumanDurationTest.java
sgcr: It is about decoupling tests and productive code.

sgcr: If someone changes the definition of the same constants then the test will not fail.
Additionally they are private currently. Making them public is possible but then tests for the constants itself should exist, otherwise the test will not work on unintentional changes in the productive code.

sgcr:
> We should combine it with HODateTime.HODuration. I think we don't need two implementations of duration.
> 
> In the current context of #2449  a method to get the number of weeks without any fraction is required.
> 
> H*Duration.get(ChronoUnit::Weeks).

I'll take a look at it...

wsbrenk:
The term 'human duration' gives me a bit of a shady vibe. I’d love it if we called the implementation HODuration.

sgcr:
> The term 'human duration' gives me a bit of a shady vibe. I’d love it if we called the implementation HODuration.

Yes. Sorry for that ;-)

Will be combined - I agree.

src/main/java/core/util/HODuration.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final long MINUTES_PER_HOUR = 60L;
+    private static final long SECONDS_PER_HOUR = MINUTES_PER_HOUR * SECONDS_PER_MINUTE;
+    private static final long HOURS_PER_DAY = 24L;
+    private static final long SECONDS_PER_DAY = HOURS_PER_DAY * SECONDS_PER_HOUR;
+    private static final long SECONDS_PER_WEEK = DAYS_PER_WEEK * SECONDS_PER_DAY;
+    private static final long SECONDS_PER_SEASON = WEEKS_PER_SEASON * SECONDS_PER_WEEK;
+
+    private final long totalSeconds;
+
+    public HODuration(long seasons, long days) {
```
wsbrenk, line 47: What about staying with int arguments here to remain compatible with former implementation. The cast is then done in one single place (here) and other code remains as it is?!

src/main/java/core/util/HODuration.java
```diff
@@ -0,0 +1,104 @@
...
+        return Math.floorMod(totalSeconds, SECONDS_PER_DAY) / SECONDS_PER_HOUR;
+    }
+
+    public long getMinutes() {
+        return Math.floorMod(totalSeconds, SECONDS_PER_HOUR) / SECONDS_PER_MINUTE;
+    }
+
+    public long getSeconds() {
+        return Math.floorMod(totalSeconds, SECONDS_PER_MINUTE);
+    }
```
wsbrenk, line 107: getWeeks is missing here.

Optional we could also add method 

long getTotalWeeks() {return getSeasons()*16 + getWeeks();}
This is what is required in #2449.

src/main/java/core/util/HODuration.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public String toString() {
+        return getSeasons() + " (" + getDays() + ")";
+    }
+
+    public double toDouble() {
+        return totalSeconds / (double) SECONDS_PER_SEASON;
+//        return getSeasons() + getDays() / (double) DAYS_PER_SEASON;
```
wsbrenk: Remove old code

src/main/java/core/util/HumanDuration.java
```diff
@@ -1,67 +1,98 @@
...
 import java.util.ArrayList;
+import java.util.List;
+import java.util.function.ObjLongConsumer;
+import java.util.function.ToLongFunction;
 
-@Builder
+@Builder(access = AccessLevel.PACKAGE) // visibility for testing
 @EqualsAndHashCode
 @Getter
 public class HumanDuration {
```
wsbrenk: Still planned to incorporate this to HODuration?

wsbrenk (changes requested):
Great job!

src/main/java/core/util/HumanDuration.java
sgcr: Yes.

src/main/java/core/util/HODuration.java
sgcr: Added int constructor for compatibility.

src/main/java/core/util/HODuration.java
sgcr: Added `getWeeks()`.
Added `getTotalWeeks()`.

src/main/java/core/util/HODuration.java
sgcr: Removed.

src/main/java/core/util/HumanDuration.java
sgcr: Assimilated. Resistance is futile! 😉

src/main/java/core/util/HODuration.java
```diff
@@ -0,0 +1,208 @@
...
+    public long getDaysInSeason() {
+        return Math.floorMod(totalSeconds, SECONDS_PER_SEASON) / SECONDS_PER_DAY;
+    }
+
+    public long getTotalWeeks() {
+        return Math.floorDiv(totalSeconds, SECONDS_PER_WEEK);
+    }
+
+    public static HODuration between(HODateTime from, HODateTime to) {
+        return new HODuration(0, Duration.between(from.instant, to.instant).plus(12, ChronoUnit.HOURS).toDays());
```
wsbrenk, line 125: Is rounding (+ 0.5 days) still correct, when HODuration now is implemented by seconds?

Maybe we should rename tthis implementation to daysBetween and implement between without any rounding.

wsbrenk (changes requested):

src/main/java/core/util/HODuration.java
sgcr: `between` should be the correct distance.
`daysBetween` sounds good. Will be added...

sgcr: Done.

src/main/java/module/transfer/history/PlayerDetailPanel.java
```diff
@@ -343,7 +343,7 @@ private void updatePanel() {
                     }
                 }
 
-                var activeDuration = HODuration.between(arrivalDate, to);
+                var activeDuration = HODuration.betweenDays(arrivalDate, to);
                 if (activeDuration.getSeasons() >= 0) {
                     lengthOfStayInTeam.setText(activeDuration.toString());
```
wsbrenk: toString will return seconds. In that case we should use between instead of betweenDays.
Alternativly use toAgeString

wsbrenk (changes requested):
One last change please

src/main/java/module/transfer/history/PlayerDetailPanel.java
sgcr: **Yes. Totally right, I didn't want to change that behavior.**

The problem with `toString` is that the calls are not so easy to find, because the function in overridden from `Object`.
Another problem is that the `toString` is used for the result of tests in the IDE.
In case of failing tests the error messages are looking weird due to missing fraction beyond days...

In the first place I wanted the behavior unchanged for `toString` but the thing with the tests is the reason for the existence of `toAgeString` and `toObjectString` and a changed `toString` function.

In general: `toString` should return a text that explains the complete content of the object and/or make it possible to identify it.
That is one reason that `toString` if most of the time generated from Lombok instead of written manually.

sgcr: Changed to 'toAgeString' here.

wsbrenk (approved):

Issue #2306, opened 2025-08-12
[BUG] font size change currently limited to size 20 due to fixed table heights in some tables


datatramp:
**Describe the bug**
font size change currently limited to size 20 due to fixed table heights in some tables.
You can change to larger fonts, but visibility is capped due to fixed cell height.

Please make table cells adjustable not only in width but also in height.

... and while you are at it,
please show all squad data (numbers and icons) in the same way - right bound OR left bound OR center bound in the columns, thank you!

wsbrenk:
<img width="2546" height="1561" alt="Image" src="https://github.com/user-attachments/assets/651c6496-2a44-4108-ab15-162b2a4c4b44" />

PR #2475, opened 2026-09-03
#2306 Fix table layouts for large user font sizes

wsbrenk:

   - fixes issue #2306

 - [x] has been updated
 - [ ] does not require update

### DIFF
--- a/src/main/java/module/series/SeriesTablePanel.java
+++ b/src/main/java/module/series/SeriesTablePanel.java
@@ -148,34 +148,34 @@ class SeriesTablePanel extends ImagePanel {
 
 	private void initComponents() {
 
-		final GridBagLayout layout = new GridBagLayout();
-		final GridBagConstraints constraints = new GridBagConstraints();
-		final JPanel panel = new JPanel(layout);
+        final GridBagLayout layout = new GridBagLayout();
+        final GridBagConstraints constraints = new GridBagConstraints();
+        final JPanel panel = new JPanel(layout);
 
-		// combobox
-		constraints.gridx = 0;
-		constraints.gridy = 0;
-		constraints.insets = new Insets(4, 0, 4, 100);
-		constraints.anchor = GridBagConstraints.LINE_END;
-		m_jcbStatsAggType = new JComboBox(new String[]{"Max", "Avg"});
-		m_jcbStatsAggType.addActionListener(e -> populateSerieTableStatsOnly());
+        // combobox
+        constraints.gridx = 0;
+        constraints.gridy = 0;
+        constraints.insets = new Insets(4, 0, 4, 100);
+        constraints.anchor = GridBagConstraints.LINE_END;
+        m_jcbStatsAggType = new JComboBox(new String[]{"Max", "Avg"});
+        m_jcbStatsAggType.addActionListener(e -> populateSerieTableStatsOnly());
 
 
-		var fontSize = UserParameter.instance().fontSize;
-		layout.setConstraints(m_jcbStatsAggType, constraints);
-		panel.add(m_jcbStatsAggType);
+        var fontSize = UserParameter.instance().fontSize;
+        layout.setConstraints(m_jcbStatsAggType, constraints);
+        panel.add(m_jcbStatsAggType);
 
-		// serie Table
-		constraints.gridy = 1;
-		constraints.insets = new Insets(0, 4, 4, 4);
-		seriesTable.setDefaultRenderer(Object.class, new HODefaultTableCellRenderer());
-		seriesTable.setRowHeight(2 * fontSize);
-		layout.setConstraints(seriesTable, constraints);
-		panel.add(seriesTable);
+        // serie Table
+        constraints.gridy = 1;
+        constraints.insets = new Insets(0, 4, 4, 4);
+        seriesTable.setDefaultRenderer(Object.class, new HODefaultTableCellRenderer());
+        seriesTable.setRowHeight(2 * fontSize);
+        layout.setConstraints(seriesTable, constraints);
+        panel.add(seriesTable);
 
-		setLayout(new BorderLayout());
-		add(panel);
-	}
+        setLayout(new BorderLayout());
+        add(panel);
+    }
 
 	//  Init from the HRF-XML
 	private void initSeriesTable() {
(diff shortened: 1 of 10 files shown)


src/main/java/tool/hrfexplorer/HrfExplorer.java
```diff
@@ -36,256 +35,171 @@
  * @author KickMuck
  */
 
-public class HrfExplorer extends ImagePanel implements ActionListener,ItemListener,MouseListener,IRefreshable,TableColumnModelListener
```
sgcr, line 39: @wsbrenk At lot of reformatting in this PR?

sgcr (changes requested):
Sorry, it is not easy to find the real changes beside reformatting the code.

Can you assist?

src/main/java/tool/hrfexplorer/HrfExplorer.java
wsbrenk: "hide whitespace" should be selected. Then I think the display is acceptable.
HrfExplorer was a very old code needing a lot of code cleaning.

sgcr: Unfortunately not sufficient.

I appreciate your reformatting but that do you think about reformatting all the code of the application in a separate commit using a defined clear code formatter?

I know, it is a radical step, but it hurts only one time! ;-D

wsbrenk: I can't manage that. Even if I branch off from master and initially just reformat the file, the diff of the new file to the formatted one still becomes totally unreadable.

Basically, only the hard-coded table row heights have been adjusted. Beyond that, a lot of the analyzer's suggestions were implemented, which unfortunately led to the reformatting.

src/main/java/tool/hrfexplorer/HrfExplorer.java
```diff
@@ -318,78 +233,71 @@ private void initialize() {
...
-        m_Table_Details.setDefaultRenderer(JPanel.class, m_renderer );
-    	m_Table_Details.setRowHeight(1,m_int_Hoehe_DetailPanels);
-    	m_Table_Details.setRowHeight(0,m_int_Hoehe_Label + 5);
-    	TableColumn colFixed = m_Table_Details.getColumnModel().getColumn(0);
-    	colFixed.setPreferredWidth(m_int_Breite_Detail_Fixed);
-
-
+        m_TableModel_Details.addColumn(" ", fixedColumn);
+        m_Table_Details.setDefaultRenderer(JPanel.class, m_renderer);
+        setDetailsPanelRowHeights();
```
wsbrenk, line 295: new private method (called twice)

src/main/java/tool/hrfexplorer/HrfExplorer.java
sgcr: But you can mark lines with IntelliJ and just reformat the marked lines.
That is a general approach instead of reformatting whole files.

Solution A for me:
I've used the diff view now in IntelliJ from this file against your file in your temporary branch `merge/2306`.
That works because that is like diffing against a reformatted project but only for that file.

Solution B for me:
I can review the diff the IntelliJ while comparing the file against master with the option `Ignore import and formatting`.
The diff is very similar to solution A.

sgcr, line 39: Please use the code formatter for this last missing spot in the file here.

sgcr: Changes are good.

There is still a lot of stuff to do here but this exceeds the intention of this PR... ;-)

src/main/java/tool/hrfexplorer/HrfDetails.java
```diff
@@ -88,8 +90,8 @@ void createDates() {
 							.substring(0, 19));
 				}
 			}
-		} catch (SQLException sexc) {
-			HrfExplorer.appendText("" + sexc);
+		} catch (SQLException e) {
+            HOLogger.instance().error(this.getClass(), "Error in createDates: " + e.getMessage());
```
sgcr: Please add a `private static final String` for the prefix of the error log message and use it. It is used 3 times in that file.

src/main/java/module/series/SeriesTablePanel.java
```diff
@@ -148,34 +148,34 @@ private void setTableColumnWidth() {
 
 	private void initComponents() {
 
-		final GridBagLayout layout = new GridBagLayout();
```
sgcr, line 151: Only the changes for reformatting without any change.

src/main/resources/release_notes.md
```diff
@@ -53,6 +53,9 @@
 
 * Added support of all 13 **Player Categories** (#2418)
 * Fix line feed handling in owner notes. Tooltip shows notes as multiline (#2455)
+* The bug where the number `100` was displayed for the shirt number when the player had no shirt number has been fixed.
```
sgcr, line 56: Why is that in the diff here?

I am wondering because the fix is a long time ago.

src/main/java/tool/hrfexplorer/HrfTableCellRenderer.java
```diff
@@ -11,142 +11,98 @@
...
+        } else {
+            label = new JLabel();
+        }
 
-		}
-		else
-		{
-			label.setIcon((ImageIcon)value);
-		}
+        if (value instanceof ImageIcon) {
```
sgcr: You can write `if (value instanceof ImageIcon imageIcon) {` and then avoid the superfluous cast in the next line:

```java
label.setIcon(imageIcon);
```

sgcr (approved):
Approved.

You can make the small corrections regarding my comments if you like.

However, please consider establishing a clear policy for reformatting and documenting it, or alternatively reformatting the whole project once and introducing a common automatic code formatter.

Large diffs containing mostly reformatting and no actual code changes make reviews unnecessarily difficult for both of us.

😃

src/main/java/tool/hrfexplorer/HrfExplorer.java
wsbrenk: Eagle eye

src/main/java/module/series/SeriesTablePanel.java
wsbrenk: Found setRowHeight in this file and maybe i tried to remove it. This might have caused the damned reformatting. Sorry for that!

src/main/resources/release_notes.md
wsbrenk: There were two sections called Youth. I put them together.

Issue #2432, opened 2026-02-17
[BUG] Matches/General: 'Style of play' is alyways 'Neutral'


sgcr:
**Describe the bug**
The information in the `General` tab of `Matches` shows the `Style of play` play for a game.

**To Reproduce**
1. Go to `Matches` / `General`
2. Click on a game that you know how you have played regarding the `Style of play`.
3. In the match with matchid=758154320 (of course my game) I played `100% Defensive` and it shows `Neutral` for the home team.
4. The away team played `100% Offensive` and it shows also `Neutral`.
5. All of games show `Neutral` which is definitely not true. 

**Screenshots of [matchid=758154320](https://www87.hattrick.org/Club/Matches/Match.aspx?matchId=758154320&sourceSystem=Hattrick)**
Hattrick web page:
<img width="664" height="489" alt="Image" src="https://github.com/user-attachments/assets/d5b415e8-626b-45e2-8b9e-6c48de5b14d4" />

HO:
<img width="1053" height="399" alt="Image" src="https://github.com/user-attachments/assets/b02d35b9-ad0f-480b-9dba-f945a9b8577a" />

**Platform information:**
 - OS: [Windows / macOS / Linux]
 - Version (9.x and 10.x)

PR #2476, opened 2026-09-03
#2432 Fix download of style of play

wsbrenk:

   - fixes issue #2432

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+     * 3	30% offensive
+     * 4	40% offensive
+     * 5	50% offensive
+     * 6	60% offensive
+     * 7	70% offensive
+     * 8	80% offensive
+     * 9	90% offensive
+     * 10	100% offensive
+     */
+    private final Integer val;
```
sgcr: `int`

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+        HOLogger.instance().warning(StyleOfPlay.class, "Unknown style of play: " + styleOfPlay);
+        return new StyleOfPlay(null);
+    }
+
+    public static StyleOfPlay Neutral() {
+        return new StyleOfPlay(0);
+    }
+
+    public static Integer toInt(StyleOfPlay in) {
+        if (in != null) return in.val;
```
sgcr: Please add curly braces.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+        if (styleOfPlay == null) {
+            return null;
+        }
+        if (styleOfPlay >= -10 && styleOfPlay <= 10) {
+            return new StyleOfPlay(styleOfPlay);
+        }
+        if (styleOfPlay == -1000) { // old matches has -1000
+            return StyleOfPlay.Neutral();
+        }
+        HOLogger.instance().warning(StyleOfPlay.class, "Unknown style of play: " + styleOfPlay);
```
sgcr: I think that it is not a good idea to allow and warn unknown style of play values.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+        return new StyleOfPlay(null);
```
sgcr, line 60: Instead of returning a `null` style of play we should throw here.
I think it is a parser error then.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+
+    private StyleOfPlay(Integer styleOfPlay) {
+        this.val = styleOfPlay;
+    }
+
+    public static StyleOfPlay fromInt(Integer styleOfPlay) {
+        if (styleOfPlay == null) {
+            return null;
+        }
+        if (styleOfPlay >= -10 && styleOfPlay <= 10) {
```
sgcr: Please add constants for `-10` and `10`.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+            return new StyleOfPlay(styleOfPlay);
+        }
+        if (styleOfPlay == -1000) { // old matches has -1000
```
sgcr: Please add a constant.

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: Old database entries has null values stored in the styleOfPlay column. I want to handle this as "unknown" (null) values.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+    public static StyleOfPlay Neutral() {
+        return new StyleOfPlay(0);
+    }
+
+    public static Integer toInt(StyleOfPlay in) {
+        if (in != null) return in.val;
+        return null;
+    }
+
+    public static StyleOfPlay Defensive() {
```
sgcr: Lower case and on the other side it could be a public static final constant.
I see not benefit of creating a new object each time.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+        return new StyleOfPlay(-10);
+    }
+
+    public static StyleOfPlay Offensive() {
```
sgcr: Lower case and on the other side it could be a public static final constant.
I see not benefit of creating a new object each time.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+            return new StyleOfPlay(styleOfPlay);
+        }
+        if (styleOfPlay == -1000) { // old matches has -1000
+            return StyleOfPlay.Neutral();
+        }
+        HOLogger.instance().warning(StyleOfPlay.class, "Unknown style of play: " + styleOfPlay);
+        return new StyleOfPlay(null);
+    }
+
+    public static StyleOfPlay Neutral() {
```
sgcr: Lower case and on the other side it could be a public static final constant.
I see not benefit of creating a new object each time.

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: I'm not sure if -1000 is the only undocumented style of play code. If so, this warning should never appear in any log.

src/main/java/module/lineup/Lineup.java
```diff
@@ -334,12 +337,12 @@ public String getAttitudeName(int attitude) {
...
 
-	public void setStyleOfPlay(int style) {
+	public void setStyleOfPlay(Integer style) {
 		ratingRevision++;
 		settings.m_iStyleOfPlay = style;
 	}
-	
-	public int getCoachModifier() {
+
+	public Integer getCoachModifier() {
```
sgcr, line 345: Other bugfix?

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: No - here i want to handle old matches stored in database where no style of play were stored (null).

src/main/java/module/lineup/Lineup.java
wsbrenk: No - coach modifier is a synonym to style of play, used in the chpp files.

src/main/java/core/model/match/MatchLineupTeam.java
```diff
@@ -224,30 +224,31 @@ public void calcStyleOfPlay (TrainerType trainerType, int tacticAssistants) {
...
 	 */
 	public final StyleOfPlay getStyleOfPlay() {
 		return StyleOfPlay.fromInt(this.lineup.getCoachModifier());
 	}
-	
+
 	// returns offensive, defensive or neutral depending on styleOfPlay
 	// e.g. -3 is 30% defensive, 10 is 100% offensive
 	public static String getStyleOfPlayName(StyleOfPlay styleOfPlay) {
+        if (StyleOfPlay.toInt(styleOfPlay) == null) { return ""; }
```
sgcr: Please use the standard code-formatter. ;-)

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
-4	40% offensive
-5	50% offensive
-6	60% offensive
-7	70% offensive
-8	80% offensive
-9	90% offensive
-10	100% offensive
-*/
-
-   private final int val;
```
sgcr, line 30: Provide a get function for the value and rename to `value` because it is one. ;-)

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+     * 5	50% offensive
+     * 6	60% offensive
+     * 7	70% offensive
+     * 8	80% offensive
+     * 9	90% offensive
+     * 10	100% offensive
+     */
+    private final Integer val;
+
+    private StyleOfPlay(Integer styleOfPlay) {
```
sgcr: `styleOfPlay` is the `value` here.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+        this.val = styleOfPlay;
+    }
+
+    public static StyleOfPlay fromInt(Integer styleOfPlay) {
```
sgcr: `styleOfPlay` is the `value` here

src/main/java/core/model/match/MatchLineupTeam.java
sgcr: Two times getting the int value of the `StyleOfPlay` is one time to much.
See line 239

src/main/java/module/lineup/lineup/MatchAndLineupSelectionPanel.java
```diff
@@ -500,8 +500,9 @@ private List<Integer> getValidStyleOfPlayValues()
         return IntStream.rangeClosed(min, max).boxed().collect(Collectors.toList());
     }
 
-    public void setStyleOfPlay(StyleOfPlay style){
-        Helper.setComboBoxFromID(m_jcbStyleOfPlay, StyleOfPlay.toInt(style));
+    public void setStyleOfPlay(StyleOfPlay style) {
+        int id = StyleOfPlay.toInt(style) != null ? StyleOfPlay.toInt(style) : 0;
```
sgcr: Means `0` neutral?

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Okay. I see no benefit in storing the read `null` from the database.
I think we use "neutral" then. `null` values from the database should be read to neutral.

src/main/java/core/model/match/StyleOfPlay.java
sgcr: What do you think is the benefit of holding a style of play `null` in the application?

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Okay.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,66 @@
...
-
-   public static StyleOfPlay Defensive() {
-      return new StyleOfPlay(-10);
-   }
-
-   public static StyleOfPlay Offensive() {
-      return new StyleOfPlay(10);
-   }
+    private static final int NEUTRAL_STYLE_OF_PLAY = 0;
+    private static final int DEFENSIVE_STYLE_OF_PLAY = -10;
```
sgcr: Suggestion `MAX_DEFENSIVE_STYLE_OF_PLAY`

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,66 @@
...
+    private static final int OFFENSIVE_STYLE_OF_PLAY = 10;
```
sgcr: Suggestion `MAX_OFFENSIVE_STYLE_OF_PLAY`

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: Nope. Unknown Style of play is displayed in panel as empty string. Using the update button will download the correct values and display them.

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: There are databases in use, which are older than the introduction of the download of style of play. They contain match entries with null values. 
The user has an option to reload those matches and get the correct values.

sgcr: But that is actively done and you have to know to reload, right?

I guess `null` is treated as `NEUTRAL` in the application, right?

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Ok...

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: Done

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
-   public static StyleOfPlay Offensive() {
-      return new StyleOfPlay(10);
-   }
+    private static final int NEUTRAL_STYLE_OF_PLAY = 0;
+    private static final int MIN_DEFENSIVE_STYLE_OF_PLAY = -10;
+    private static final int MAX_OFFENSIVE_STYLE_OF_PLAY = 10;
+    private static final int OLD_MATCHES_STYLE_OF_PLAY = -1000;
+
+    public static final StyleOfPlay NEUTRAL = new StyleOfPlay(NEUTRAL_STYLE_OF_PLAY);
+    public static final StyleOfPlay DEFENSIVE = new StyleOfPlay(MIN_DEFENSIVE_STYLE_OF_PLAY);
```
sgcr, line 14: I think it is the max and not min.
Like playing with the strongest (max) defense.

src/main/java/core/model/match/MatchLineupTeam.java
```diff
@@ -224,30 +224,33 @@ public void calcStyleOfPlay (TrainerType trainerType, int tacticAssistants) {
...
-		} else {
-			s = (style > 0) ? TranslationFacility.tr("ls.team.styleofplay.offensive") :
-				TranslationFacility.tr("ls.team.styleofplay.defensive"); 
-		}
-		return Math.abs(style * 10) + "% " + s;
-	}
-	
+        var style = StyleOfPlay.toInt(styleOfPlay);
+        if (style == null) {
+            return "";
```
sgcr, line 239: What about `TranslationFacility.tr("ls.general_label.not_available_abbreviation")` instead of `""`?

src/main/java/core/model/match/MatchLineupTeam.java
```diff
@@ -224,30 +224,33 @@ public void calcStyleOfPlay (TrainerType trainerType, int tacticAssistants) {
...
+        }
+        String s;
+        if (style == 0) {
```
sgcr, line 242: Same code here to get the label like in:
`MatchAndLineupSelectionPanel.updateStyleOfPlayComboBox`.

There should be one function that creates that label.

What do you think?

sgcr (changes requested):

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: `        if (value == null || value >= MIN_DEFENSIVE_STYLE_OF_PLAY && value <= MAX_OFFENSIVE_STYLE_OF_PLAY) {
            return new StyleOfPlay(value);
        }
`

value >= MAX_DEFENSIVE doesn't look like a valid state

src/main/java/core/model/match/MatchLineupTeam.java
wsbrenk: Changed MatchAndLineupSelectionPanel.updateStyleOfPlayComboBox

src/main/java/core/model/match/MatchLineupTeam.java
wsbrenk: I find the empty string intuitive and simpler

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Then `FULLY_DEFENSIVE_STYLE_OF_PLAY`  and `FULLY_OFFENSIVE_STYLE_OF_PLAY`.

src/main/java/core/model/match/StyleOfPlay.java
```diff
@@ -1,57 +1,69 @@
...
+    private final Integer value;
+
+    private StyleOfPlay(Integer value) {
+        this.value = value;
+    }
+
+    public Integer getValue() {
+        return value;
+    }
+
```
sgcr, line 51: I am missing here:
```
public int getPercentage() {
        if (value == null) {
            throw new IllegalStateException("Unknown style of play has no percentage");
        }
        return Math.abs(value) * 10;
    }
```
because the value object should know its percentage value.

sgcr, line 51: I am missing here:
```
    public boolean isNull() {
        return value == null;
    }
```
because the value shall have a `null` state but that must not be exposed with the value itself.
Other name `isNotSet`.

sgcr, line 51: I am missing also:
```
    public boolean isNeutral() {
        return value != null && value == NEUTRAL_STYLE_OF_PLAY;
    }

    public boolean isOffensive() {
        return value != null && value > NEUTRAL_STYLE_OF_PLAY;
    }

    public boolean isDefensive() {
        return value != null && value < NEUTRAL_STYLE_OF_PLAY;
    }
```
so that you can ask the value what it "thinks" that it is or not.

This makes it possible to add also:
```
    public enum Classification {
        NEUTRAL,
        DEFENSIVE,
        OFFENSIVE
    }

    public Optional<Classification> classify() {
        if (isNull()) {
            return Optional.empty();
        } else if (isNeutral()) {
            return Optional.of(Classification.NEUTRAL);
        } else if (isOffensive()) {
            return Optional.of(Classification.OFFENSIVE);
        } else if (isDefensive()) {
            return Optional.of(Classification.DEFENSIVE);
        }
        throw new IllegalStateException("Unknown classification for value '%s'".formatted(value));
    }
```
because a classification of the value is necessary...

sgcr (changes requested):

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: If I will introduce this method, the analyzer will complain, that there is no usage of this method.

wsbrenk: I would introduce this, if it will be used anywhere

wsbrenk: I would introduce this, if it will be used anywhere

sgcr: Well you said you want to have a null value for the object.
You can use the function for the classification:
Can be used for the classification and/or generation of the label (e.g.):
```
   public enum Classification {
        NEUTRAL,
        DEFENSIVE,
        OFFENSIVE
    }

    public Optional<Classification> classify() {
        if (isNull()) {
            return Optional.empty();
        } else if (isNeutral()) {
            return Optional.of(Classification.NEUTRAL);
        } else if (isOffensive()) {
            return Optional.of(Classification.OFFENSIVE);
        } else if (isDefensive()) {
            return Optional.of(Classification.DEFENSIVE);
        }
        throw new IllegalStateException("Unknown classification for value '%s'".formatted(value));
    }
```

sgcr: Should be used within the label function `getStyleOfPlayName`.

src/main/java/core/model/match/MatchLineupTeam.java
sgcr: Well, I guess you and only a few others may actually have `null` values for style of play in the database.

In my database I only have `0` values, which seems to correspond to the bug.

In general, I don't think an empty label is a good representation for a value that is not available. Applications commonly use an explicit placeholder like `-`, `(empty)`, `(not set)`, `n.a.`, or `N/A`, optionally combined with an additional visual distinction such as a different color or italic font.

An empty string is ambiguous and can easily be interpreted as a rendering bug rather than as an intentionally unavailable value.

Since we already have a localized `not_available_abbreviation`, I would prefer using that here.

That said, I don't feel strongly about this particular case. I just wouldn't consider an empty string a clear design choice for representing an unavailable value.

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Should be used then in `getStyleOfPlayName` e.g.:
```
    public static String getStyleOfPlayName(StyleOfPlay styleOfPlay) {
        return styleOfPlay.classify().map(classification -> {
                final var classificationString = getTranslation(classification);
                if (classification != StyleOfPlay.Classification.NEUTRAL) {
                    final var percentValue = styleOfPlay.getPercentage();
                    return percentValue + "% " + classificationString;
                } else {
                    return classificationString;
                }
            })
            .orElse(TranslationFacility.tr("ls.general_label.not_available_abbreviation"));
    }
```

sgcr (changes requested):

src/main/java/core/model/match/StyleOfPlay.java
sgcr: Alternatively
`MOST_DEFENSIVE_STYLE_OF_PLAY` and `MOST_OFFENSIVE_STYLE_OF_PLAY` 

or back to mathematics 

`MIN_STYLE_OF_PLAY` and
`MAX_STYLE_OF_PLAY`.

src/main/java/core/model/match/MatchLineupTeam.java
wsbrenk: "not available" could be misunderstood as a valid setting, in line with Beckenbauer's motto: „Geht's raus, habt Spaß und spielt's Fußball“

I will follow your suggestion and introduce ls.team.styleofplay.na=-
New work for our translators;-)

src/main/java/core/model/match/StyleOfPlay.java
wsbrenk: Changed it

src/main/java/core/model/match/MatchLineupTeam.java
```diff
@@ -224,30 +224,32 @@ public void calcStyleOfPlay (TrainerType trainerType, int tacticAssistants) {
...
-	}
-	
+        if (styleOfPlay.isNotAvailable()) {
+            return TranslationFacility.tr("ls.team.styleofplay.na");
+        }
+        String s;
+        if (styleOfPlay.isNeutral()) {
+            return TranslationFacility.tr("ls.team.styleofplay.neutral");
+        } else {
+            s = (styleOfPlay.isOffensive()) ? TranslationFacility.tr("ls.team.styleofplay.offensive") :
```
sgcr: I am missing the `isDefensive()` call because there could be the case that the value is not valid for that.

src/main/java/core/model/match/MatchLineupTeam.java
```diff
@@ -224,30 +224,32 @@ public void calcStyleOfPlay (TrainerType trainerType, int tacticAssistants) {
...
+                TranslationFacility.tr("ls.team.styleofplay.defensive");
+        }
+        return Math.abs(styleOfPlay.getValue() * 10) + "% " + s;
```
sgcr: Can be moved in the `else` case together with the declaration of `s`.

sgcr (changes requested):

PR #2478, opened 2026-09-04
feat: Arena Sizer: Tab 'Arena': Display of number with the thousands separator

sgcr:
Display the values in the tab 'Arena' with the thousands separator.

 - [x] has been updated

### DIFF
--- a/src/main/java/tool/arenasizer/ArenaPanel.java
+++ b/src/main/java/tool/arenasizer/ArenaPanel.java
@@ -14,156 +14,171 @@ import javax.swing.*;
 import javax.swing.table.TableColumnModel;
 import java.awt.*;
 import java.math.BigDecimal;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 final class ArenaPanel extends JPanel {
 
-	private final JTable m_jtArena = new JTable();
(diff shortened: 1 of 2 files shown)


src/main/java/tool/arenasizer/ArenaPanel.java
```diff
@@ -14,156 +14,171 @@
...
+    private static final int COLUMN_INDEX_MAXIMUM = 2;
+    private static final int COLUMN_INDEX_AVERAGE = 3;
+    private static final int COLUMN_INDEX_MINIMUM = 4;
+    private static final int NUMBER_OF_COLUMNS = 5;
+
+    private final JTable jTable = new JTable();
+
+    private Stadium currentStadium;
+    private Stadium[] stadiumsMaxAvgMin;
+    private IHOTableCellEntry[][] table;
```
wsbrenk, line 48:
```suggestion
    private IHOTableCellEntry[][] tableCellEntries;
```

wsbrenk (changes requested):
With so much work you're putting into this module, does it maybe make sense to convert the tables to the FixedColumnTable and HOTableModel classes?

src/main/java/tool/arenasizer/ArenaPanel.java
sgcr: Done.

sgcr:
> With so much work you're putting into this module, does it maybe make sense to convert the tables to the FixedColumnTable and HOTableModel classes?

Well, I want to redesign the module in general and this is only an intermediate step to make it better according to common standards in the application.

I will create an Issue with the idea based on that page here afterwards:
https://wiki.hattrick.org/wiki/Arena

wsbrenk (approved):

PR #2488, opened 2026-09-11
feat: Override Access Token for Developers

sgcr:
I wanted to have the possibility to force a specific Token and Token Secret so that it is easier to connect always everywhere with my Hattrick account.
For that purpose there is now the environment variable `HO_ENCRYPTED_TOKEN` and `HO_ENCRYPTED_TOKEN_SECRET`.
If these variables are set with the encrypted values of the Token and the Token Secret, then the application will read and use these values every time when the encrypted Token and encrypted Token Secret is read.
If read from the environment this is logged in the log file of the application.

 - [x] does not require update


wsbrenk (approved):

sgcr:
@wsbrenk Shall I add some developer documentation for that feature?
Do you need that feature too?
Should it be possible to override with an unencrypted Token and Token Secret?

src/main/java/core/net/Connector.java
```diff
@@ -58,14 +59,34 @@ private Connector() {
...
+    private static OAuth1AccessToken createOAuthAccessToken() {
+        var accessToken = createAccessToken();
+        return new OAuth1AccessToken(accessToken.getToken(), accessToken.getTokenSecret());
+    }
+
+    private static AccessToken createAccessToken() {
+        return AccessToken.ofEncryptedData(getCryptedToken(), getCryptedTokenSecret());
+    }
+
+    private static String getCryptedToken() {
```
wsbrenk: getEncryptedToken?

src/main/java/core/net/Connector.java
```diff
@@ -58,14 +59,34 @@ private Connector() {
...
+        return fromEnvAndLogOrElseGet("HO_ENCRYPTED_TOKEN", () -> UserParameter.instance().getEncryptedToken());
+    }
+
+    private static String getCryptedTokenSecret() {
```
wsbrenk: getEncryptedTokenSecret?

src/main/java/core/net/Connector.java
```diff
@@ -58,14 +59,34 @@ private Connector() {
...
+        return fromEnvAndLogOrElseGet("HO_ENCRYPTED_TOKEN_SECRET", () -> UserParameter.instance().getEncryptedTokenSecret());
+    }
+
+    private static String fromEnvAndLogOrElseGet(String name, Supplier<String> supplier) {
+        var env = Optional.ofNullable(System.getenv(name));
```
wsbrenk: Isn't the environment just another supplier for the info in addition to the userparameter access?
What about 
private static String getFromSuppliers

src/main/java/core/util/HOEncryption.java
```diff
@@ -39,9 +42,26 @@ public static String decryptString(String text) {
...
+     *
+     * @param text the text that shall be encrypted
+     * @return encrypted text
+     * @deprecated Provided for for convenience. Please use {@link #encryptString(String)} instead. Can be removed
+     * in a later version.
      */
+    @Deprecated(since = "10.0")
     public static String cryptString(String text) {
+        return encryptString(text);
+    }
```
wsbrenk: Are there so many places using this method? I think simply renaming would do the job here.

wsbrenk (approved):
I think i will not use this new feature. If I create a new database, I will implicitly test the workflow to create the new access tokens with hattrick web site. 

Another point: I would prefer to discuss such things in the issues. Doing it there will result in a changed log in the milestones. A link in the release notes will reference the detailed changed log in the milestones.

src/main/java/core/net/Connector.java
sgcr: Yes it is another supplier in some way but the reason here is to prevent the read call of the getter from the user parameter if we have the environment variable set.

src/main/java/core/net/Connector.java
sgcr: Done.

src/main/java/core/net/Connector.java
sgcr: Done.

src/main/java/core/net/Connector.java
wsbrenk: I didn't think the implementation should be changed – only the name of the method I find a bit 'over the top'

sgcr: Name changed to `getFromEnvOrElse`.
Implementation slightly changed.

Issue #2491, opened 2026-09-13
[BUG] Developer Tools: Debug / SQL Editor: SQL statements are always executed as upper case


sgcr:
**Describe the bug**
When executing SQL statements in the SQL Editor from the menu `Debug` these statements are always executed in upper case.

**To Reproduce**
1. Go to 'Debug'
2. Choose `SQL Editor`
3. Enter the statement `select config_key from USERCONFIGURATION where config_key like 'A%'`
4. Press `F5` to execute.
5. Verify that you see at least one entry starting with an `A` and followed by any lower case character.
6. Change the statement so that the lower case character is entered after the `A`. Example: `Enter the statement `select config_key from USERCONFIGURATION where config_key like 'Ac%'``
7. Press `F5` to execute.
8. The same entry that should have been selected by `'A%'` could not be found by `'Ac%'`.

**Platform information:**
 - OS: Windows, macOS,  Linux
 - Version 10.x

**Additional context**
-

PR #2492, opened 2026-09-13
[BUG] #2491: Developer Tools: Debug / SQL Editor: Allow entered lower case characters to be executed as lower case instead of upper case

sgcr:
   - fixes issue #2491 

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/db/frontend/SQLDialog.java
```diff
@@ -161,43 +160,44 @@ private JScrollPane getTable() {
     }
 
     public void refresh() {
-        String txt = getTextArea().getText().toUpperCase();
-        var pattern = Pattern.compile("SELECT .* FROM ");
-        var matcher = pattern.matcher(txt);
-        if(matcher.find()){
+        final String sqlStatement = getTextArea().getText().toUpperCase();
```
wsbrenk: Other statements than selects may also fail if the statement is converted to uppersase.

wsbrenk (changes requested):

src/main/java/core/db/frontend/SQLDialog.java
sgcr: Yes, of course. I saw that but forgot to change it here.
Fixed.

wsbrenk (approved):

PR #2494, opened 2026-09-13
refactor: Move StringUtils.getResultString to MatchResultFormatter.format

sgcr:
Move `StringUtils.getResultString` to `MatchResultFormatter.format`

 - [x] does not require update


src/main/java/core/model/match/MatchResultFormatter.java
```diff
@@ -0,0 +1,36 @@
...
+public final class MatchResultFormatter {
+
+    private MatchResultFormatter() {
+        throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
+    }
+
+    public static final String TRANSLATOR_KEY_SEPARATOR = "ls.match.result.separation";
+
+    private static String getSeparator() {
+        return " " + TranslationFacility.tr(TRANSLATOR_KEY_SEPARATOR) + " ";
```
wsbrenk: Is it a good idea to retranslate everything every time a result is displayed? I think the old solution is better, the one that remembers the result.

src/test/java/core/model/match/MatchResultFormatterTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.MockedStatic;
+
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.params.provider.Arguments.of;
+import static org.mockito.Mockito.mockStatic;
```
wsbrenk, line 17: Can't we test this with the real Translator class? Does it have to be a mock?

wsbrenk (changes requested):

src/test/java/core/model/match/MatchResultFormatterTest.java
sgcr: Yes, we could use a real `Translator`.

The main reason for the mock is that the translation result depends on **global state**.
I wanted this test to be independent of that global state and also avoid changing it as part of the test.

The test is only about how `MatchResultFormatter` formats a given translation result, so mocking that result keeps the test deterministic and isolated.

src/main/java/core/model/match/MatchResultFormatter.java
sgcr: The `Translator` already uses a loaded `ResourceBundle`, so we don't actually retranslate or reload anything here. `tr(...)` basically performs a lookup in that bundle.

We do these translation lookups throughout the application anyway, and I don't think caching this single separator separately gives us any meaningful benefit. It just introduces additional state and logic for something that is already a very cheap lookup.

So I would prefer to keep the simpler implementation without the extra cache.

wsbrenk (approved):

Issue #2484, opened 2026-09-11
[FEATURE] New backup strategy


wsbrenk:
Keep latest backup of each year if backup is older than 12 month
Keep latest backup of each month if backup is older than 31 days
Keep at least 3 latest backups even if they are older than 31 days

wsbrenk:
Alternatively 

Keep latest Backup of each season if backup is older than 112 days
Keep latest Backup of each week if backup is older than 7 days

sgcr:
What is the goal you want to archive and why?

wsbrenk:
Many of our issues are caused by corrupt databases. 
Goal is to keep more backups.
The current approach still keeps the latest three backup zips.
Additionally each latest file of the latest 16 weeks (season) are kept
plus the latest file of previous seasons.

sgcr:
Just to make sure I understand the intended retention rules correctly:

1. Always keep the latest 3 backups (or the user-configured number).
2. Additionally, keep the latest backup of each of the latest 16 HT weeks.
3. For all earlier seasons, keep the latest backup of each season.
4. All other backups can be deleted.

Is that correct?

PR #2495, opened 2026-09-13
#2484 New Backup strategy

wsbrenk:

   - fixes issue #2484 

 - [x] has been updated
 - [ ] does not require update


src/main/java/core/util/HODateTime.java
```diff
@@ -131,6 +131,10 @@ public static long toEpochSecond(HODateTime ts) {
         return 0L;
     }
 
+    public static HODateTime fromEpochSecond(long epochSecond) {
```
sgcr, line 134: Please add a test for this new function.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,49 @@ object BackupHelper {
...
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { f -> f.lastModified() }
+                .forEach { f ->
+                    val lastModified = HODateTime(Instant.ofEpochMilli(f.lastModified()))
+                    val lastModifiedWeek = lastModified.toHTWeek();
+                    val fileAgeInDays = HODateTime.between(lastModified, currentTimestamp).toDays();
+                    if (keptFiles < UserManager.instance().currentUser.numberOfBackups ||
+                        fileAgeInDays < 112 && !lastModifiedWeek.equals(keptBackupFileLastModifiedWeek) ||
```
sgcr: What about a constant for `112`?

src/test/java/core/db/backup/BackupHelperTest.kt
```diff
@@ -1,18 +1,19 @@
...
 import java.nio.file.Files
 import java.nio.file.attribute.FileTime
 import java.time.LocalDate
-import java.time.LocalDateTime
-import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
+import java.time.temporal.ChronoUnit
 import java.util.zip.ZipFile
-
+import kotlin.time.ExperimentalTime
```
sgcr: Is there an alternative for an experimental class?

sgcr (changes requested):
The idea of this feature is not clear for me.

src/test/java/core/util/HODateTimeTest.java
```diff
@@ -44,34 +44,50 @@ void test() {
         var dti = HODateTime.fromHT("2022-02-19 23:11:00");
         assertThat(dti.toHTWeek().season).isEqualTo(80);
         assertThat(dti.toHTWeek().week).isEqualTo(10);
+        assertThat(HODateTime.toEpochSecond(dti)).isEqualTo(1645308660L);
```
sgcr, line 47: Works but a test should be better an isolated test function for a specific thing:

Here an isolated test for `fromEpochSecond` is wanted.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+        if (files != null) {
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { f -> f.lastModified() }
+                .forEach { f ->
+                    val lastModified = HODateTime(Instant.ofEpochMilli(f.lastModified()))
+                    val lastModifiedWeek = lastModified.toHTWeek()
+                    val fileAgeInDays = HODateTime.between(lastModified, currentTimestamp).toDays()
+                    val DAYS_PER_SEASON = 112
```
sgcr: Please move it to the top of the object:

`private const val DAYS_PER_SEASON = 112`

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+                    if (keptFiles < UserManager.instance().currentUser.numberOfBackups ||
+                        fileAgeInDays < DAYS_PER_SEASON && !lastModifiedWeek.equals(keptBackupFileLastModifiedWeek) ||
+                        fileAgeInDays >= DAYS_PER_SEASON && !lastModifiedWeek.season.equals(keptBackupFileLastModifiedWeek!!.season)
```
sgcr: `ileAgeInDays >= DAYS_PER_SEASON && lastModifiedWeek.season != keptBackupFileLastModifiedWeek!!.season`

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+     *  plus the configured count of latest backups
+     */
+    private fun getBackupFilesToDelete(files: List<File>?) : List<File> {
+        val ret = mutableListOf<File>()
+        if (files != null) {
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { f -> f.lastModified() }
+                .forEach { f ->
```
sgcr: Please rename `f` to `file`.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+     *  and additionally one file for each week for the last 16 weeks
+     *  plus the configured count of latest backups
+     */
+    private fun getBackupFilesToDelete(files: List<File>?) : List<File> {
+        val ret = mutableListOf<File>()
+        if (files != null) {
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { f -> f.lastModified() }
```
sgcr: Please rename `f` to `file`.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
 	 * @param dbDirectory Directory where to find the zip files to be deleted.
 	 */
 	private fun deleteOldFiles(dbDirectory: File) {
-		val files = dbDirectory.listFiles { file: File ->
-			file.isFile && file.extension == HOZip.zipExt
+        val files = dbDirectory.listFiles { file: File ->
+            file.isFile && file.extension == HOZip.zipExt
+        }?.toList()
+        val deleteBackupList = getBackupFilesToDelete(files)
+        deleteBackupList.forEach { f -> f.delete() }
```
sgcr: Please rename `f` to `file` and it should be logged in the application file that it is deleted.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
 	 * @param dbDirectory Directory where to find the zip files to be deleted.
 	 */
 	private fun deleteOldFiles(dbDirectory: File) {
-		val files = dbDirectory.listFiles { file: File ->
-			file.isFile && file.extension == HOZip.zipExt
+        val files = dbDirectory.listFiles { file: File ->
```
sgcr, line 58: `listFiles` can return `null` and should log in that case that the files could not be read due some reason.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+     * Filter backup files from list that should be deleted
+     *  - The configured count of youngest backups are removed from the list (not deleted)
+     *  - If the backup file is younger than 112 days the latest file of each week is removed
+     *  - If the backup file is older than 112 days the latest file of each season is removed
+     *
+     *  At the end the user has one backup file for each previous season
+     *  and additionally one file for each week for the last 16 weeks
+     *  plus the configured count of latest backups
+     */
+    private fun getBackupFilesToDelete(files: List<File>?) : List<File> {
```
sgcr: I think `null` should be not allowed here for `files`.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+        val ret = mutableListOf<File>()
+        if (files != null) {
```
sgcr: Suggestion: if can be removed if `null` is not allowed anymore

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +54,50 @@ object BackupHelper {
...
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { f -> f.lastModified() }
+                .forEach { f ->
+                    val lastModified = HODateTime(Instant.ofEpochMilli(f.lastModified()))
+                    val lastModifiedWeek = lastModified.toHTWeek()
+                    val fileAgeInDays = HODateTime.between(lastModified, currentTimestamp).toDays()
+                    val DAYS_PER_SEASON = 112
+                    if (keptFiles < UserManager.instance().currentUser.numberOfBackups ||
```
sgcr: I think the opposite perspective is more useful here:
When should backup files be deleted, rather than when should they be kept?

Deletion is the destructive operation, while keeping backups is not.
So I would focus the logic on the conditions under which a backup should be deleted.

sgcr: `val numberOfBackups = UserManager.instance().currentUser.numberOfBackups`

Should be moved to the top of the function and used here.

sgcr (changes requested):

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: I've learned to code the more likely case into the if branch and the rarer one into the else branch.

And I think negating the logical expression won't make it more readable

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: Done

src/test/java/core/util/HODateTimeTest.java
```diff
@@ -44,34 +44,73 @@ void test() {
...
+        assertThat(HODateTime.fromEpochSecond(1600034400L)).isEqualTo(dti);
 
         dti = HODateTime.fromHT("2020-09-21 01:30:00");
         assertThat(dti.toHTWeek().season).isEqualTo(76);
         assertThat(dti.toHTWeek().week).isEqualTo(1);
+        assertThat(HODateTime.toEpochSecond(dti)).isEqualTo(1600644600L);
+        assertThat(HODateTime.fromEpochSecond(1600644600L)).isEqualTo(dti);
+    }
+
+    static Stream<Arguments> testEpochSeconds() {
```
sgcr, line 93: Should be private.

src/main/java/core/db/backup/BackupHelper.kt
sgcr: I am missing the logging for the files that will be deleted. 😉

src/main/java/core/util/HODateTime.java
sgcr, line 13: Can be removed.

src/test/java/core/util/HODateTimeTest.java
```diff
@@ -44,34 +44,73 @@ void test() {
...
+            Arguments.of("2020-09-14 00:00:00", 1600034400L),
+            Arguments.of("2020-09-21 01:30:00", 1600644600L)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testEpochSeconds(String date, long epochSeconds) {
+        var dti = HODateTime.fromHT(date);
+        assertThat(HODateTime.toEpochSecond(dti)).isEqualTo(epochSeconds);
```
sgcr, line 112: Please add a separate test for `toEpochSecond` and don't mix it with the test of `fromEpochSecond`.

I appreciate that for this PR but that was not my intention.

Have a look at this:
```
    private static Stream<Arguments> testFromEpochSecond() {
        return Stream.of(
            Arguments.of(-3600L, HODateTime.fromHT("1970-01-01 00:00:00")),
            Arguments.of(0L, HODateTime.fromHT("1970-01-01 01:00:00")),
            Arguments.of(1645308660L, HODateTime.fromHT("2022-02-19 23:11:00")),
            Arguments.of(1613340660L, HODateTime.fromHT("2021-02-14 23:11:00")),
            Arguments.of(1593208800L, HODateTime.fromHT("2020-06-27 00:00:00")),
            Arguments.of(1525903200L, HODateTime.fromHT("2018-05-10 00:00:00")),
            Arguments.of(1243461600L, HODateTime.fromHT("2009-05-28 00:00:00")),
            Arguments.of(1599429600L, HODateTime.fromHT("2020-09-07 00:00:00")),
            Arguments.of(1600034400L, HODateTime.fromHT("2020-09-14 00:00:00")),
            Arguments.of(1600644600L, HODateTime.fromHT("2020-09-21 01:30:00"))
        );
    }

    @ParameterizedTest
    @MethodSource
    void testFromEpochSecond(long epochSeconds, HODateTime expected) {
        final var result = HODateTime.fromEpochSecond(epochSeconds);
        assertThat(result).isEqualTo(expected);
        assertThat(HODateTime.toEpochSecond(result)).isEqualTo(epochSeconds);
    }
```

The seconds assertion is optional but it make sense to verify the value that comes in is the same that comes out from the result.
But this is the intentionally the second assertion. The first assertion is the important one here.

The other thing is that we want to give the test function an `HODateTime` object that is equal to our **expected** result. I think the construction of a `HODateTime` from a `String` works and it is well tested enough.

src/main/java/core/db/backup/BackupHelper.kt
sgcr: In general, that's not a bad idea, but I don't have a strong opinion on it at the moment. I have to think about it a bit more.

I agree that simply negating the existing expression would probably not improve readability.

My point is rather that I would formulate the condition directly from the deletion perspective, e.g. as a `shouldBeDeleted` condition, instead of deriving deletion indirectly from the cases in which we keep a backup.

I don't think putting the more likely case into the `if` branch should outweigh the semantics here. Deletion is the actual operation performed after this method, so from my point of view the condition should make it easy to understand why a file is deleted rather than why it is kept.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +55,53 @@ object BackupHelper {
...
+    }
 
-		}?.toList()
+    /**
+     * Filter backup files from list that should be deleted
+     *  - The configured count of youngest backups are removed from the list (not deleted)
+     *  - If the backup file is younger than 112 days the latest file of each week is removed
+     *  - If the backup file is older than 112 days the latest file of each season is removed
+     *
+     *  At the end the user has one backup file for each previous season
```
sgcr, line 76: What means previous season?
The season before the current season?
Is the current season derived from the the youngest backup or from the current season in the application?

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +55,53 @@ object BackupHelper {
...
+            deleteBackupList.forEach { file -> file.delete() }
+        } else {
+            HOLogger.instance().warning(this.javaClass, "No files to delete in directory $dbDirectory")
+        }
+    }
 
-		}?.toList()
+    /**
+     * Filter backup files from list that should be deleted
+     *  - The configured count of youngest backups are removed from the list (not deleted)
```
sgcr, line 72: This is also not clear:
They are not removed from the list but they are never included in the returned list of the files that should be deleted.
The original list is not modified.

sgcr (changes requested):

src/test/java/core/util/HODateTimeTest.java
wsbrenk: Changed it.

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: this was not in your request ;-)

I will add it.

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: I see it exactly the other way around: A long list of files is passed, from which usually 2 files are deleted: the fourth-to-last of the current week and the weekly backup from the previous season that is no longer needed.

src/test/java/core/util/HODateTimeTest.java
wsbrenk: I copied it from your example!

src/test/java/core/util/HODateTimeTest.java
wsbrenk: These are exactly the same asserts than my implementation. I don't see the reason, why i should double it?

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: for 16 weeks one file per week is kept.  For the time (seasons) before, one file per season is kept.

src/test/java/core/util/HODateTimeTest.java
sgcr: My proposal is the substitute for your `testEpochSeconds`.
1. The name of the function should correspond to the function it is testing.
2. Your function uses a `String` for the `HODateTime` from the outside. Better is to do that in the stream of arguments function and give the test the direct object that shall be compared.
Construction from `String` works but `fromHT` may fail. So best isolated option would be to construct the `HODateTime` from the local function `localDateTimeToHoDateTime` because there is lowest chance that anything can fail (basically from an Instant).
Due to trust in `fromHT` from String is here acceptable.
3. The parameters for the test should be distinguishable so that you can see the incoming parameters and the **expected** result(s). Typically named `expected`, `expectedXyz` or`result`.
4. There should be at least one assertion that point directly to what shall be tested here. This should be the first one. Other assertions are possible but should be only there if really necessary.

I would add a second test explicitly for testing `toEpochSecond`.

src/main/java/core/db/backup/BackupHelper.kt
sgcr: Lets see where it ends but it should be readable in general.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +55,59 @@ object BackupHelper {
...
-		val files = dbDirectory.listFiles { file: File ->
-			file.isFile && file.extension == HOZip.zipExt
+        val files = dbDirectory.listFiles { file: File ->
+            file.isFile && file.extension == HOZip.zipExt
+        }
+        if (files != null && files.isNotEmpty()) {
+            val fileList = files.toList()
+            val deleteBackupList = getBackupFilesToDelete(fileList)
+            deleteBackupList.forEach { file ->
+                if ( file.delete() ) {
```
sgcr: Code-Formatter

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +55,59 @@ object BackupHelper {
...
+        var keptFiles = 0
+        val currentTimestamp = HODateTime.now()
+        files.sortedByDescending { file -> file.lastModified() }
+            .forEach { file ->
+                val lastModified = HODateTime(Instant.ofEpochMilli(file.lastModified()))
+                val lastModifiedWeek = lastModified.toHTWeek()
+                val fileAgeInDays = HODateTime.between(lastModified, currentTimestamp).toDays()
+                if (keptFiles < numberOfBackups ||
+                    fileAgeInDays < DAYS_PER_SEASON && !lastModifiedWeek.equals(keptBackupFileLastModifiedWeek) ||
+                    fileAgeInDays >= DAYS_PER_SEASON && lastModifiedWeek.season != keptBackupFileLastModifiedWeek!!.season
```
sgcr: If `numberOfBackups` can be `0` then this may fail due to `keptBackupFileLastModifiedWeek` is not initialized.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -52,19 +55,59 @@ object BackupHelper {
 	 * @param dbDirectory Directory where to find the zip files to be deleted.
 	 */
 	private fun deleteOldFiles(dbDirectory: File) {
-		val files = dbDirectory.listFiles { file: File ->
-			file.isFile && file.extension == HOZip.zipExt
+        val files = dbDirectory.listFiles { file: File ->
+            file.isFile && file.extension == HOZip.zipExt
+        }
+        if (files != null && files.isNotEmpty()) {
```
sgcr: This mixes two cases:
`files == null` is listing of files failed due to permissions of something like that.
Should be logged.

`files.isNotEmpty()` the may be zero backups resident.

src/main/java/core/db/backup/BackupHelper.kt
sgcr: If creating the backup itself failed the old backups are deleted anyway.
Is this the intention?

wsbrenk: In the previous implementation, this was probably a no-op if the user didn't reduce the number of backups. Now, at most, one expired weekly backup might disappear, and there could also be an adjustment to a reduced number of backups.

sgcr: I understand that the practical impact is probably small in most cases.

My concern is more about the semantics: if creating the new backup fails, should we perform any retention cleanup at all?

From my point of view, cleanup should only happen after a new backup has been created successfully. Otherwise we may delete an existing backup even though no replacement was created.

wsbrenk: Changed it.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -14,57 +16,95 @@ import java.util.*
...
+    private const val DAYS_PER_SEASON = 112
 
 	// zip and delete db
 	@JvmStatic
 	fun backup(dbDirectory: File) {
-		if (!dbDirectory.exists()) {
-			return
-		}
+        val numberOfBackups = UserManager.instance().currentUser.numberOfBackups
+        if (!dbDirectory.exists() || numberOfBackups < 1) {
```
sgcr: Should be separated:

It should be logged:
```
if (!dbDirectory.isDirectory) {
    HOLogger.instance().error(
        BackupHelper::class.java,
        "Backup directory does not exist or is not a directory: $dbDirectory"
    )
    return
}
```

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -14,57 +16,95 @@ import java.util.*
...
+     *  - The configured count of youngest backups are not included to the result (not deleted)
+     *  - If the backup file is younger than 112 days the latest file of each week is not included
+     *  - If the backup file is older than 112 days the latest file of each season is not included
+     *
+     * When the returned files are deleted, the user has a backup file for each previous season
+     * and also a file for each week of the last 16 weeks, plus the configured number of the most recent backups.
+     */
+    private fun getBackupFilesToDelete(files: List<File>, numberOfBackups : Int) : List<File> {
+        val ret = mutableListOf<File>()
+        if (numberOfBackups > 0) {
```
sgcr: `0` cannot happen anymore and due to a private function I would add in the first line of the function:
`require(numberOfBackups > 0) { "numberOfBackups must be greater than 0" }`

And then remove `if (numberOfBackups > 0) {` completely.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -14,57 +16,95 @@ import java.util.*
...
+            var keptBackupFileLastModifiedWeek: HODateTime.HTWeek? = null
+            var keptFiles = 0
+            val currentTimestamp = HODateTime.now()
+            files.sortedByDescending { file -> file.lastModified() }
+                .forEach { file ->
+                    val lastModified = HODateTime(Instant.ofEpochMilli(file.lastModified()))
+                    val lastModifiedWeek = lastModified.toHTWeek()
+                    val fileAgeInDays = HODateTime.between(lastModified, currentTimestamp).toDays()
+                    if (keptFiles < numberOfBackups || keptBackupFileLastModifiedWeek == null ||
+                        fileAgeInDays < DAYS_PER_SEASON && !lastModifiedWeek.equals(keptBackupFileLastModifiedWeek) ||
```
sgcr: Although they are not strictly necessary, I suggest adding parentheses around

`fileAgeInDays < DAYS_PER_SEASON && !lastModifiedWeek.equals(keptBackupFileLastModifiedWeek)`

and around

`fileAgeInDays >= DAYS_PER_SEASON && lastModifiedWeek.season != keptBackupFileLastModifiedWeek.season`.

I would also put `keptBackupFileLastModifiedWeek == null ||` on a separate line for readability.

src/test/java/core/util/HODateTimeTest.java
```diff
@@ -74,6 +74,49 @@ void test() {
...
+            Arguments.of(HODateTime.fromHT("2009-05-28 00:00:00"), 1243461600L),
+            Arguments.of(HODateTime.fromHT("2020-09-07 00:00:00"), 1599429600L),
+            Arguments.of(HODateTime.fromHT("2020-09-14 00:00:00"), 1600034400L),
+            Arguments.of(HODateTime.fromHT("2020-09-21 01:30:00"), 1600644600L)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testFromEpochSecond(HODateTime date, long epochSeconds) {
```
sgcr, line 115: Perfect.

If you want to prevent duplicating the function with the stream, then
you can write this `@MethodSource("testToEpochSecond")`.

Beside that its better to switch the arguments:
`epochSeconds` first and then `expectedDate`.

It is possible to get the function with the arguments from the existing function and switching the arguments with this:
```
private static Stream<Arguments> testFromEpochSecond() {
        return testToEpochSecond().map(arguments -> Arguments.of(arguments.get()[1], arguments.get()[0]));
}
```

But you can leave it as it is.

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: Extra line is done - unnecessary parenthesis not - i don't like them

src/test/java/core/util/HODateTimeTest.java
wsbrenk: I'll leave it up to you to refine this in a future ticket

src/main/java/core/db/backup/BackupHelper.kt
sgcr: Very good. 😄

Just keep in mind that, although I appreciate making use of well-known operator precedence, I still think readability is important. Explicit parentheses can make complex conditions easier to verify and therefore less prone to bugs. 😉

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -21,8 +21,20 @@ object BackupHelper {
...
+            )
+            return
+        }
+
         val numberOfBackups = UserManager.instance().currentUser.numberOfBackups
-        if (!dbDirectory.exists() || numberOfBackups < 1) {
+        if (numberOfBackups < 1) {
+            HOLogger.instance().warning(
+                BackupHelper::class.java,
+                "Number of backups is set to 0. No backup of HO database will be created"
```
sgcr, line 36: Proposal:
`"Number of backups is less than 1. No backup of the HO database will be created."`

Also consider whether this should be logged as `info` rather than `warning`.

I think both are reasonable. From a backup perspective, having no backups is not very risk-aware, but on the other hand, this may be an intentional user decision.

src/main/java/core/db/backup/BackupHelper.kt
sgcr, line 57: Should be changed to `error`.

src/main/java/core/db/backup/BackupHelper.kt
```diff
@@ -14,57 +16,107 @@ import java.util.*
...
+    /**
+     * Filter backup files from list that should be deleted
+     *  - The configured count of youngest backups are not included to the result (not deleted)
+     *  - If the backup file is younger than 112 days the latest file of each week is not included
+     *  - If the backup file is older than 112 days the latest file of each season is not included
+     *
+     * When the returned files are deleted, the user has a backup file for each previous season
+     * and also a file for each week of the last 16 weeks, plus the configured number of the most recent backups.
+     */
+    private fun getBackupFilesToDelete(files: List<File>, numberOfBackups : Int) : List<File> {
```
sgcr, line 94: Code-Formatter

src/main/java/core/db/backup/BackupHelper.kt
sgcr: Acceptable from my point of view.

sgcr (approved):
Approved in general. 🥳

Please have a look at my last comments and decide what to do with that.

src/main/java/core/db/backup/BackupHelper.kt
wsbrenk: Sure, this can be an intentional user choice—but I think it deserves a warning because it creates work for us when the consequences end up as an error message on our side and can't be fixed.

PR #2498, opened 2026-09-14
feat: Show the age of the player additionally with the days

sgcr:
Show the age of the player additionally with the days.

 - [x] has been updated


src/main/java/module/matches/SpielerDetailDialog.java
```diff
@@ -228,7 +229,7 @@ private void setLabels(Player m_clPlayer) {
 				.getCurrentPlayer(m_clPlayer.getPlayerId());
 
 		m_jpName.setText(m_clPlayer.getFullName());
-		m_jpAlter.setText(m_clPlayer.getAge() + "");
+		m_jpAlter.setText(new HODuration((long)m_clPlayer.getAge(), m_clPlayer.getAgeDays()).toAgeString());
```
wsbrenk:
```suggestion
		m_jpAlter.setText(m_clPlayer.getAgeWithDaysAsString());
```

wsbrenk (changes requested):

src/main/java/module/matches/SpielerDetailDialog.java
sgcr: In general a good idea but `Player.getAgeWithDaysAsString` calculates the age with the current date and time `HODateTime.now()`.

What we want to see here is the historic age of the player when the match happened.
So the values from the player are right.

Any alternative?

wsbrenk: m_clPlayer.getAgeWithDaysAsString(m_clPlayer.getHrfDate())

sgcr: Fixed.

wsbrenk (approved):

Issue #2414, opened 2026-02-01
[FEATURE] Smart Player Names


sgcr:
**Is your feature request related to a problem? Please describe.**
The application shows in several places the name of a player (team and youth).
Depending on the part of the application the style of the name you want so see is different.

Example:
For the lineup it is sufficient if only the surname (last name) of a player is shown.
But what if you have two players with the same surname in the team?
What about the nickname of a player that makes it unique?

On the other side most of the time the Hattrick style of player names is used so that lists start and are sorted by the first name of player.
This is also not very convenient because in reality you are talking about players with their surname first or the nick name.

**Describe the solution you'd like**
I want to have the possibility to adjust in the settings the player name style for each different parts of the application.

Available styles should be:
1. Full name starting with the first-name with optional nickname in the format: `[FIRST_NAME] [OPTIONAL_NICK_NAME] [LAST_NAME]`. The optional nickname is displayed in single quotes.
E.g. `Christiano 'CR7' Ronaldo` with nickname and `Christiano Ronaldo` without..
2. Full name starting with the last-name with optional nickname in the format: `[LAST_NAME], FIRST_NAME] [OPTIONAL_NICK_NAME]`
E.g. `Ronaldo, Christiano 'CR7'` with nickname and `Ronaldo, Christiano` without..
3. Smart player name starting with the first-name, optional nickname and optional firstname:
If there is a nickname which is unique, that unique nickname will be displayed in quotes.
E.g. `'CR7'`
If there is no nickname, the format will be `[OPTIONAL_FIRST_NAME_SHORT] [LAST_NAME]`.
`[OPTIONAL_FIRST_NAME_SHORT]` is only displayed if the player is not unique by its last-name. If the first-name of player is necessary because the last-name is not unique, then there are so many characters of the first-name displayed until the whole name is unique.
E.g. if you have a `Christiano Ronaldo` and a `Fred Ronaldo` then `C. Ronaldo` and `F. Ronaldo` is displayed.
But if you have `Christiano Ronaldo` and `Chris Ronaldo` then `Christ. Ronaldo` and `Chris Ronaldo` is displayed. `Chris` without a dot because the first-name is not cut.
`Christ.` because it becomes unique with the `t` and is cut. But these are very rare cases.
If you have really two players the same name then the Hattrick player id is append in braces like `(12345)`.
4. Smart player name starting with the lame-name that works like 3. but starting with the last-name.

**Additional context**
-

PR #2499, opened 2026-09-14
refactor: PlayerPerformance: Add English get functions and deprecate the ones in German

sgcr:
Add English get functions and deprecate the ones in German.
I need some small refactorings for: #2414

 - [x] does not require update


src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
```diff
@@ -26,47 +28,77 @@ public class PlayerPerformance {
...
+     *
+     * @deprecated Please use {@link #getLastName()} instead.
+     */
+    @Deprecated(since = "10.0", forRemoval = true)
     public String getSpielerName() {
-        return mlp.getSpielerName();
+        return getLastName();
     }
 
+    public String getLastName() {
```
wsbrenk, line 86: getPlayerName 
I think LastName is no attribute for PlayerPerformance

Please remove the old (deprecated) method.

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
```diff
@@ -26,47 +28,77 @@ public class PlayerPerformance {
...
+     * @deprecated Please use {@link #getFirstName()} instead.
+     */
+    @Deprecated(since = "10.0", forRemoval = true)
     public String getSpielerVName() {
-        return mlp.getSpielerVName();
+        return getFirstName();
+    }
+
+    public String getFirstName() {
+        return matchLineupPosition.getSpielerVName();
```
wsbrenk, line 101: This is not used.

wsbrenk (changes requested):

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
sgcr: What do you mean by that it is no attribute?
I have 3 matches for `getLastName` now that was former `getSpielerName`

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
sgcr: True but that is the problem at other places.
I want to use it in my draft for the player names feature.

I think First Name, Last Name and Nick Name should be provided.

Do you have another opinion?

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
wsbrenk: I would prefer an exact translation of the former name: getPlayerName

It is the name of the player of the object PlayerPerformance and not the last name of the PlayerPerformance.

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
wsbrenk: If you want to do this, you could introduce a method getPlayer that calls mlp.getPlayer()
Giving you access to all player infos. 
I would implement the name feature only once (in class Player) and not additionally in PlayerPerformance and MatchLineupPosition.

sgcr: Added.

sgcr: Removed `getFirstName` and unused `getSpielerVName`

sgcr: Removed `getNickName

src/main/java/module/teamanalyzer/vo/PlayerPerformance.java
```diff
@@ -26,47 +29,63 @@ public class PlayerPerformance {
...
-        return mlp.getSortId();
+        return matchLineupPosition.getSortId();
     }
 
+    /**
+     * Gets the ID of the player.
+     *
+     * @deprecated Please use {@link #getPlayerId()} instead.
+     */
+    @Deprecated(since = "10.0", forRemoval = true)
```
wsbrenk: Don't deprecate - remove the old getSpieler*

sgcr: Removed.

What about `getSpielerId`?

wsbrenk: Also - that's why I added the asterisk (*)

sgcr: Done.

What is your expectation when should we use `@Deprecated` and when not?

wsbrenk: To be honest, I don't have a concrete idea about that. But in this case, when it's just simple getters and not even a handful of files are affected, I think deprecated is over the top.

wsbrenk (approved):

Issue #2310, opened 2025-08-24
new team


ryszaardo-cmyk:
Hi Guys, 
i came back to hattrick after long break with new team, I installed HO and there was my old team. I gave authorization to my new team and tried to download all of the information but error occurred. I tried to uninstall and install again but its not helping. I tried to remove data from HO but it doesn't work as well. Any ideas?

<img width="783" height="448" alt="Image" src="https://github.com/user-attachments/assets/08824a7a-bcf3-477d-9f5d-a47a8de72967" />

java.lang.NullPointerException: Cannot invoke "core.util.HODateTime.toHTWeek()" because "fetchedDate" is null
	at core.file.xml.XMLEconomyParser.parseEconomyFromString(XMLEconomyParser.java:25)
	at core.file.xml.ConvertXml2Hrf.createHrf(ConvertXml2Hrf.java:145)
	at core.net.OnlineWorker.getHrf(OnlineWorker.java:63)
	at core.net.DownloadDialog.startDownload(DownloadDialog.java:278)
	at core.net.DownloadDialog.actionPerformed(DownloadDialog.java:87)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:297)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6626)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3389)
	at java.desktop/java.awt.Component.processEvent(Component.java:6391)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5001)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4833)
[...]

PR #2504, opened 2026-09-21
#2310 Replace database of previous team

wsbrenk:

   - fixes issue #2310

 - [x] has been updated
 - [ ] does not require update


src/main/resources/language/German.properties
```diff
@@ -2248,6 +2248,13 @@ ls.update_status.create_world = Erstelle Welt...
 ls.update_status.create_last_lineup = Erstelle letzte Aufstellung...
 ls.update_status.create_staff = Erstelle Mitarbeiter...
 ls.update_status.complete = Vollständig
+
+ls.download.replace.old.database=Soll die alte Datenbank ersetzt werden?
+ls.download.database.of.other.team=Dies Team gehört dir nicht (mehr)
+ls.download.cannot.save.old.database=Die alte Datenbank kann nicht gesichert werden.
+ls.download.remove.database=Entferne die alte Datenkbank
```
sgcr:
```suggestion
ls.download.remove.database=Entferne die alte Datenbank
```

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
 		var teamInfoList = XMLTeamDetailsParser.getTeamInfoFromString(teamDetails);
 		var usersPremierTeamInfo = teamInfoList.stream().filter(TeamInfo::isPrimaryTeam).findFirst().orElse(teamInfoList.stream().findFirst().orElseThrow());
 		var usersPremierTeamId = usersPremierTeamInfo.getTeamId();
-		var initTeamId = teamId <=0;
+        int finalTeamId1 = teamId;
+        var isDatabaseFromOtherTeam = teamId > 0 && teamInfoList.stream().noneMatch(team->team.getTeamId() == finalTeamId1);
+        if (isDatabaseFromOtherTeam) {
+            // Ask user if old database should be deleted
+            var message = TranslationFacility.tr("ls.download.replace.old.database");
+            var title = TranslationFacility.tr("ls.download.database.of.other.team");
```
sgcr:
```suggestion
            final var title = TranslationFacility.tr("ls.download.database.of.other.team");
```

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
...
+            int choice = JOptionPane.showConfirmDialog(parent, message, title, JOptionPane.OK_CANCEL_OPTION);
+            if (choice == JOptionPane.OK_OPTION) {
+                if (!DBManager.deleteDatabaseOfPreviousTeam(teamId)){
```
sgcr: Code-Formatter

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
 		var teamInfoList = XMLTeamDetailsParser.getTeamInfoFromString(teamDetails);
 		var usersPremierTeamInfo = teamInfoList.stream().filter(TeamInfo::isPrimaryTeam).findFirst().orElse(teamInfoList.stream().findFirst().orElseThrow());
 		var usersPremierTeamId = usersPremierTeamInfo.getTeamId();
-		var initTeamId = teamId <=0;
+        int finalTeamId1 = teamId;
+        var isDatabaseFromOtherTeam = teamId > 0 && teamInfoList.stream().noneMatch(team->team.getTeamId() == finalTeamId1);
```
sgcr:
```suggestion
        final var isDatabaseFromOtherTeam = teamId > 0 && teamInfoList.stream().noneMatch(team -> team.getTeamId() == finalTeamId1);
```

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
...
+        if (isDatabaseFromOtherTeam) {
+            // Ask user if old database should be deleted
+            var message = TranslationFacility.tr("ls.download.replace.old.database");
+            var title = TranslationFacility.tr("ls.download.database.of.other.team");
+            int choice = JOptionPane.showConfirmDialog(parent, message, title, JOptionPane.OK_CANCEL_OPTION);
```
sgcr:
```suggestion
            final int choice = JOptionPane.showConfirmDialog(parent, message, title, JOptionPane.OK_CANCEL_OPTION);
```

sgcr: I think it is a `YES_NO_OPTION` option and not `OK_CANCEL_OPTION`.

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
...
+            if (choice == JOptionPane.OK_OPTION) {
+                if (!DBManager.deleteDatabaseOfPreviousTeam(teamId)){
+                    // Backup of old database cannot be created
+                    message = "<html>" +
+                        TranslationFacility.tr("ls.download.cannot.save.old.database") + "<br>" +
```
sgcr, line 79: It is better in general to provide a text with a placeholder (`%s` or another if we have a system for that) instead of translating text fragments.
Reason: In other languages the fragments may be arranges in a totally different way.

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
 		var teamInfoList = XMLTeamDetailsParser.getTeamInfoFromString(teamDetails);
 		var usersPremierTeamInfo = teamInfoList.stream().filter(TeamInfo::isPrimaryTeam).findFirst().orElse(teamInfoList.stream().findFirst().orElseThrow());
 		var usersPremierTeamId = usersPremierTeamInfo.getTeamId();
-		var initTeamId = teamId <=0;
+        int finalTeamId1 = teamId;
+        var isDatabaseFromOtherTeam = teamId > 0 && teamInfoList.stream().noneMatch(team->team.getTeamId() == finalTeamId1);
+        if (isDatabaseFromOtherTeam) {
+            // Ask user if old database should be deleted
+            var message = TranslationFacility.tr("ls.download.replace.old.database");
```
sgcr: You may use this message text directly for the confirmation dialog instead of reassigning `message` later.
`message` could be moved in the if below if needed anymore here.

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -63,7 +66,34 @@ private ConvertXml2Hrf() {
...
+                // Forget the old team id
+                teamId = 0;
+                HOVerwaltung.instance().getModel().getBasics().setTeamId(0);
+            }
+            else {
+                // Terminate HO. User has to remove database folder manually
+                System.exit(0);
+            }
+        }
+        var initTeamId = teamId <= 0;
```
sgcr, line 70:
```suggestion
        final var initTeamId = teamId <= 0;
```

src/main/java/core/db/DBManager.java
```diff
@@ -223,6 +218,30 @@ public static synchronized DBManager instance() {
 		return m_clInstance;
 	}
 
+    public static File getDbFolder() {
+        var current_user = UserManager.instance().getCurrentUser();
```
sgcr, line 222:
```suggestion
        var currentUser = UserManager.instance().getCurrentUser();
```

src/main/java/core/db/DBManager.java
```diff
@@ -223,6 +218,30 @@ public static synchronized DBManager instance() {
...
+        var dbFolderName = current_user.getDbFolder();
+        return new File(dbFolderName);
+    }
+
+    public static boolean deleteDatabaseOfPreviousTeam(int teamId) {
```
sgcr, line 227: Delete or rename?

sgcr (changes requested):

src/main/java/core/db/DBManager.java
wsbrenk: I was wondering about that too and went with delete, since that's what it's for in HO use. The rename is just a backup to give the user a copy.

src/main/java/core/file/xml/ConvertXml2Hrf.java
wsbrenk: Chnaged it.

src/main/java/core/db/DBManager.java
sgcr: Yes, got it.

Please add a comment to the function.
Something like that:
```
/**
 * Removes the database of the previous team from active use.
 * <p>
 * The database files are not physically deleted. Instead, the database folder
 * is renamed to {@code <database-folder>-old-<teamId>} to keep a backup for the user.
 *
 * @param teamId the id of the previous team
 * @return {@code true} if no database exists or if the database was successfully moved
 *         to the backup location; {@code false} otherwise
 */
 ```

File: src/main/java/core/file/xml/ConvertXml2Hrf.java [hunk 11]
```diff
@@ -47,218 +50,243 @@ private ConvertXml2Hrf() {
 	/**
 	 * Create the HRF data and return it in one string.
 	 */
-	public static @Nullable String createHrf() throws IOException {
```
sgcr, line 50: In general:

The whole function seems to have been reformatted, which I think is a good idea in general.
Maybe we should document this as a recommendation for functions that are touched anyway.

In this case, however, it makes it harder to see the actual changes.
It also affects the history of the code, because some lines now show you as the latest author even though they were only reformatted and not functionally changed.

When viewing the diff without whitespace changes, the actual changes become much clearer.

So the question is:
Do you want to preserve the existing history of where the code originally came from, or was it your intention to replace it with the reformatted version?

It is not that important to me in this particular case, because the function is already far too long and should be refactored anyway, in my opinion.

My proposal as a compromise would be:
Move the newly added functionality into a separate sub-function and call it from here.

That would also avoid making the already long function even longer.

What do you think?

sgcr (changes requested):

src/main/resources/language/English.properties
```diff
@@ -2299,6 +2299,12 @@ ls.update_status.complete=Complete
 ls.update_status.hall_of_fame=Downloading hall of fame ...
 ls.next.daily.update=Next daily update
 
+ls.download.replace.old.database=Should the old database be replaced?
+ls.download.database.of.other.team=This team doesn't belong to you (anymore)
+ls.download.cannot.save.old.database=Cannot save the old database
+ls.download.remove.database=Remove the old database %s
+ls.download.restart=and restart HO
```
sgcr: My intention was that this text is part of `ls.download.remove.database`.

src/main/java/core/file/xml/ConvertXml2Hrf.java
wsbrenk: Was not intentional - some review changes must trigger that.
I tried to reset the previous commit.

src/main/resources/language/English.properties
wsbrenk: I thought it wasn't a good idea to have translators mess around with HTML syntax.

But there exists already some examples, where they have to do so.

src/main/resources/language/English.properties
```diff
@@ -2299,6 +2299,10 @@ ls.update_status.complete=Complete
 ls.update_status.hall_of_fame=Downloading hall of fame ...
 ls.next.daily.update=Next daily update
 
+ls.download.replace.old.database=Should the old database be replaced?
+ls.download.database.of.other.team=This team doesn't belong to you (anymore)
+ls.download.cannot.save.old.database=<html>Cannot save the old database<br>Remove the old database %s<br>and restart HO</html>
```
sgcr, line 2304: We have this other option with `{0}` with the translations function with parameters.

See below `ls.player_details.desc1`.

Is there any advantage for the placeholders with `{N}`?
Well the disadvantage is that you have to count. 😉

sgcr:
```suggestion
ls.download.cannot.save.old.database=<html>Cannot save the old database.<br>Remove the old database %s<br>and restart HO?</html>
```

Added the dot and the question mark as proposal.

src/main/resources/language/German.properties
```diff
@@ -2250,6 +2250,11 @@ ls.update_status.create_staff = Erstelle Mitarbeiter...
 ls.update_status.complete = Vollständig
 ls.update_status.hall_of_fame=Hall of Fame herunterladen...
 ls.next.daily.update=Nächste tägliche Aktualisierung
+
+ls.download.replace.old.database=Soll die alte Datenbank ersetzt werden?
+ls.download.database.of.other.team=Dies Team gehört dir nicht (mehr)
+ls.download.cannot.save.old.database=<html>Die alte Datenbank kann nicht gesichert werden<br>Entferne die alte Datenbank %s<br>und starte HO erneut</html>
```
sgcr:
```suggestion
ls.download.cannot.save.old.database=<html>Die alte Datenbank kann nicht gesichert werden.<br>Entferne die alte Datenbank %s<br>und starte HO erneut?</html>
```

Added the dot and the question mark as proposal.

src/main/resources/language/English.properties
sgcr, line 2304:
```suggestion
ls.download.cannot.save.old.database=<html>Cannot save the old database.<br>Please remove the old database %s<br>and restart HO.</html>
```

Added dot and the word `Please` as proposal.

src/main/resources/language/German.properties
sgcr, line 2256:
```suggestion
ls.download.cannot.save.old.database=<html>Die alte Datenbank kann nicht gesichert werden.<br>Bitte entferne die alte Datenbank %s<br>und starte HO erneut.</html>
```

Added the dots and the word `Bitte` as proposal.

src/main/java/core/file/xml/ConvertXml2Hrf.java
```diff
@@ -260,7 +264,41 @@ private ConvertXml2Hrf() {
...
+        int finalTeamId1 = teamId;
+        final var isDatabaseFromOtherTeam = teamId > 0 && teamInfoList.stream().noneMatch(team -> team.getTeamId() == finalTeamId1);
+        if (isDatabaseFromOtherTeam) {
+            // Ask user if old database should be deleted
+            final var title = TranslationFacility.tr("ls.download.database.of.other.team");
+            final int choice = JOptionPane.showConfirmDialog(parent, TranslationFacility.tr("ls.download.replace.old.database"), title, JOptionPane.YES_NO_OPTION);
+            if (choice == JOptionPane.YES_OPTION) {
+                if (!DBManager.deleteDatabaseOfPreviousTeam(teamId)) {
+                    // Backup of old database cannot be created
+                    var message = String.format(TranslationFacility.tr("ls.download.database.of.database"), DBManager.getDbFolder().getAbsolutePath());
```
sgcr, line 286:
```suggestion
                    var message = String.format(TranslationFacility.tr("ls.download.cannot.save.old.database"), DBManager.getDbFolder().getAbsolutePath());
```

Translation key was renamed.

sgcr (approved):
In general approved 🥳 but please have a look at my last comments.

